Issue #5, opened 2020-06-22
Tools for enforcing coding standards


tomwhite:
Which tools should we use for enforcing coding standards?

Here is a table summarising what (related projects) [Zarr](https://zarr.readthedocs.io/en/stable/contributing.html#code-standards), [Dask](https://docs.dask.org/en/latest/develop.html#code-formatting), and [Xarray](https://xarray.pydata.org/en/latest/contributing.html#contributing-to-the-code-base) use.

| |Zarr|Dask|Xarray|
|-|----|----|------|
|Formatting|-|Black|Black|
|Linting|Flake8|Flake8|Flake8|
|Imports|-|-|isort|
|Type checking|-|-|Mypy|
|Code coverage|Coveralls|Coveralls|Codecov|
|CI|Travis|Travis|Azure Pipelines|

I would like to propose we use:
* [Black](https://black.readthedocs.io/en/stable/)
* [Flake8](https://flake8.pycqa.org/en/latest/)
* [isort](https://timothycrosley.github.io/isort/)
* [Mypy](https://mypy.readthedocs.io/en/stable/)
* [Coveralls](https://coveralls.io/)
* [GitHub Actions](https://help.github.com/en/actions)

GitHub Actions is the only one that isn't used by any of the three other related projects, but it seems to be a popular choice for new projects due to its close integration with GitHub.

hammer:
I decided to learn a bit more about Codecov vs. Coveralls. Based on zero experiences with either product I would vote for Codecov, mostly because it's nearly standard across Python projects, to the point that I wonder how Dask and Zarr ended up with Coveralls!

- Coveralls is developed by [Lemur Heavy Industries](http://lemurheavy.com/), which appears to be a small consulting shop in LA. One of the founders, amusingly, appears to have been the guitarist for Capital Cities, whose [song](https://www.youtube.com/watch?v=47dtFZ8CFo8) "Safe and Sound" you've surely heard.
- Codecov is developed by a company called [Codecov](https://codecov.io/) with a CEO is in SF and CTO in Nashville.
- A brief [discussion](https://discourse.julialang.org/t/codecov-vs-coverall/27595) in the Julia community favored Codecov. [Merging reports](https://docs.codecov.io/docs/merging-reports) was cited as a feature in Codecov but not in Coveralls.
[...]

PR #6, opened 2020-06-22
Enforce coding standards

tomwhite:
This uses the tools discussed in #5, except Coveralls (it does test code coverage though).

The actual code and the test are just for illustration and would be replaced by something more realistic later, such as that being discussed in other issues and in https://discourse.smadstatgen.org/t/xarray-conventions-for-variation-data/44/4.

The settings are intentionally strict: code coverage has to be 100%, and mypy is run with the `--strict` option. The latter may have to be relaxed in the future, but let's see how far we can get with it.

Comments and suggestions welcome of course.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,16 +6,13 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: 3.8
 
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: '3.8'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,18 +17,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt -r requirements-dev.txt
-    - name: Check imports with isort
-      run: |
-        isort -rc --check .
-    - name: Check formatting with black
-      run: |
-        black --check
-    - name: Lint with flake8
-      run: |
-        flake8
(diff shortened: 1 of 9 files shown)


tomwhite:
I think GitHub Actions needs to be enabled to run in this repo, but I don't have permission to do that. @hammer or @eric-czech can you enable it please? (We can always turn it off if we decide to use a different CI.)

eric-czech:
Maybe Jeff beat me to it?  It seems to be enabled afaict (https://github.com/pystatgen/sgkit/actions/new).

tomwhite:
Thanks @eric-czech. It actually ran in my forked repo: https://github.com/tomwhite/sgkit/runs/795587237?check_suite_focus=true.

jeromekelleher:
Looks good @tomwhite. Might be worth taking a look at pre-commit to automate some of this stuff, as I mentioned in #5.

Strict is good!

tomwhite:
Added pre-commit. I think this probably good enough to start with - we can adjust things as needed later on.

jeromekelleher:
LGTM. One nice thing about using pre-commit is that you can run it as part of CI. That way, you only have one definition of what the lint requirements are, and what you check in CI is the same as what users run on their end. This is how we do it with a [github action](https://github.com/tskit-dev/tskit/blob/7a33ceb9d123c0dc9c22bb1e430d0de866ca5e8b/.github/workflows/lint.yml#L23) in tskit.

tomwhite:
@jeromekelleher good suggestion. I've replaced the previous CI linting checks with one that uses pre-commit.

jeromekelleher:
What's the process for merging one of these PRs? Should we formalise waiting for approval for one-or-more maintainers, or what? (Maybe this is forking another process type issue?)

Anyway, I think we can merge this, as no one has complained.

tomwhite:
Thanks @jeromekelleher. Merging this now.

> What's the process for merging one of these PRs? Should we formalise waiting for approval for one-or-more maintainers, or what?

I think we should require a +1 or approval from at least one other committer/maintainer.

PR #10, opened 2020-06-30
Initial Xarray call data representation

tomwhite:
I've started to attempt to translate the discussion in https://discourse.smadstatgen.org/t/xarray-conventions-for-variation-data/44 into code.

Very much a work in progress, but comments very much appreciated.

* This is branched off #6, and passes all checks from there, as well as having 100% code coverage. I will rebase once #6 is committed.
* Dimension names are lowercase singular: "variant", "sample", "ploidy", etc.
* The mask is computed. `True` values are those that are missing - i.e. masked - in line with PyData convention.
* It's not possible to type check the array values with `mypy`, so these are checked explicitly using `check_array_like` that I've copied from skallel-tensor. (Thanks for writing that Alistair.)
* Array dimension consistency is checked by xarray, and fails with a `ValueError` if any dimensions have different lengths for different arrays.
* The order of the arguments to `create_genotype_call_dataset` might be tricky to get right, since there is a tension between grouping similar fields together and whether they are optional or not.
* The "call/GT/mask" name causes a problem when writing to Zarr, since "call/GT" is a Zarr array and a Zarr group can't be nested under it. I've changed it to "call/GT_mask" to avoid that problem.
* I haven't included metadata, or probabilities/dosages. Metadata could be added later, although it would be good to add contig names since it's only possible to store the index currently.


sgkit/api.py
```diff
@@ -0,0 +1,70 @@
...
+    variant_id: array_like, str, optional
+        The unique identifier of the variant.
+
+    Returns
+    -------
+    xr.Dataset
+        The dataset of genotype calls.
+
+    """
+    check_array_like(variant_contig, kind="i", ndim=1)
```
eric-czech, line 51: I know I mentioned 8-bit ints for this since the huge majority of datasets don't actually make use of all the alt contigs that there are now in newer reference builds but there are definitely some situations where 128 contigs wouldn't be enough.  We could just ignore that for now but I thought I'd point out that it may make sense for this function to do things like take any data type for positions, contig indexes, and calls and carefully coerce them to something more parsimonious as a convenience.

alimanfoo:
Nice. A few comments...

> * Dimension names are lowercase singular: "variant", "sample", "ploidy", etc.

For some reason I have a mild preference for plural, but not sure why. Happy to go with consensus.

> * The mask is computed. `True` values are those that are missing - i.e. masked - in line with PyData convention.

Note that a genotype call for a given variant and diploid sample may be totally missing (both alleles -1), partly missing (one allele -1, the other >=0), or totally called (both alleles >=0).

In most datasets you will only ever see totally missing or totally called genotypes, but there are some where you can get partly missing genotypes.

So one option is to leave the mask as a 3D (variants, samples, ploidy) array and make sure all downstream code using this can deal with partly missing genotypes.

Another option is to have a 2D (variants, samples) mask and make it true if totally or partly missing. I.e., treat totally and partly missing genotypes the same. 

(For reference, the [scikit-allel GenotypeArray.is_missing()](https://scikit-allel.readthedocs.io/en/stable/_modules/allel/model/ndarray.html#Genotypes.is_missing) function behaves like this.)

> * It's not possible to type check the array values with `mypy`, so these are checked explicitly using `check_array_like` that I've copied from skallel-tensor. (Thanks for writing that Alistair.)

Glad that was useful :-)

> * The "call/GT/mask" name causes a problem when writing to Zarr, since "call/GT" is a Zarr array and a Zarr group can't be nested under it. I've changed it to "call/GT_mask" to avoid that problem.

👍 

> * I haven't included metadata, or probabilities/dosages. Metadata could be added later, although it would be good to add contig names since it's only possible to store the index currently.

Adding contig names sounds like a good idea.

eric-czech:
Thanks @tomwhite!  LGTM -- I have some questions/thoughts but I'll hold off until they come up later.   I do think this will come up immediately as we get into QC though:

> Metadata could be added later, although it would be good to add contig names since it's only possible to store the index currently.

Do you have any thoughts on ways to do that other than putting metadata into attributes?

tomwhite:
Thanks for the comments!

> For some reason I have a mild preference for plural, but not sure why.

I noticed that dimensions often seem to be singular in xarray examples (e.g. "latitude"). Also "ploidy" is singular - not sure the plural sounds right in this context.

> In most datasets you will only ever see totally missing or totally called genotypes, but there are some where you can get partly missing genotypes.

Good point. Perhaps we should make it possible to specify the mask so this can be controlled. A 2D mask might need converting into a 3D one to work properly with whatever masking library is being used.

> Do you have any thoughts on ways to do that other than putting metadata into attributes?

I don't have a better idea than adding them to the metadata attributes. I'll update to allow that.

sgkit/api.py
tomwhite: The checks are actually quite loose in that they just check the dtype's kind (integer, string, boolean, etc), not its width (8 bit, 32 bit, etc). I should change the documentation to reflect that though since it currently says `np.int8`.

tomwhite:
I've updated this to store contig names in the xarray attributes. I just stored them as a list - I imagine we can add more information (contig length etc) later as needed.

Here's an example:

```
<xarray.Dataset>
Dimensions:          (allele: 2, ploidy: 2, sample: 3, variant: 2)
Dimensions without coordinates: allele, ploidy, sample, variant
Data variables:
    variant/CONTIG   (variant) int8 0 0
    variant/POS      (variant) int32 1000 2000
    variant/ALLELES  (variant, allele) |S1 b'A' b'C' b'G' b'A'
    sample/ID        (sample) <U8 'sample_1' 'sample_2' 'sample_3'
    call/GT          (variant, sample, ploidy) int8 0 0 0 1 1 0 -1 0 0 -1 -1 -1
    call/GT_mask     (variant, sample, ploidy) bool False False ... True True
    call/GT_phased   (variant, sample) bool True True False True False False
    variant/ID       (variant) <U3 'rs1' 'rs2'
Attributes:
    contigs:  ['chr1']
```

alimanfoo:
> > For some reason I have a mild preference for plural, but not sure why.
> 
> I noticed that dimensions often seem to be singular in xarray examples (e.g. "latitude"). Also "ploidy" is singular - not sure the plural sounds right in this context.

So FWIW I think I prefer plural for variants and samples because if you saw this from the xarray repr:

```
<xarray.Dataset>
Dimensions:          (alleles: 2, ploidy: 2, samples: 3, variants: 2)
```

...it kinda intuitively tells you how many samples, how many variants, and what is the ploidy.

jeromekelleher:
This looks great, thanks @tomwhite!

A high-level point: I don't think we should VCF as the template for choosing the names for things. The ALL_CAPS thing is weird for Python code, and we should call things by proper names like ``variants/position`` rather than ``variants/POS``. Using the VCF form will make people think that we are building an alternative encoding for VCF (i.e., something that will store VCF data losslessly) which we do *not* want to do, I think. We should choose names that accurately reflect the data that we are encoding. Of course there should be a natural mapping to VCF fields, but there should also be a natural mapping to PLINK and BGEN fields too.

One option would be to use the [GA4GH variants API](https://ga4gh-schemas.readthedocs.io/en/latest/schemas/variants.proto.html#protobuf.Variant) as a rough guide to what we might call things. Some very smart people (Heng Li, Richard Durbin, ...) were involved in designing this, and it was intended to be the successor to VCF, for a while. So, we could reuse some of the hard work that went in there, by cherry picking out some of the field names that are in there to reduce the work we have to do here.

Note that some of the modelling in there is pretty weird (CallSets are a pretty strange concept, IIRC), so I'm not suggesting for a second that we use all of it. But there are good ideas in there that could be reused.

alimanfoo:
> A high-level point: I don't think we should VCF as the template for choosing the names for things. The ALL_CAPS thing is weird for Python code, and we should call things by proper names like `variants/position` rather than `variants/POS`.

FWIW the ALL_CAPS thing does make me feel like I'm being shouted at, I like the suggestion to use lower case and proper names at least for the reserved fields we define in our spec.

Looking at the GA4GH API and wondering if we need `start` and `end` coordinates, or maybe `position` and `end`. When we come to implementing genome region queries we'll need to make sure those work properly with deletions. But maybe that's thinking too far ahead, and we can just start with `position`. 

Also noting that the GA4GH API is 0-based, but scikit-allel stuck with the VCF convention of being 1-based, that seemed more user-friendly.

jeromekelleher:
> Looking at the GA4GH API and wondering if we need start and end coordinates, or maybe position and end. When we come to implementing genome region queries we'll need to make sure those work properly with deletions. But maybe that's thinking too far ahead, and we can just start with position.

That's a good question. Will we be working with indels soon? If so, then maybe it's better to grasp the "interval-ness" nettle at the start and use ``(start, end)`` with perhaps ``end`` being optional and assumed to be ``start + 1`` if it doesn't exist. If not, then ``position`` is a good option.

> Also noting that the GA4GH API is 0-based, but scikit-allel stuck with the VCF convention of being 1-based, that seemed more user-friendly.

I'd hoped we could sit on the fence here and say that the coordinate system is whatever the user wants it to be. We're not trying to be a normative interchange standard here, just a useful data format that lets users get their computation done. Enforcing either choice would cause more pain to users than not, I think. (In practise, we can't enforce it anyway, and even if we say "positions are zero based" some will just put in one-based coordinates anyway. We have no way of knowing, so let's not bother trying.)

tomwhite:
Here's a rebased version that uses plural dimension names (I'm not sure which I prefer now!), and avoids ALL_CAPS in favour of more expressive lowercase names ("position", "genotype").

```
<xarray.Dataset>
Dimensions:               (alleles: 2, ploidy: 2, samples: 3, variants: 2)
Dimensions without coordinates: alleles, ploidy, samples, variants
Data variables:
    variant/contig        (variants) int8 0 0
    variant/position      (variants) int32 1000 2000
    variant/alleles       (variants, alleles) |S1 b'A' b'C' b'G' b'A'
    sample/id             (samples) <U8 'sample_1' 'sample_2' 'sample_3'
    call/genotype         (variants, samples, ploidy) int8 0 0 0 1 ... -1 -1 -1
    call/genotype_mask    (variants, samples, ploidy) bool False False ... True
    call/genotype_phased  (variants, samples) bool True True ... False False
    variant/id            (variants) <U3 'rs1' 'rs2'
Attributes:
    contigs:  ['chr1']
```

> We're not trying to be a normative interchange standard here, just a useful data format that lets users get their computation done.

Well said!

sgkit/api.py
```diff
@@ -0,0 +1,80 @@
...
+    check_array_like(variant_position, kind="i", ndim=1)
+    check_array_like(variant_alleles, kind="S", ndim=2)
+    check_array_like(sample_id, kind="U", ndim=1)
+    check_array_like(call_genotype, kind="i", ndim=3)
+    data_vars: Mapping[Hashable, Any] = {
+        "variant/contig": ([DIM_VARIANT], variant_contig),
```
jeromekelleher, line 60: Shouldn't this be 
```python
f"{DIM_VARIANT}/contig": (...)
```
etc?

alimanfoo: I think there is (or can be) a distinction between the prefixes that we use for the different arrays, and the names of the dimensions.

E.g.,, we could have arrays named like "variant/position", "variant/id", "sample/id", "call/GT", etc.

And we could have dimension names like "variants", "samples", "ploidy", etc.

There is an implicit convention here that:

* arrays with the prefix "variant/" can be 1D or 2D but all have dimensions (variants,) or (variants, foo) where foo is some other dimension than samples
* arrays with the prefix "sample/" are all 1D and have dimensions (samples,) 
* arrays with the prefix "call/" are all 2D or 3D and have dimensions (variants, samples) or (variants, samples, bar) where bar can be ploidy or something else.

alimanfoo:
> I'd hoped we could sit on the fence here and say that the coordinate system is whatever the user wants it to be. We're not trying to be a normative interchange standard here, just a useful data format that lets users get their computation done. Enforcing either choice would cause more pain to users than not, I think. (In practise, we can't enforce it anyway, and even if we say "positions are zero based" some will just put in one-based coordinates anyway. We have no way of knowing, so let's not bother trying.)

Fair point, though we may want to have a preference. E.g., it would be helpful if all ingest tools coming from the main formats (VCF, plink, bgen) use the same convention. Documentation we create (tutorials, examples) should probably be consistent too.

Also on a technical note, if we decide to support intervals and allow both closed and half-open, then our region query functions will need to be aware of that somehow.

alimanfoo:
> Here's a rebased version that uses plural dimension names (I'm not sure which I prefer now!), and avoids ALL_CAPS in favour of more expressive lowercase names ("position", "genotype").

```
<xarray.Dataset>
Dimensions:               (alleles: 2, ploidy: 2, samples: 3, variants: 2)
Dimensions without coordinates: alleles, ploidy, samples, variants
Data variables:
    variant/contig        (variants) int8 0 0
    variant/position      (variants) int32 1000 2000
    variant/alleles       (variants, alleles) |S1 b'A' b'C' b'G' b'A'
    sample/id             (samples) <U8 'sample_1' 'sample_2' 'sample_3'
    call/genotype         (variants, samples, ploidy) int8 0 0 0 1 ... -1 -1 -1
    call/genotype_mask    (variants, samples, ploidy) bool False False ... True
    call/genotype_phased  (variants, samples) bool True True ... False False
    variant/id            (variants) <U3 'rs1' 'rs2'
Attributes:
    contigs:  ['chr1']
```

FWIW this looks very good to my eyes. Only question is whether variables can be sorted lexically in the repr so all with the same prefix appear together, that's very helpful when visually scanning through which variables are present.

sgkit/api.py
jeromekelleher: OK, I'm happy to go with whatever the usual conventions are.

jeromekelleher:
> Fair point, though we may want to have a preference. E.g., it would be helpful if all ingest tools coming from the main formats (VCF, plink, bgen) use the same convention. Documentation we create (tutorials, examples) should probably be consistent too.

Fair enough. I'd say we stay agnostic as far as possible though.

> Also on a technical note, if we decide to support intervals and allow both closed and half-open, then our region query functions will need to be aware of that somehow.

We probably don't need to worry about this for a bit, but I'd argue strongly for supporting only half-open intervals. It's too confusing to have both.

eric-czech:
Is there anything left to change for this?  I have some other changes I'd like to PR on top of this but I wanted to have it merged first.

tomwhite:
I think this is a good starting point for building on. There seems to be broad agreement on this now (correct me if I'm wrong!), so we should be able to merge this.

eric-czech:
thanks guys!

PR #32, opened 2020-07-09
Add mypy type checking to GitHub action

### LINKED ISSUE #27: Mypy/Typing check
Current master doesn't pass mypy check:

```
(sgkit-dev) ➜  sgkit git:(master) ✗ mypy --strict .
setup.py:2: error: Skipping analyzing 'setuptools': found module but no type hints or library stubs
sgkit/utils.py:4: error: Function is missing a type annotation
sgkit/tests/test_utils.py:2: error: Skipping analyzing 'pytest': found module but no type hints or library stubs
sgkit/tests/test_utils.py:2: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
sgkit/tests/test_utils.py:9: error: Call to untyped function "check_array_like" in typed context
sgkit/tests/test_utils.py:12: error: Call to untyped function "check_array_like" in typed context
sgkit/tests/test_utils.py:14: error: Call to untyped function "check_array_like" in typed context
sgkit/tests/test_utils.py:16: error: Call to untyped function "check_array_like" in typed context
sgkit/tests/test_utils.py:18: error: Call to untyped function "check_array_like" in typed context
sgkit/tests/test_utils.py:20: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:55: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:56: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:57: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:58: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:59: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:72: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:73: error: Unsupported target for indexed assignment ("Mapping[Hashable, Any]")
sgkit/api.py:78: error: Call to untyped function "check_array_like" in typed context
sgkit/api.py:79: error: Unsupported target for indexed assignment ("Mapping[Hashable, Any]")
sgkit/api.py:81: error: Argument "attrs" to "Dataset" has incompatible type "Dict[str, List[str]]"; expected "Mapping[Hashable, Any]"
Found 19 errors in 4 files (checked 7 source files)
```

It goes without saying that typing provides huge benefits (especially over time), there are also many typing limitations in the python (data) ecosystem. Currently our build and pre-commit don't run mypy check, there isn't much code in this repo, but there is a bunch coming in soon. So some questions about this:
 * do we want to enable mypy check for both build and pre-commit right now (I'm +1)
 * we likely won't be able to type everything, so we need to agree which parts need to be typed and which part can be omitted

tomwhite:
I couldn't see a way of getting [mypy precommit](https://github.com/pre-commit/mirrors-mypy) to use the settings in `setup.cfg`, so instead I added a separate GH step to run mypy. This also means you have to run mypy yourself manually on your machine to check types.

Fixes #27.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,9 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
+    - name: Check types with mypy
+      run: |
+        mypy --strict .
     - name: Test with pytest and coverage
       run: |
         pytest -v --cov=sgkit
--- a/sgkit/utils.py
+++ b/sgkit/utils.py
@@ -5,7 +5,7 @@ import numpy as np
 
 def check_array_like(
     a: Any,
-    dtype: Union[None, str, Set[str]] = None,
+    dtype: Any = None,
     kind: Union[None, str, Set[str]] = None,
     ndim: Union[None, int, Set[int]] = None,
 ) -> None:
--- a/sgkit/utils.py
+++ b/sgkit/utils.py
@@ -2,10 +2,12 @@ from typing import Any, Set, Union
 
 import numpy as np
 
+from .typing import DType
+
 
 def check_array_like(
     a: Any,
-    dtype: Any = None,
+    dtype: DType = None,
     kind: Union[None, str, Set[str]] = None,
     ndim: Union[None, int, Set[int]] = None,
 ) -> None:
--- a/sgkit/utils.py
+++ b/sgkit/utils.py
@@ -1,7 +1,14 @@
+from typing import Any, Set, Union
+
 import numpy as np
 
 
-def check_array_like(a, dtype=None, kind=None, ndim=None):
+def check_array_like(
+    a: Any,
+    dtype: Union[None, str, Set[str]] = None,
+    kind: Union[None, str, Set[str]] = None,
+    ndim: Union[None, int, Set[int]] = None,
+) -> None:
     array_attrs = "ndim", "dtype", "shape"
     for k in array_attrs:
         if not hasattr(a, k):
--- a/README.md
+++ b/README.md
@@ -34,3 +34,5 @@ You can use [pre-commit](https://pre-commit.com/) to check or enforce the coding
 ```bash
 pre-commit install
 ```
+
+Note that pre-commit does not run the `mypy` check, so you will need to run that manually before submitting a pull request.
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,6 +59,9 @@ line_length = 88
 
 [mypy-numpy.*]
 ignore_missing_imports = True
-
+[mypy-pytest.*]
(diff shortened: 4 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
+    - name: Check types with mypy
+      run: |
+        mypy --strict .
```
ravwojdyla, line 24: I found that when mypy fails on CI, it's useful to get html report via `--html-report`, I don't know whether that would work well with github actions tho 🤔 ?

sgkit/utils.py
```diff
@@ -1,7 +1,14 @@
+from typing import Any, Set, Union
+
 import numpy as np
 
 
-def check_array_like(a, dtype=None, kind=None, ndim=None):
+def check_array_like(
+    a: Any,
+    dtype: Union[None, str, Set[str]] = None,
```
ravwojdyla, line 8: Should we also allow `numpy.dtype`?

.github/workflows/build.yml
tomwhite: That would be very useful! However, I'm not sure it's currently possible with GH actions, see https://github.community/t/publishing-test-results/16215/11. I suggest we just leave it as the text report.

sgkit/utils.py
tomwhite: Good point. The number of different ways of specifying types is quite long (see https://numpy.org/doc/stable/reference/arrays.dtypes.html#arrays-dtypes-constructing), so I think specifying Any is probably the best we can do. I noticed that xarray doesn't generally declare types for dtype arguments, and where it does they are Any.

.github/workflows/build.yml
ravwojdyla: @tomwhite oh, that not great. Anyway makes sense if that's the limitation of GH actions.

sgkit/utils.py
ravwojdyla: @tomwhite yea, I try to avoid `Any` for obvious reasons. I wonder if it would make sense to add a typing module within sgkit to start gradually defining these kind of types, so we could create an `DTypeType` (or sth like that) which right now could either be `Any` or more gradual `Union[Any, <other_types>]` (the extra types are just informative), neither way is type safe but at least it's a bit more future proof, when maybe in the future when we have a better way to define these types it would be just a matter of swapping the type definition?

tomwhite: Sounds good. I added a `DType` alias for `Any`.

hammer:
I filed https://github.com/pystatgen/sgkit/issues/39 to be sure we follow up on the issues encountered in this PR.

Issue #25, opened 2020-07-08
Genome region query API


alimanfoo:
Raising this issue to discuss API for selecting data from a given genome region, which could be either a whole contig or a contiguous region within a contig.

Breaking this out from [discussion thread here](https://github.com/pystatgen/sgkit-plink/pull/1#discussion_r451397877).

alimanfoo:
> A few options that come to mind:
> 
> * Using an accessor:
> 
> ```python
> ds.genetics.sel_contig(slice('chr1'))
> ```
> 
> * Accessor specific to functions on attributes:
> 
> ```python
> ds.sel(variant=ds['variant/contig'] == ds.meta.contig_index('chr1'))
> ```
> 
> * Use contig index directly:
> 
> ```python
> ds.sel(variant=ds['variant/contig'] == 0) # for chr1
> ```
> 
> I think it makes more sense to focus on the situation with an index though, which is a bit more clear on it's own IMO:
> 
> ```python
> contig_index = ds.attrs['contigs'].index('chr1')
> contig_range = slice(contig_index, contig_index)
> # or contig_range = slice(0, 0) for chr1
> bp_pos_range = slice(0, 1000000)
> ds.set_index(variant=('variant/contig', 'variant/pos')).sel(variant=(contig_range, bp_pos_range))
> ```
> 
> Any thoughts/preferences on those?

FWIW I don't know the best API but as a user I'd like to never have to use contig indices. That creates indirection that makes code harder to read and understand. Particularly where you have contig IDs that don't have a natural sequential order (e.g., in Anopheles we have "2L", "2R", "3L", "3R", "X"). 

Ideally I'd like a simple method or function where I can pass in three arguments: the contig name, and the start and stop positions of the region (1-based, stop-inclusive). Like the Python equivalent of a tabix query.

alimanfoo:
Btw apologies in advance if I get opinionated here or elsewhere. I just know I'm going to end up having to explain this to others, and it will really help to have an API that is just obvious and natural from a geneticist point of view. If ever I express a strong opinion please feel free to push back :smile:

alimanfoo:
> * Using an accessor:
> 
> ```python
> ds.genetics.sel_contig(slice('chr1'))
> ```

[...]

PR #36, opened 2020-07-09
Allele count implementation

eric-czech:
This is an implementation for https://github.com/pystatgen/sgkit/issues/3. 

A few notes/questions:

- I was initially going to parameterize these tests across all array backends, and possibly mixtures of them.  I decided not to though because it doesn't seem like this is the right level to be validating ufunc reduce implementations that Xarray is ultimately dispatching to.  A convention I'd like to propose is that we don't write tests for multiple backends unless we're writing code that makes special accommodations for them (you'd need to to get 100% coverage anyhow).  I don't feel very strongly about that though so let me know what you guys think.
- We definitely need a convenient way to generate Dataset instances given that they'll often be the sole input to functions.  I put a method for this in `sgkit/testing.py`, but I'm happy to move that around.  I also think at least one simple function like [simulate_genotype_call_dataset](https://github.com/eric-czech/sgkit/blob/allele_count/sgkit/testing.py#L8) should probably be a part of the public API.  It's hard to write examples/tests without it.
- I put the function in `sgkit/stats/aggregation.py` under the assumption that a lot of similarly simple axis-wise aggregation functions will eventually make sense in the same location.

### DIFF
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -3,20 +3,20 @@ from xarray import DataArray, Dataset
 
 
 def allele_count(ds: Dataset) -> DataArray:
-    """Compute allele count from genotype calls
+    """Compute allele count from genotype calls.
 
     Parameters
     ----------
     ds : Dataset
         Genotype call dataset such as from
-        `sgkit.create_genotype_call_dataset`
+        `sgkit.create_genotype_call_dataset`.
 
     Returns
     -------
-    DataArray
+    variant/allele_count : DataArray
         Allele counts with shape (variants, alleles) and values
         corresponding to the number of non-missing occurrences
-        of each allele
+        of each allele.
 
     Examples
     --------
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -25,7 +25,7 @@ def count_alleles(ds: Dataset) -> DataArray:
 
     >>> import sgkit as sg
     >>> from sgkit.testing import simulate_genotype_call_dataset
-    >>> ds = simulate_genotype_call_dataset(n_variant=3, n_sample=2, seed=1)
+    >>> ds = simulate_genotype_call_dataset(n_variant=4, n_sample=2, seed=1)
     >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
     samples 0   1
     variants
@@ -34,13 +34,11 @@ def count_alleles(ds: Dataset) -> DataArray:
     2       0/1	1/0
     3       0/0	0/0
 
-    >>> sg.allele_count(ds)
-    <xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
+    >>> sg.count_alleles(ds).values
     array([[2, 2],
-        [1, 3],
-        [2, 2],
-        [4, 0]])
-    Dimensions without coordinates: variants, alleles
+           [1, 3],
+           [2, 2],
+           [4, 0]])
     """
     # Count each allele index individually as a 1D vector and
     # restack into new alleles dimension with same order
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -1,3 +1,5 @@
+import dask.array as da
+import numpy as np
 import xarray as xr
 from xarray import DataArray, Dataset
 
@@ -42,11 +44,30 @@ def count_alleles(ds: Dataset) -> DataArray:
     """
     # Count each allele index individually as a 1D vector and
     # restack into new alleles dimension with same order
-    gt, mask = ds["call/genotype"], ds["call/genotype_mask"]
-    acs = [
-        xr.where(mask, 0, gt == i).sum(dim=("samples", "ploidy"))  # type: ignore[no-untyped-call]
-        for i in range(ds.dims["alleles"])
-    ]
-    ac = xr.concat(acs, dim="alleles")  # type: ignore[no-untyped-call]
-    ac = ac.T.rename("variant/allele_count")
-    return ac  # type: ignore[no-any-return]
+    G = ds["call/genotype"].stack(calls=("samples", "ploidy"))
+    M = ds["call/genotype_mask"].stack(calls=("samples", "ploidy"))
+    n_variant, n_allele = G.shape[0], ds.dims["alleles"]
+    max_allele = n_allele + 1
+
+    # Recode missing values as max allele index
+    G = xr.where(M, n_allele, G)  # type: ignore[no-untyped-call]
+    G = da.asarray(G)
+
+    # Count allele indexes within each block
+    CT = da.map_blocks(
+        lambda x: np.apply_along_axis(np.bincount, 1, x, minlength=max_allele),
+        G,
+        chunks=(G.chunks[0], max_allele),  # type: ignore[index]
+    )
+    assert CT.shape == (n_variant, max_allele)
+
+    # Stack the column blocks on top of each other
+    CTS = da.stack([CT.blocks[:, i] for i in range(CT.numblocks[1])])
+    assert CTS.shape == (CT.numblocks[1], n_variant, max_allele)
+
+    # Sum over column blocks and slice off allele
+    # index corresponding to missing values
+    AC = CTS.sum(axis=0)[:, :n_allele]
+    assert AC.shape == (n_variant, n_allele)
+
+    return DataArray(data=AC, dims=("variants", "alleles"), name="variant/allele_count")
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -1,0 +1,55 @@
+import xarray as xr
+from xarray import DataArray, Dataset
+
+
+def allele_count(ds: Dataset) -> DataArray:
+    """Compute allele count from genotype calls
+
+    Parameters
+    ----------
+    ds : Dataset
+        Genotype call dataset such as from
+        `sgkit.create_genotype_call_dataset`
+
+    Examples
+    --------
+
+    >>> import sgkit as sg
+    >>> from sgkit.testing import simulate_genotype_call_dataset
+    >>> ds = simulate_genotype_call_dataset(n_variant=3, n_sample=2, seed=1)
+    # TODO: Should we expose visual convenience functions like this?
+    >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
+    samples 0   1
+    variants
+    0       1/0	1/0
+    1       1/0	1/1
+    2       0/1	1/0
+    3       0/0	0/0
+
+    >>> allele_count(ds)
+    <xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
+    array([[2, 2],
+        [1, 3],
+        [2, 2],
+        [4, 0]])
+    Dimensions without coordinates: variants, alleles
+
+    Returns
+    -------
+    DataArray
+        Allele counts with shape (variants, alleles) and values
+        corresponding to the number of non-missing occurrences
+        of each allele
+    """
+    # Collapse 3D calls into 2D array where calls are flattened into columns
+    gt = ds["call/genotype"].stack(calls=("samples", "ploidy"))
+    mask = ds["call/genotype_mask"].stack(calls=("samples", "ploidy"))
+
+    # Count each allele index individually as a 1D vector and
+    # restack into new alleles dimension with same order
+    acs = [
+        xr.where(mask, 0, gt == i).sum(dim="calls") for i in range(ds.dims["alleles"])
+    ]
+    ac = xr.concat(acs, dim="alleles")
+    ac = ac.T.rename("variant/allele_count")
+    return ac
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -2,7 +2,7 @@ import xarray as xr
 from xarray import DataArray, Dataset
 
 
-def allele_count(ds: Dataset) -> DataArray:
+def count_alleles(ds: Dataset) -> DataArray:
     """Compute allele count from genotype calls.
 
     Parameters
@@ -32,7 +32,7 @@ def allele_count(ds: Dataset) -> DataArray:
     2       0/1	1/0
     3       0/0	0/0
 
-    >>> allele_count(ds)
+    >>> sg.allele_count(ds)
     <xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
     array([[2, 2],
         [1, 3],
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -11,13 +11,20 @@ def allele_count(ds: Dataset) -> DataArray:
         Genotype call dataset such as from
         `sgkit.create_genotype_call_dataset`
 
+    Returns
+    -------
+    DataArray
+        Allele counts with shape (variants, alleles) and values
+        corresponding to the number of non-missing occurrences
+        of each allele
+
     Examples
     --------
 
     >>> import sgkit as sg
     >>> from sgkit.testing import simulate_genotype_call_dataset
     >>> ds = simulate_genotype_call_dataset(n_variant=3, n_sample=2, seed=1)
-    # TODO: Should we expose visual convenience functions like this?
+    # TODO: Use solution for https://github.com/pystatgen/sgkit/issues/37 here instead
     >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
     samples 0   1
     variants
@@ -33,22 +40,13 @@ def allele_count(ds: Dataset) -> DataArray:
         [2, 2],
         [4, 0]])
     Dimensions without coordinates: variants, alleles
-
-    Returns
-    -------
-    DataArray
-        Allele counts with shape (variants, alleles) and values
-        corresponding to the number of non-missing occurrences
-        of each allele
     """
-    # Collapse 3D calls into 2D array where calls are flattened into columns
-    gt = ds["call/genotype"].stack(calls=("samples", "ploidy"))
-    mask = ds["call/genotype_mask"].stack(calls=("samples", "ploidy"))
-
     # Count each allele index individually as a 1D vector and
     # restack into new alleles dimension with same order
+    gt, mask = ds["call/genotype"], ds["call/genotype_mask"]
     acs = [
-        xr.where(mask, 0, gt == i).sum(dim="calls") for i in range(ds.dims["alleles"])
+        xr.where(mask, 0, gt == i).sum(dim=("samples", "ploidy"))
+        for i in range(ds.dims["alleles"])
     ]
     ac = xr.concat(acs, dim="alleles")
     ac = ac.T.rename("variant/allele_count")
--- a/sgkit/testing.py
+++ b/sgkit/testing.py
@@ -13,7 +13,15 @@ def simulate_genotype_call_dataset(
     n_contig: int = 1,
     seed: Optional[int] = None,
 ):
-    """Simulate genotype calls and variant/sample data
+    """Simulate genotype calls and variant/sample data.
+
+    Note that the data simulated by this function has no
+    biological interpretation and that summary statistics
+    or other methods applied to it will produce meaningless
+    results.  This function is primarily a convenience on
+    generating `Dataset` containers so quantities of interest
+    should be overwritten, where appropriate, within the
+    context of a more specific application.
 
     Parameters
     ----------
@@ -33,7 +41,7 @@ def simulate_genotype_call_dataset(
     Returns
     -------
     Dataset
-        Dataset from `sgkit.create_genotype_call_dataset`
+        Dataset from `sgkit.create_genotype_call_dataset`.
     """
     rs = np.random.RandomState(seed=seed)
     call_genotype = rs.randint(
--- a/sgkit/testing.py
+++ b/sgkit/testing.py
@@ -1,6 +1,7 @@
 from typing import Optional
 
 import numpy as np
+from xarray import Dataset
 
 from .api import create_genotype_call_dataset
 
@@ -12,7 +13,7 @@ def simulate_genotype_call_dataset(
     n_allele: int = 2,
     n_contig: int = 1,
     seed: Optional[int] = None,
-):
+) -> Dataset:
     """Simulate genotype calls and variant/sample data.
 
     Note that the data simulated by this function has no
@@ -31,6 +32,8 @@ def simulate_genotype_call_dataset(
         Number of samples to simulate
     n_ploidy : int
         Number of chromosome copies in each sample
+    n_allele: int
+        Number of alleles to simulate
     n_contig : int, optional
         Number of contigs to partition variants with,
         controlling values in `variant/contig`. Values
--- a/sgkit/testing.py
+++ b/sgkit/testing.py
@@ -1,0 +1,56 @@
+from typing import Optional
+
+import numpy as np
+
+from .api import create_genotype_call_dataset
+
+
+def simulate_genotype_call_dataset(
+    n_variant: int,
+    n_sample: int,
+    n_ploidy: int = 2,
+    n_allele: int = 2,
+    n_contig: int = 1,
+    seed: Optional[int] = None,
+):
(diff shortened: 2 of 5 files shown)


sgkit/testing.py
```diff
@@ -0,0 +1,56 @@
...
+
+def simulate_genotype_call_dataset(
+    n_variant: int,
+    n_sample: int,
+    n_ploidy: int = 2,
+    n_allele: int = 2,
+    n_contig: int = 1,
+    seed: Optional[int] = None,
+):
+    """Simulate genotype calls and variant/sample data
```
jeromekelleher, line 16: This is a good idea, and I agree it's a useful thing to expose to let us define examples and so on. I'd just add a note emphasising that the data isn't in any way realistic, and it shouldn't be used for anything other than testing and examples.

eric-czech: Will do! good point.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    # Collapse 3D calls into 2D array where calls are flattened into columns
+    gt = ds["call/genotype"].stack(calls=("samples", "ploidy"))
+    mask = ds["call/genotype_mask"].stack(calls=("samples", "ploidy"))
+
+    # Count each allele index individually as a 1D vector and
+    # restack into new alleles dimension with same order
+    acs = [
+        xr.where(mask, 0, gt == i).sum(dim="calls") for i in range(ds.dims["alleles"])
+    ]
```
alimanfoo, line 52: Using dims['alleles'] here is neat, avoids having to specify the number of alleles (max_allele) as in scikit-allel.

Question re the xarray implementation here, is it necessary to do the stack? Slightly surprised you can't just do a sum over both the samples and ploidy dimensions. E.g., `xr.where(mask, 0, gt == i).sum(dim=("samples", "ploidy"))?

Also just noting that there might be noticeable performance differences using the xarray approach versus writing a numba kernal, primarily because I suspect the xarray implementation will end up taking multiple passes over the genotype calls, whereas a numba kernal can just make a single pass. But don't let that block this PR, can always do some benchmarking and revisit later.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+    <xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
+    array([[2, 2],
+        [1, 3],
+        [2, 2],
+        [4, 0]])
+    Dimensions without coordinates: variants, alleles
+
+    Returns
+    -------
+    DataArray
```
alimanfoo, line 39:
```suggestion
    variant/allele_count : DataArray
```

Could put the name here?

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
+import xarray as xr
+from xarray import DataArray, Dataset
+
+
+def allele_count(ds: Dataset) -> DataArray:
+    """Compute allele count from genotype calls
```
alimanfoo, line 6:
```suggestion
    """Compute allele count from genotype calls.
```

Nit, punctuation.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+
+    Parameters
+    ----------
+    ds : Dataset
+        Genotype call dataset such as from
+        `sgkit.create_genotype_call_dataset`
```
alimanfoo, line 12:
```suggestion
        `sgkit.create_genotype_call_dataset`.
```


Nit, punctuation.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+        [2, 2],
+        [4, 0]])
+    Dimensions without coordinates: variants, alleles
+
+    Returns
+    -------
+    DataArray
+        Allele counts with shape (variants, alleles) and values
+        corresponding to the number of non-missing occurrences
+        of each allele
```
alimanfoo, line 42:
```suggestion
        of each allele.
```

Nit, punctuation.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+    """
```
alimanfoo, line 43: [Numpydoc convention](https://numpydoc.readthedocs.io/en/latest/format.html#sections) puts Returns section before Examples section.

eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/656ecea34964e2c2ca1b458287bb66930311ba98

sgkit/stats/aggregation.py
eric-czech: That works too, I switched it in https://github.com/pystatgen/sgkit/pull/36/commits/656ecea34964e2c2ca1b458287bb66930311ba98#diff-e2edcf5f78ccf0e17093a6cbe4afeda6R46.

I'll break the performance discussion out into the top-level but I completely agree.

sgkit/stats/aggregation.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

sgkit/stats/aggregation.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

sgkit/stats/aggregation.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

sgkit/stats/aggregation.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

sgkit/stats/aggregation.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/885a50d30f72d0c81ce56e01e3a13975fac9d02d

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,55 @@
...
+        `sgkit.create_genotype_call_dataset`
+
+    Examples
+    --------
+
+    >>> import sgkit as sg
+    >>> from sgkit.testing import simulate_genotype_call_dataset
+    >>> ds = simulate_genotype_call_dataset(n_variant=3, n_sample=2, seed=1)
+    # TODO: Should we expose visual convenience functions like this?
+    >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
```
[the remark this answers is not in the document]
eric-czech: I added https://github.com/pystatgen/sgkit/issues/37 but will leave it as a to-do to change after a separate PR for the viz function.  

Do you have any suggestions on how to organize things that come up in PRs that aren't critical to block on but should still be revisited at some point in the future?

eric-czech:
(breaking out the performance discussion from code comments)

> Also just noting that there might be noticeable performance differences using the xarray approach versus writing a numba kernal, primarily because I suspect the xarray implementation will end up taking multiple passes over the genotype calls, whereas a numba kernal can just make a single pass. But don't let that block this PR, can always do some benchmarking and revisit later.

I think that's true too, that dask is going to make multiple passes this way.  I thought I would keep it simple to start in case there is a more elegant way to get dask to optimize the related tasks, but here is at least one idea on an extension that would avoid multiple passes:

```python
# Call array just like 'call/genotype'
gt = np.array([
    [[0, 1], [1, 0], [1, 1]],
    [[0, 0], [1, 1], [0, 0]],
    [[0, 0], [0, 0], [0, 0]],
    [[1, 1], [1, 1], [1, 1]],
])
# Ravel calls to 1D array with original variant index
idx = np.repeat(np.arange(gt.shape[0]), gt.shape[1] * gt.shape[2])
gtf = np.stack([idx, gt.ravel()], axis=1)
# Count unique variant index + call combinations
idx, cts = np.unique(gtf, axis=0, return_counts=True)
# Pivot back out into (variant, alleles) count array
ac = np.zeros((gt.shape[0], 2), dtype=int)
ac[tuple(idx.T)] = cts
ac
array([[2, 4],
       [4, 2],
       [6, 0],
       [0, 6]])
```

I think that should be a reasonable approach in principle with dask minus caveats for multi-dimensional fancy indexing and `da.unique` along an axis.  I can imagine a few workarounds for both but in general there must be an efficient way to get counts of unique elements along an axis -- would be a good question for @mrocklin.

sgkit/testing.py
eric-czech: Changed in https://github.com/pystatgen/sgkit/pull/36/commits/5815cac9ae12c3211720acf475573ee78061a9c1

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,53 @@
...
+    <xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
+    array([[2, 2],
+        [1, 3],
+        [2, 2],
+        [4, 0]])
+    Dimensions without coordinates: variants, alleles
+    """
+    # Count each allele index individually as a 1D vector and
+    # restack into new alleles dimension with same order
+    gt, mask = ds["call/genotype"], ds["call/genotype_mask"]
```
ravwojdyla, line 46: Shouldn't these default names be parameters with defaults in the method signature? (plus the ones below)
And the expectation that `ds` should contain both genotype and genotype_mask should also be documented and also probably validated + descriptive `ValueError` if missing. This kind of `dataset` validation is likely to exists in most methods - so it would make sense to have a helper method for this so other functions can reuse the validation of inputs.

eric-czech: I'm a -1/2 on that because:

1. Our ValueError wouldn't be more informative than the xarray KeyError there
2. The thrust of https://discourse.smadstatgen.org/t/xarray-conventions-for-variation-data/44/4 and https://github.com/pystatgen/sgkit/pull/10 was to hit one representation of a Dataset that all downstream methods could make assumptions about.  My early attempts to put some kind of type around the Dataset instances that would formalize those assumptions didn't make it in, but I still think some assumptions should be allowed
3. I think mentioning that the Dataset should come from `create_genotype_call_dataset` at least makes it clear to the user what assumptions are made by the function.  Since there's not a type to refer to, it would be nice to have a better way to express that than "A dataset from `sgkit.create_genotype_call_dataset`" but I'm not sure how else to put it. 

I think most libraries can get around this easily with subclasses/wrapper types for data containers, but since we're not doing that, I'm not sure what the right way to avoid all the boilerplate is (I think there will be a ton of it for those two arrays).  [docrep](https://github.com/Chilipp/docrep) looks like a possibility I guess (example: https://github.com/xgcm/xgcm/blob/master/xgcm/grid.py#L545).  I haven't used it but have you or anybody else tried something that reduces repetition in docstrings like that?

ravwojdyla: Re: 1. Validation of user inputs is public API a common good practice, and throwing meaningful `ValueError` is also a standard over leaking errors/assumptions from the implementation. But if you are strongly against it, I won't block on this.

Re: 2. I see. Sorry to be late to the discussion - I do like the idea of encapsulation (especially over inheritance, which can be problematic over time) of Dataset/Array, but I don't want to restart that discussion, unless you believe there is a good reason to do that. I wouldn't be surprised if eventually we end up with encapsulation anyway, one pattern that I had a good experience with was sth like this:

```python

class SgKitDataset:
  def __init__(...):
    self.dataset: Dataset = ... # note this is intentionally a public member

  @classmethod
  def create_genotype_call_dataset(...):
    ...

  @classmethod
  def from_dataset(...):
    ...

  def useful_sgkit_dataset_fun(...):
    ...
```

This pattern allows to capture domain/lib specific API + expose the underlying data structure if needed for the users, overall is a good middle ground. We used this specifically for Scio: [here](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala) to encapsulated Apache Beam's/Java `PCollection` with our Scala based `SCollection`, and it worked great in that case. I don't want to block your PR on this obviously, and if you like I could give it a try to sketch the API together in a PR given the functionalities currently in the code? It would mitigate issue from 2 and 3.

eric-czech: > Re: 1. Validation of user inputs to public API is a common good practice, and throwing meaningful ValueError is also a standard over leaking errors/assumptions from the implementation. But if you are strongly against it, I won't block on this.

I think for just checking data variable presence, it's a grey area and I'd say no here but yes to that in an API like this if it happens: https://github.com/pystatgen/sgkit/issues/25#issuecomment-656093123.  But, I can definitely see the value in a set of checks specific to this Dataset that do more substantial things like check dimensions/dtypes and possibly assert things about values too (though the last one won't be great though without https://github.com/dask/dask/issues/97).  

> This pattern allows to capture domain/lib specific API + expose the underlying data structure if needed for the users (given the public field/accessor)

We wanted to avoid wrapper classes in the beginning and we talked about a bunch of other possibilities with subclasses and Xarray accessors.  My initial proposal was pretty similar to yours but tried to do some of it with [accessors](https://salem.readthedocs.io/en/latest/xarray_acc.html) instead since they behave much like a wrapper class.  Ultimately, @tomwhite was an advocate of using `Dataset`'s as-is in https://github.com/pystatgen/sgkit/pull/10 so that's the direction we've been heading.  I think it's a good one if we can iron out some of the boilerplate and I think there was also some general agreement that a validation function could be added to the top of every other function to replace what you and I are trying to do with types.  

A question I have for @tomwhite then is if a `check_genotype_call_dataset` function exists that we use at the beginning of every other analytical function, how much of the dataset do we validate?  On one hand it would be limiting to not be able to use `allele_count` because no `sample_id` variable exists (since it doesn't use it) but on the other it would be nice to enforce the same structure everywhere, like like a subclass or wrapper class would.  I'd defer to him on that.

tomwhite: I agree that subclassing is more problematic than encapsulation, and we should consider the "middle ground" approach being suggested by @ravwojdyla. The advantage would be that checks are only needed on creation, rather than before every analytic function call.

So definitely worth sketching out such an API to see what we think of it. I'll add a few thoughts to #16 since that seems to be where the API discussion is taking place.

> if a check_genotype_call_dataset function exists that we use at the beginning of every other analytical function, how much of the dataset do we validate?

I was thinking that the validation would be lightweight - basically presence of variables, and checking dtypes.

sgkit/testing.py
```diff
@@ -0,0 +1,64 @@
...
+
+    Parameters
+    ----------
+    n_variant : int
+        Number of variants to simulate
+    n_sample : int
+        Number of samples to simulate
+    n_ploidy : int
+        Number of chromosome copies in each sample
+    n_contig : int, optional
```
tomwhite, line 34: Doc `n_allele`

sgkit/tests/test_aggregation.py
```diff
@@ -0,0 +1,69 @@
...
+    ds["call/genotype_mask"] = xr.DataArray(calls < 0, dims=dims)
+    return ds
+
+
+def test_allele_count():
+    # Single-variant, single-sample case
+    ac = allele_count(get_dataset([[[1, 0]]]))
+    np.testing.assert_equal(ac, np.array([[1, 1]]))
+
+    # Multi-variant, single-sample case
```
tomwhite, line 24: Can you break each of these tests into its own test function. They are independent, and the function name could then document them. E.g. `test_allele_count__single_variant_single_sample` (not sure if double underscore in the middle there is the standard way of delimiting the function name from extra description).

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,53 @@
+import xarray as xr
+from xarray import DataArray, Dataset
+
+
+def allele_count(ds: Dataset) -> DataArray:
```
tomwhite, line 5: I think this should be called `count_alleles`. Functions are normally verbs.

sgkit/stats/aggregation.py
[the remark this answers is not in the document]
tomwhite: Doing what you did here is fine I think (creating a linked issue).

sgkit/stats/aggregation.py
eric-czech: > I was thinking that the validation would be lightweight - basically presence of variables, and checking dtypes.

What I mean is what variables should be validated?  A function like `count_alleles` only needs some of them so checking all of them is limiting for using the function.  In other words, you don't need variables like`variant/contig` and `sample/id` so why should the `check_genotype_call_dataset` look for them?  

On the other hand, validating all the variables every time maintains more consistency with how `create_genotype_call_dataset` acts like a constructor for a wrapper or subclass type.

tomwhite: I see. In that case I think you would only check for what you need.

alimanfoo: > 3\. Since there's not a type to refer to, it would be nice to have a better way to express that than "A dataset from `sgkit.create_genotype_call_dataset`" but I'm not sure how else to put it.

I'd suggest we write up a page of documentation which describes the xarray+sgkit dataset convention. That would describe what arrays (variables) and dimensions are expected, expected constraints on data types and dimensions, any naming conventions, etc.

hammer: @alimanfoo sounds great! want to file an issue and add the `documentation` label?

alimanfoo: > @alimanfoo sounds great! want to file an issue and add the `documentation` label?

Done: #42

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,53 @@
...
+    # TODO: Use solution for https://github.com/pystatgen/sgkit/issues/37 here instead
+    >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
+    samples 0   1
+    variants
+    0       1/0	1/0
+    1       1/0	1/1
+    2       0/1	1/0
+    3       0/0	0/0
+
+    >>> allele_count(ds)
```
alimanfoo, line 36:
```suggestion
    >>> sg.allele_count(ds)
```

Which reminds me, are we running doctests as part of our test suite? If not, would recommend.

alimanfoo:
> I think that's true too, that dask is going to make multiple passes this way. I thought I would keep it simple to start in case there is a more elegant way to get dask to optimize the related tasks

Just to say we could use skallel_tensor.genotypes_count_alleles() as a benchmark comparison as that uses numba kernels for both CPU and CUDA execution, would be interesting. [Here's a short ASV benchmark suite](https://github.com/scikit-allel/skallel-tensor/blob/master/benchmarks/benchmarks.py) I wrote for that and other core functions.

sgkit/stats/aggregation.py
eric-czech: https://github.com/pystatgen/sgkit/pull/36/commits/802c8d72dd1289ff4901db24755ff44ca51ec715#diff-e2edcf5f78ccf0e17093a6cbe4afeda6R5

sgkit/tests/test_aggregation.py
eric-czech: https://github.com/pystatgen/sgkit/pull/36/commits/802c8d72dd1289ff4901db24755ff44ca51ec715#diff-9468bf42c32cc3634ef3f9386adc5bacR23

sgkit/testing.py
eric-czech: https://github.com/pystatgen/sgkit/pull/36/commits/802c8d72dd1289ff4901db24755ff44ca51ec715#diff-0e928a779c495759235e889dd0a09342R35

sgkit/stats/aggregation.py
eric-czech: https://github.com/pystatgen/sgkit/pull/36/commits/802c8d72dd1289ff4901db24755ff44ca51ec715#diff-e2edcf5f78ccf0e17093a6cbe4afeda6R35

eric-czech:
Ok I made all the changes you guys suggested (@tomwhite / @alimanfoo).  Let me know if there's anything else you think I should do for a v1 of this.

eric-czech:
I take that back -- I changed the implementation a good bit based on https://github.com/dask/dask/issues/6423.

This would definitely work in a single pass and only relies on `bincount`, which is implemented in [cupy](https://docs.cupy.dev/en/stable/reference/generated/cupy.bincount.html).  Hypothetically then, using a GPU to do the counting is as simple as:

```python
import sgkit as sg
ds = ...
# Assuming these variables weren't already loaded as cupy blocks:
ds['call/genotype'].data  = ds['call/genotype'].data.map_blocks(cupy.asarray)
ds['call/genotype_mask'].data  = ds['call/genotype_mask'].data.map_blocks(cupy.asarray)
sg.count_alleles(ds)
<xarray.DataArray 'variant/allele_count' (variants: 4, alleles: 2)>
array([[2, 2],
       [1, 3],
       [2, 2],
       [4, 0]])
```

It would be great if somebody would take another look if you get the chance.

sgkit/stats/aggregation.py
```diff
@@ -0,0 +1,73 @@
...
+    >>> ds = simulate_genotype_call_dataset(n_variant=3, n_sample=2, seed=1)
+    >>> ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
+    samples 0   1
+    variants
+    0       1/0	1/0
+    1       1/0	1/1
+    2       0/1	1/0
+    3       0/0	0/0
+
+    >>> sg.allele_count(ds)
```
tomwhite, line 37: This should be `sg.count_alleles(ds)` now. When I run it I don't see the array (since it's a Dask array, not a NumPy array), just 
```
<xarray.DataArray 'variant/allele_count' (variants: 3, alleles: 2)>
dask.array<getitem, shape=(3, 2), dtype=int64, chunksize=(3, 2), chunktype=numpy.ndarray>
Dimensions without coordinates: variants, alleles
```

Also, the output from the unstack should have 3 variants, not 4.

eric-czech: Whoops, ty!  Will change that soon.

eric-czech:
One more note on the differences between this and the original:

To recap, the original looped over the alleles and counted them individually.  This new version counts them in a single pass instead.  What wasn't clear to me before is exactly how Dask would define "multiple passes".  I ran a quick experiment to see what the original task graph would have looked like to see if it would have meant multiple passes over the entire genotypes array (which would be very bad) or multiple passes over the arrays for each chunk.  It turns out that it's the latter: https://nbviewer.jupyter.org/gist/eric-czech/e96a5838a52702c20d9e56139bda7760

My guess would be that this would make little or no difference vs a `bincount`-based solution for a small number of alleles, but avoiding traversing the chunks multiple times would save a lot of memory and likely be a good bit faster.  Having ASV set up like @alimanfoo mentioned would have been a nice way to know for sure.

sgkit/stats/aggregation.py
eric-czech: https://github.com/pystatgen/sgkit/pull/36/commits/52f9d5598e2ea66f834733ca3fee21ea14ec89da

tomwhite: I don't want to hold this up, but it would be easy to enable doctest testing by adding the following to `setup.cfg`.

```
[tool:pytest]
addopts = --doctest-modules
```

We can do that in another PR though if you like.

eric-czech: I'll give it a shot and see what happens

eric-czech: I had to add `# doctest: +NORMALIZE_WHITESPACE` to get the outputs to match but it worked!  I'm a little hesitant to say that's a good global setting since I know they only use it when necessary in the dask codebase so presumably that option is better left to individual lines like this.

Changes in: https://github.com/pystatgen/sgkit/pull/36/commits/ea0185ec4684a3591f554522186166474e9218db

tomwhite: +1 this looks great

Issue #21, opened 2020-07-08
Genotype call array to dosage


alimanfoo:
Raising this issue to discuss implementing a function to convert a [genotype call array](https://discourse.smadstatgen.org/t/n-d-array-conventions-for-variation-data/16/4?u=alimanfoo) to a dosage array.

In general, the output array should have at least two dimensions, with the first two dimensions being (variants, samples). The array elements give the dosage of each allele, i.e., how many copies of an allele are carried by the individual. 

Some questions for discussion
* How do we handle biallelic and multiallelic variants?
* How do we handle missing genotype calls and avoid creating a bias towards a particular allele (e.g., the reference allele)?
* What should the output dtype be? (int or float or either)? (Some programs think of dosage as a continuous variable.)

Some related functions (not necessarily what we want to copy, but for reference):
* scikit-allel has [to_n_ref](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#allel.Genotypes.to_n_ref), [to_n_alt](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#allel.Genotypes.to_n_alt), [to_allele_counts](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#allel.Genotypes.to_allele_counts)
* skallel prototype has [genotypes_3d_to_allele_counts](https://github.com/scikit-allel/skallel-tensor/blob/eddd4b1760d9249b60b996730381318dd262c048/src/skallel_tensor/numpy_backend.py#L292), [genotypes_3d_to_allele_counts_melt](https://github.com/scikit-allel/skallel-tensor/blob/eddd4b1760d9249b60b996730381318dd262c048/src/skallel_tensor/numpy_backend.py#L313), [genotypes_3d_to_major_allele_counts](https://github.com/scikit-allel/skallel-tensor/blob/eddd4b1760d9249b60b996730381318dd262c048/src/skallel_tensor/numpy_backend.py#L334)

alimanfoo:
Perhaps this issue should be merged with #3?

tomwhite:
In #38 I have created an implementation that
- Is biallelic only
- Represents missing calls with NaN in `call/dosage` and a corresponding mask in `call/dosage_mask`
- Uses float types

hammer:
> Is biallelic only

I don't want to complicate things too early but as WES/WGS becomes more common than genotyping arrays, multiallelic sites become more common.

[...]

PR #38, opened 2020-07-13
Add create_genotype_dosage_dataset

tomwhite:
A dosage representation is needed for https://github.com/pystatgen/sgkit-bgen/pull/1 - this is an example to help discussion in #21

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -4,6 +4,7 @@ from .api import (  # noqa: F401
     DIM_SAMPLE,
     DIM_VARIANT,
     create_genotype_call_dataset,
+    create_genotype_dosage_dataset,
 )
 from .stats.association import gwas_linear_regression
 
@@ -13,5 +14,6 @@ __all__ = [
     "DIM_SAMPLE",
     "DIM_VARIANT",
     "create_genotype_call_dataset",
+    "create_genotype_dosage_dataset",
     "gwas_linear_regression",
 ]
--- a/sgkit/api.py
+++ b/sgkit/api.py
@@ -1,5 +1,6 @@
 from typing import Any, Dict, Hashable, List
 
+import numpy as np
 import xarray as xr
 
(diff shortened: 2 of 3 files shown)


alimanfoo:
Just wondering, do we need separate concepts of "genotype call dataset" and "genotype dosage dataset", or would it be more natural to have a single concept (e.g., "variation dataset") which could contain either a genotype call array or a genotype dosage array (or a genotype likelihoods array or a genotype probabilities array)?

tomwhite:
The reason I added a separate function was so it didn't have to check that one of the required arrays was supplied. Could unify into one though.

eric-czech:
On the call today I think was trying to communicate the same thing as @alimanfoo's "variation dataset" idea where this could be common point of convergence for many workflows:

```python
create_genotype_dataset(
  ...,
  # From VCF.GP or bgen 
  call_genotype_probability: Optional[float[VARIANTS, SAMPLES, GENOTYPES]],
  # Genotypes derived from probabilities/dosages or provided directly 
  call_genotype: Optional[int[VARIANTS, SAMPLES, PLOIDY]],
  # Dosages derived from probabilities, which would come from imputation or sequencing
  call_dosage: Optional[float[VARIANTS, SAMPLES]]
)
```

I'm not aware of many workflows that make use of genotype probabilities other than using them to create hard calls or dosages, but this would group together everything I've ever seen get used as a starting point for GWAS workflows.

If I had to pick one though, I would lean towards the single method for each genotype variable like you did already did @tomwhite since the results would be easy to merge and it takes the ambiguity out of all the optional fields.

tomwhite:
Rebased and added a test. 

> I would lean towards the single method for each genotype variable like you did already did @tomwhite since the results would be easy to merge and it takes the ambiguity out of all the optional fields.

Let's go ahead with this way - we can have a unifying function later if we feel this way becomes too cumbersome.

Issue #39, opened 2020-07-13
Fix mypy continuous integration issues


hammer:
From https://github.com/pystatgen/sgkit/pull/32:

- [x] [mypy precommit](https://github.com/pre-commit/mirrors-mypy) should use the settings in `setup.cfg`
- [ ] GH actions should publish the mypy HTML report (cf. [Publishing Test Results](https://github.community/t/publishing-test-results/16215) on the GitHub community Discourse).
- [ ] precommit mypy hook requires duplication of requirements

hammer:
@tomwhite should we file an upstream issue to inquire about making precommit use `setup.cfg`? It appears that https://github.com/asottile is the person who would know best so we could also just ask him directly.

tomwhite:
Opened https://github.com/pre-commit/mirrors-mypy/issues/22

ravwojdyla:
precommit + mypy fixed in #47, I've opened pre-commit/mirrors-mypy#23

hammer:
Do we want to adopt the workaround from https://github.com/pre-commit/pre-commit/pull/880#issuecomment-633284290?

ravwojdyla:
@hammer yea - looks like we gonna have to script it, should not be too hard, but since we have multiple repos, probably makes sense to build it as a precommit hook so it can be reused (+ open source for others). I don't think this is time sensitive, nevertheless would be good to have. I image it would be a simple check that parses requirements file and precommit config and fails if not in sync.

tomwhite:
No longer relevant as we're not currently using mypy.

PR #46, opened 2020-07-16
Proposal: standardize style checks on pre-commit

ravwojdyla:
Benefits:
 * make sure the versions of style checks are the same on all the checks (local, GH actions, etc)
 * style check tools live in a separate venv
 * less commands to run/remember

Regarding `mypy`, hopefully we can plug that into pre-commit as well sometime in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,19 +20,25 @@ pytest --cov=sgkit
 
 ### Code standards
 
-Run the following to enforce (or check) the source code adheres to our coding standards.
-
-```bash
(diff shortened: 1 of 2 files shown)


hammer:
> Regarding mypy, hopefully we can plug that into pre-commit as well sometime in the future.

ICYMI #39 tracks that work

ravwojdyla:
Will test mypy in pre-commit and if it works fine, follow up with another PR.

Issue #19, opened 2020-07-07
Documentation process and tools


hammer:
The Hitchhiker's Guide to Python [suggests](https://docs.python-guide.org/writing/documentation/) [Sphinx](https://www.sphinx-doc.org/en/master/) to author documentation and [Read the Docs](https://docs.readthedocs.io/en/stable/) to host it.

There have been some new developments in this domain recently, so let's evaluate our options before selecting a process and tools.

hammer:
Sphinx appears to use [reStructuredText](https://docutils.sourceforge.io/rst.html) as its markup language and [Docutils](https://docutils.sourceforge.io/) to parse and generate documents.

I found [Who needs pandoc when you have Sphinx?](https://speakerdeck.com/stephenfin/who-needs-pandoc-when-you-have-sphinx?slide=12) to be a useful description of what Sphinx adds to Docutils.

hammer:
Another approach is to use [MkDocs](https://www.mkdocs.org/) to author and generate documents and then GitHub pages or Google Cloud to host the docs.

https://github.com/scifabric/docs.pybossa.com is an example of this approach.

hammer:
It appears that offline documentation browsers like [Zeal](https://zealdocs.org/), [Dash](https://kapeli.com/dash), and [DevDocs](https://devdocs.io/) are popular. I don't use any of these, but if it's not too hard, we should ensure our docs can be imported into those tools.

hammer:
[NumPy’s Documentation](https://numpy.org/doc/stable/docs/index.html) has information on how NumPy does documentation, including [numpydoc](https://github.com/numpy/numpydoc/), their Sphinx extensions.

hammer:
Some other options
- [Docusaurus](https://docusaurus.io/) from Facebook seems to have a lot of flexibility for generating nice documentation sites. I'm not sure we have enough frontend development skills to make this tool a reasonable choice.
- [GitBook](https://www.gitbook.com/#ProductsDocs) (disclosure: I'm an investor) is another option if we want to make a fancy site.

jeromekelleher:
Sphinx is great, I think we'd want some compelling motivation to use anything else tbh. I don't have any real opinions in terms of themes and hosting.

[...]

PR #51, opened 2020-07-20
Set up Sphinx documentation

tomwhite:
This is the initial work to set up docs for the project, following discussion in #19.

- Uses Sphinx reStructuredText
- Currently very barebones, but shows the two top-level functions
- Uses the ReadTheDocs theme
- Hosted on GitHub pages at https://pystatgen.github.io/sgkit/ (at least for the moment)
- Docstring style is numpydoc, but Google-style docstrings are supported due to use of [sphinx.ext.napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html)
- Xarray types in the documentation are automatically converted to links (using [sphinx.ext.intersphinx ](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html)). Note that these types must be specified using reST, e.g. ``:class:`xarray.Dataset` ``

To generate and view docs locally:

```bash
cd docs
make html
open _build/html/index.html
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
-    - name: Check for doc warnings
+    - name: Check for Sphinx doc warnings
       run: |
         cd docs
         make html SPHINXOPTS="-W --keep-going"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,11 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
+    - name: Check for doc warnings
+      uses: ammaraskar/sphinx-action@master
+      with:
+        build-command: "make html SPHINXOPTS='-W --keep-going'"
+        docs-folder: "docs/"
     - name: Test with pytest and coverage
       run: |
         pytest -v --cov=sgkit --cov-report=term-missing
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,10 +20,9 @@ jobs:
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
     - name: Check for doc warnings
-      uses: ammaraskar/sphinx-action@master
-      with:
-        build-command: "make html SPHINXOPTS='-W --keep-going'"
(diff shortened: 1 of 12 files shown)


jeromekelleher:
LGTM. What's the plan for deploying to gh-pages?

tomwhite:
> What's the plan for deploying to gh-pages?

Use a GitHub Action to deploy to the gh-pages branch on every commit to master. I think that's the simplest thing to do until we have our first release and want different versions.

hammer:
Amazing, thanks @tomwhite!

What are your thoughts on how we could incorporate docs from related packages like `sgkit-plink` and `sgkit-bed`?

hammer:
@tomwhite do we need a follow up issue for that or is it easy enough to just incorporate into this PR and document for others?

tomwhite:
I think it needs some extra work, so I've opened #54

tomwhite:
I fixed a broken link in the generated doc at https://pystatgen.github.io/sgkit/

tomwhite:
I made changes so that the build will fail if Sphinx documentation warnings are introduced. I've also added a workflow to publish the docs to gh-pages when anything is committed to master (but I can't fully test that until this is merged).

Reviews and suggestions welcome!

tomwhite:
I've rebased this. Any objections to merging this? It will allow us to press on with writing docs.

.github/workflows/build.yml
```diff
@@ -19,6 +19,10 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
+    - name: Check for Sphinx doc warnings
+      run: |
+        cd docs
+        make html SPHINXOPTS="-W --keep-going"
```
jeromekelleher, line 25: This isn't a blocker on merging, but any particular reason for not making this strict? In my experience people writing docs will make lots of mistakes, and the only way to catch these is to have strict checking on and to verify it as part of CI. Otherwise the docs get cruddy pretty quickly when less experienced developers start contributing.

jeromekelleher:
I think we should merge this ASAP - having some functional docs infrastructure is vital at this point.

tomwhite:
Thanks for the review @jeromekelleher

.github/workflows/build.yml
tomwhite: We could enable "nit-picky mode" (`-n`). I tried doing this, but it generated some errors that will need a bit more work to fix. I've opened https://github.com/pystatgen/sgkit/issues/86

Issue #50, opened 2020-07-17
Implement REGENIE


eric-czech:
As a genetic association modeling framework, [REGENIE](https://github.com/rgcgithub/regenie) has a lot of very desirable properties for us.  It's one of very few that I've seen that addresses LD, sample correlation, multiple phenotypes and proximal contamination in a way that does not require iterative algorithms or explicit computation of either a kinship matrix or it's eigendecomposition (or some approximation of them).

The initial goal of REGENIE, much like other more recent frameworks, is to first fit a polygenic model across the genome (or perhaps all chromosomes but one) as a means to create accurate phenotype predictions.  These predictions are then subtracted from the original values and p-values are then generated based on those residuals -- which bridges the gap between ML modeling and frequentist sampling statistics.  REGENIE's choice for an ML model is simply a ridge regressor stack and though some of the details are tricky, there are some nice computational shortcuts that go into making the cross validation for this efficient and expressible through only common linear algebra operators.

I wasn't sure how well the [Glow implementation](https://glow.readthedocs.io/en/latest/tertiary/whole-genome-regression.html) would map to the Dask array API but I gave it a shot in [regenie.py](https://github.com/eric-czech/sgkit/blob/regenie/sgkit/stats/regenie.py) and managed to get nearly perfect matches to Glow results using a small simulated dataset.  

If anybody has any general style suggestions for that code, I'd definitely appreciate them now before it grows much more. 
 Otherwise I'll keep chugging away on it and see what it leads to.

----

Side note: It would be very difficult (and definitely awkard) to write a numpy-only backend for this because the algorithm isn't actually defined for single chunk arrays.  The cross validation is tightly coupled to the block sizes and you can't do it with <2 folds so a numpy version of this would need to work on at least 4D, and maybe as many as 8D arrays in some sections, to handle all the blockwise aggregations.

tomwhite:
This looks great @eric-czech!

I haven't looked in detail, but it would be good to map this implementation to the original as closely as possible using the same variable names, and links to the original paper at as fine-grained a level as is possible. I like the practice you're using of adding assertions after every major linear algebra operation too.

tomwhite:
Fixed in #66

PR #52, opened 2020-07-20
Extend GWAS linear regression for multiple phenotypes

eric-czech:
This PR contains some fairly minor adjustments/improvements to https://github.com/pystatgen/sgkit/pull/16 that allow for regression against multiple phenotypes.  Changes made:

- Shifts outcomes into batch dimensions of the algebra functions
- Makes the internal `linear_regression` function no longer hidden so that it can be used in other modules such as https://github.com/pystatgen/sgkit/issues/50
- Allows for covariates and traits to be specified as any combination of 1 or 2D data variables, all of which are concatenated column-wise
- Forces covariates and outcomes to be chunked along the variant dimension only, i.e. this won't scale to giant (100k+) numbers of either of those things
- Introduces a convention where all Dask operations are followed immediately by assertions on shape (and chunk structure if an algorithm explicitly manipulates blocks).  I like this more for the sake of reading the code than anything else, b/c dimension size constants w/ informative names make it clear what the arrays contain.  The extra safety is great too.  Eventually, it would be nice to have a set of assertion utilities that would spit out more informative errors than these currently would.

### DIFF
--- a/sgkit/stats/association.py
+++ b/sgkit/stats/association.py
@@ -17,53 +17,85 @@ class LinearRegressionResult:
     p_value: ArrayLike
 
 
-def _gwas_linear_regression(
-    G: ArrayLike, X: ArrayLike, y: ArrayLike
+def linear_regression(
+    XL: ArrayLike, XC: ArrayLike, Y: ArrayLike
 ) -> LinearRegressionResult:
     """Efficient linear regression estimation for multiple covariate sets
 
     Parameters
     ----------
-    G : (M, N) array-like
+    XL : (M, N) array-like
         "Loop" covariates for which a separate regression will be fit to
         individual columns
-    X : (M, P) array-like
+    XC : (M, P) array-like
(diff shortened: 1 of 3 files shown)


sgkit/stats/association.py
```diff
@@ -114,20 +167,30 @@ def gwas_linear_regression(
     when not already present, but there is currently no parameterization for
     intercept-free regression.
 
+    Warning: Both covariate and trait arrays will be rechunked to have blocks
```
tomwhite, line 170: You can make this a `Warning` heading so it's recognised as a numpydoc warning.

eric-czech: Ok, I moved it in https://github.com/pystatgen/sgkit/pull/52/commits/32034c384049f9553a833619c92de76a2f174361.  Maybe there's a way to make the two separate warnings break into different warning boxes in the docs?  I'm having a hard time finding good examples.

PR #58, opened 2020-07-21
Display genotypes function

### LINKED ISSUE #37: Genotype call visualization function
Understanding the content of 3D (variants, samples, ploidy) call arrays is difficult and we should add something that makes the calls more readable, like [GenotypeArray](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html?highlight=genotypearray#allel.GenotypeArray) does.

Here is a quick example within Xarray/Pandas:

```python
ds['call/genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack()
samples  0      1
variants
0        1/0	1/0
1        1/0	1/1
2        0/1	1/0
3        0/0	0/0

### PR DESCRIPTION
**Pull request by tomwhite**
Fixes #37.

The `GenotypeDisplay` class just delegates to the Pandas dataframe, but it is probably worth having the wrapper to hide the implementation, and also because it allows us to customize (or completely change) the representation in the future.

### DIFF
--- a/sgkit/display.py
+++ b/sgkit/display.py
@@ -1,0 +1,192 @@
+from typing import Any, Tuple
+
+import pandas as pd
+import xarray as xr
+
+from .typing import ArrayLike
+
+
+class GenotypeDisplay:
+    """
+    A printable object to display genotype information.
+    """
+
+    def __init__(
+        self,
+        df: pd.DataFrame,
+        shape: Tuple[int, int],
+        max_variants: int,
+        max_samples: int,
+    ):
+        self.df = df
+        self.shape = shape
+        self.max_variants = max_variants
+        self.max_samples = max_samples
+        self.pd_options = [
+            "display.max_rows",
+            self.max_variants,
+            "display.min_rows",
+            self.max_variants,
+            "display.max_columns",
+            self.max_samples,
+            "display.show_dimensions",
+            False,
+        ]
+
+    def __repr__(self) -> Any:
+        with pd.option_context(*self.pd_options):
+            if (
+                len(self.df) > self.max_variants
+                or len(self.df.columns) > self.max_samples
+            ):
+                return (
+                    self.df.__repr__()
+                    + f"\n\n[{self.shape[0]} rows x {self.shape[1]} columns]"
+                )
+            return self.df.__repr__()
+
+    def _repr_html_(self) -> Any:
+        with pd.option_context(*self.pd_options):
+            if (
+                len(self.df) > self.max_variants
+                or len(self.df.columns) > self.max_samples
+            ):
+                return self.df._repr_html_().replace(
+                    "</div>",
+                    f"<p>{self.shape[0]} rows x {self.shape[1]} columns</p></div>",
+                )
+            return self.df._repr_html_()
+
+
+def _truncate(
+    arr: xr.DataArray,
+    variants: xr.DataArray,
+    samples: xr.DataArray,
+    max_variants: int,
+    max_samples: int,
+) -> xr.DataArray:
+    """Truncate the given array so it is of size (at most) `(max_variants + 1, max_samples + 1)`.
+
+    The reason the size may exceed `(max_variants, max_samples)` is so that pandas can be used to
+    display the array as a table, and correctly truncate rows or columns (shown as ellipses ...).
+    """
+
+    n_variant = arr.sizes["variants"]
+    n_sample = arr.sizes["samples"]
+
+    if n_variant <= max_variants:
+        # slice in half and recombine below
+        v_slice_0 = slice(0, n_variant // 2)
+        v_slice_1 = slice(n_variant // 2, n_variant)
+    else:
+        v_slice_0 = slice(0, max_variants // 2 + 1)
+        v_slice_1 = slice(n_variant - max_variants // 2, n_variant)
+
+    if n_sample <= max_samples:
+        # slice in half and recombine below
+        s_slice_0 = slice(0, n_sample // 2)
+        s_slice_1 = slice(n_sample // 2, n_sample)
+    else:
+        s_slice_0 = slice(0, max_samples // 2 + 1)
+        s_slice_1 = slice(n_sample - max_samples // 2, n_sample)
+
+    grid = [
+        [
+            arr.isel(variants=v_slice_0, samples=s_slice_0),
+            arr.isel(variants=v_slice_0, samples=s_slice_1),
+        ],
+        [
+            arr.isel(variants=v_slice_1, samples=s_slice_0),
+            arr.isel(variants=v_slice_1, samples=s_slice_1),
+        ],
+    ]
+    combined: xr.DataArray = xr.combine_nested(grid, concat_dim=["variants", "samples"])  # type: ignore[no-untyped-call]
+
+    variants_subset: xr.DataArray = xr.combine_nested([variants.isel(variants=v_slice_0), variants.isel(variants=v_slice_1)], concat_dim=["variants"])  # type: ignore[no-untyped-call]
+    samples_subset: xr.DataArray = xr.combine_nested([samples.isel(samples=s_slice_0), samples.isel(samples=s_slice_1)], concat_dim=["samples"])  # type: ignore[no-untyped-call]
+    combined = combined.assign_coords(variants=variants_subset)  # type: ignore[no-untyped-call]
+    combined = combined.assign_coords(samples=samples_subset)  # type: ignore[no-untyped-call]
+
+    assert combined.shape[0] <= max_variants + 1
+    assert combined.shape[1] <= max_samples + 1
+
+    return combined
+
+
+def display_genotypes(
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jeromekelleher** [comment]
Nice! Will this cope with being called on a 100G dataset (i.e., truncate the axes)?

**Comment by tomwhite** [comment]
No 😃 

I need to just get the first n rows (and cols) from dask.

**Comment by jeromekelleher** [comment]
I think it's worth dealing with this situation from the start - this'll get called on some awful big datasets!

**Comment by hammer** [comment]
I've found [printing tibbles](https://tibble.tidyverse.org/reference/formatting.html) to be a pleasant experience, so there may be aspects of their API we can borrow.

**Comment by tomwhite** [comment]
This was more work than I expected 😄 . Pandas has nice formatting for dataframes: it handles alignment and truncation (replacing a range or rows or columns with ellipses ...), however my initial implementation wouldn't scale since it simply converted the xarray dataset to a Pandas dataframe.

The new approach selects the corners of the xarray arrays (for genotypes, mask, phasing information) and then combines them before converting to Pandas just before displaying the array. E.g.

```
samples    S0   S1  ... S998 S999
variants            ...          
0         0/0  1/0  ...  0/1  1/1
1         1/1  1/1  ...  0/1  1/1
...       ...  ...  ...  ...  ...
99998     0/1  1/1  ...  1/0  0/1
99999     1/0  1/0  ...  1/0  1/0

[100000 rows x 1000 columns]
```

So to show no more than `M` variants (rows) and `N` samples (columns) we need to pull out four corners each of size `(M/2, N/2)` (ignoring fencepost issues), and combine them into an array of shape `(M, N)`. Actually, there is a trick to create an array of shape `(M+1, N+1)` so when we ask Pandas to render it with max rows/cols `(M, N)` it truncates in a way that makes it look like the truncated large array, even though it is actually truncating a smaller array.

I've tested this with 500K rows or columns and it runs in a few seconds (most of the time was generating sample data).

There's probably scope for improving some of the xarray operations, but it would be good if someone could review this before I spend any more time on it.

**Comment by eric-czech** [comment]
Nice!  I like the approach @tomwhite.

Good to know you can trick pandas like that.  The `combine_nested` function is good to know about too -- haven't even seen that one before.  I definitely agree with the use of coordinates in there so they pass through to pandas.

This exploring/thinking got away from me (it was fun to experiment with some of these things) so I apologize if this is overly suggestive, but I have a few higher-level ideas for potential improvements and/or extensions that are probably worth considering.  This is an example I was tinkering around with:

```python
# Generate some random data
ds = simulate_genotype_call_dataset(n_variant=100000, n_sample=1000)
ds['call_genotype_phased'] = ds['call_genotype'].sum(dim='ploidy') > 0 # Illustrative nonsense

# Set whatever indexes or coords I want as a user
ds = ds.set_index(variants=('variant_contig', 'variant_position'), samples='sample_id')

# Half of max variants and max samples, respectively
mv, ms = 2, 3 
limits = dict(variants=mv, samples=ms)
slices = {k: slice(v) for k, v in limits.items()}
ds_abbr = xr.combine_nested([
    [
        # Roll all of these simultaneously along with any indexes/coords
        ds[['call_genotype', 'call_genotype_mask', 'call_genotype_phased']]
            .roll(dict(zip(limits, roll)), roll_coords=True)
            .isel(**slices)
        for roll in row
    ]
    for row in [
        [(0,   0), (0,  ms)],
        [(mv,  0), (mv, ms)]
    ]
], concat_dim=limits.keys())

# Show the original array vs the corners array:
ds['call_genotype'][..., 0].values
array([[1, 1, 0, ..., 1, 1, 1],
       [1, 0, 1, ..., 0, 1, 1],
       [0, 1, 1, ..., 1, 0, 0],
       ...,
       [1, 0, 1, ..., 1, 0, 1],
       [0, 0, 0, ..., 0, 1, 1],
       [1, 1, 1, ..., 1, 1, 1]], dtype=int8)

ds_abbr['call_genotype'][..., 0].values
array([[1, 1, 0, 1, 1, 1],
       [1, 0, 1, 0, 1, 1],
       [0, 0, 0, 0, 1, 1],
       [1, 1, 1, 1, 1, 1]], dtype=int8)
```

Then from there with the truncated dataset, I think this could be a representation that makes all the special cases easy to handle:

```python
df = ds_abbr.to_dataframe().unstack(level='ploidy')
df.head(4)
```
Screen Shot 2020-08-04 at 3 59 31 PM

Some of the row-wise logic with conditionals for is_phased, is_missing or maybe multiallelic considerations down the road seems like it would be simpler this way.  Ignoring some of the details, I mean something like this:

```python
def to_str(r):
    if r['call_genotype_mask'][0]:
        return '.'
    # --> other special cases for phasing or more alleles
    return '/'.join(r['call_genotype'].astype(str))
df.apply(to_str, axis=1).unstack('samples')
```

Screen Shot 2020-08-04 at 4 02 06 PM

One nice advantage of all of that is that it would let users decide what they want to see in the row/column indexes.  I'm particularly interested in seeing the row index `contig:position:alleles` as opposed to `variant_id`.  Perhaps this function could use coordinates/indexes already present and default to `variant_id`/`sample_id` if there aren't any?

sgkit/display.py
```diff
@@ -0,0 +1,192 @@
...
+
+    gt = _truncate(
+        ds["call_genotype"], variants, ds["sample_id"], max_variants, max_samples
+    ).astype(str)
+    missing = _truncate(
+        ds["call_genotype_mask"], variants, ds["sample_id"], max_variants, max_samples
+    )
+    calls = xr.where(missing, ".", gt)  # type: ignore[no-untyped-call]
+
+    def make_unphased_genotype(x: ArrayLike) -> str:
```
eric-czech, line 157: Is it possible for this to be anything other than a DataArray type?

sgkit/display.py
```diff
@@ -0,0 +1,192 @@
...
+            ):
+                return self.df._repr_html_().replace(
+                    "</div>",
+                    f"<p>{self.shape[0]} rows x {self.shape[1]} columns</p></div>",
+                )
+            return self.df._repr_html_()
+
+
+def _truncate(
+    arr: xr.DataArray,
```
eric-czech, line 62: Should we have a preference for qualified type annotations like this vs imports of the type at the top level?  I'm not aware of any meaningful difference, but I've been doing the latter everywhere.  I like the qualified version to avoid DataFrame conflicts so perhaps we should make that a standard.

tomwhite:
Thanks for all the ideas @eric-czech, they were very helpful! I've now generalised the truncate function so it works on a dataset (not an array), and so the dimensions can be specified. The `roll` operation is a neat way of doing truncation, so I'm now using that too, although there are some extra complications if only one dimension needs truncating. I've also implemented your suggestion to use existing indexes and only set them if they haven't been set by the user.

sgkit/display.py
tomwhite: I don't have a strong view on this.

sgkit/display.py
tomwhite: This code no longer exists in the latest commit.

eric-czech:
Awesome LGTM!

One last thought I had is that setting an index pulls the values into memory.  Setting coordinates instead is one way to avoid that though it's been unclear to me exactly when Xarray will force those into memory too.  Assuming the `roll` and `isel` commands don't do it, then I *think* this might be a way to treat everything as chunked arrays and use a default index in the dataframe at the end that doesn't change the order:

```python
ds = simulate_genotype_call_dataset(n_variant=100000, n_sample=1000, seed=0)
ds = ds.assign(
    variant_idx=('variants', da.arange(ds.dims['variants'])),
    sample_idx=('samples', da.arange(ds.dims['samples']))
).set_coords(['variant_idx', 'sample_idx'])
ds_abbr = ... # truncate to corners
# Move coordinates to index now that they're smaller
df = ds_abbr.set_index(variants='variant_idx', samples='sample_idx')\
  .to_dataframe().unstack(level='ploidy')
df.apply(to_str, axis=1).unstack('samples')
```
Screen Shot 2020-08-06 at 8 45 35 AM

Although maybe an even better way in a case where you might not want to use any index/coords at all would be:

```python
ds = simulate_genotype_call_dataset(n_variant=100000, n_sample=1000, seed=0)
# Skip .set_coords or .set_index
# ... roll and convert to df as before
df.apply(to_str, axis=1).unstack('samples')
```
Screen Shot 2020-08-06 at 8 45 39 AM

The default indexes don't make as much sense but they would at least be in the original sort order and perhaps it's easy enough to just override them at the end so it reflects the trailing indexes of the corners (997, 998, etc.)?

I certainly wouldn't block on those things since I think letting the user choose the index is nice and flexible, but some mode for no index that doesn't impose a sort and keeps everything chunked could be a good future extension.

tomwhite:
Thanks @eric-czech for the further suggestions. I've opened https://github.com/pystatgen/sgkit/issues/97 to investigate these improvements, and I'll merge this PR in the meantime.

Issue #28, opened 2020-07-08
Add HWE test implementation


eric-czech:
This is a somewhat critical function for UKB GWAS and a very important one for pipelines that start further upstream.  

The API as I have it sketched out in [stats/hwe.py](https://github.com/eric-czech/sgkit/blob/hwe/sgkit/stats/hwe.py#L124) works by taking a dataset and producing a dataset with a single variable:

```python
call_genotype = simulate_genotype_calls(...)
ds = to_genotype_call_dataset(call_genotype)
p = sgkit.hardy_weinberg_test(ds)["variant/hwe_p_value"]
```

This is currently blocked by https://github.com/pystatgen/sgkit/pull/16 and to some extent https://github.com/pystatgen/sgkit/issues/3.

eric-czech:
Done in https://github.com/pystatgen/sgkit/pull/76

PR #76, opened 2020-07-29
HWE Test Implementation

eric-czech:
This PR contains a new implementation for https://github.com/pystatgen/sgkit/issues/28.

Summary of changes:

- Introduces numba as a dependency
- Adds a `sgkit/stats/hwe.py` module
- Adds test data exported from a reference implementation
  - I will add the code that generates the export in a separate PR 
- Wraps jit-compiled functions as separate variables and tests them where appropriate, which makes it possible to handle these scenarios:
    - A part of the unit tests is ensuring that results don't overflow with large numbers, but the speed of the test is dependent on the size of those numbers so the no-jit function should never be run with these same inputs
    - Codecov can't follow execution of the compiled functions and simply thinks they're not covered, so a limited set of tests can be run with the no-jit function to hit coverage thresholds

There may be better ways to organize jit-compiled function tests, see https://github.com/pystatgen/sgkit/issues/77.


sgkit/stats/association.py
```diff
@@ -164,24 +164,13 @@ def gwas_linear_regression(
...
-    --------
-    Regression statistics from this implementation are only valid when an
-    intercept is present. The `add_intercept` flag is a convenience for adding one
-    when not already present, but there is currently no parameterization for
-    intercept-free regression.
-
-    Additionally, both covariate and trait arrays will be rechunked to have blocks
-    along the sample (row) dimension but not the column dimension (i.e.
-    they must be tall and skinny).
-
```
tomwhite, line 184: Was this warning removed for a reason?

sgkit/stats/hwe.py
```diff
@@ -0,0 +1,178 @@
...
+    p = np.empty(n, dtype=np.float64)
+    for i in range(n):
+        p[i] = hardy_weinberg_p_value_jit(obs_hets[i], obs_hom1[i], obs_hom2[i])
+    return p
+
+
+hardy_weinberg_p_value_vec_jit = njit(hardy_weinberg_p_value_vec, fastmath=True)
+
+
+def hardy_weinberg_test(
```
tomwhite, line 122: This function should be added to the top-level `__init__.py` so it's a part of the public API.

Also, +1 on the name (even though it's not a verb!). I think using the abbreviation `hwe` _internally_ is fine too (as you have done).

sgkit/stats/hwe.py
```diff
@@ -0,0 +1,178 @@
...
+) -> ndarray:
+    arrs = [obs_hets, obs_hom1, obs_hom2]
+    if len(set(map(len, arrs))) != 1:
+        raise ValueError("All arrays must have same length")
+    if list(set(map(lambda x: x.ndim, arrs))) != [1]:
+        raise ValueError("All arrays must be 1D")
+    n = len(obs_hets)
+    p = np.empty(n, dtype=np.float64)
+    for i in range(n):
+        p[i] = hardy_weinberg_p_value_jit(obs_hets[i], obs_hom1[i], obs_hom2[i])
```
tomwhite, line 115: Observation: we're using Dask for parallelization here (across blocks) which is fine, but we may want to consider numba's `parallel=True` in the future for this loop as another dispatch option.

sgkit/stats/hwe.py
```diff
@@ -0,0 +1,178 @@
...
+        raise NotImplementedError("HWE test only implemented for diploid genotypes")
+    if ds.dims["alleles"] != 2:
+        raise NotImplementedError("HWE test only implemented for biallelic genotypes")
+    # Use precomputed genotype counts if provided
+    if genotype_counts is not None:
+        obs = list(da.asarray(ds[genotype_counts]).T)
+    # Otherwise compute genotype counts from calls
+    else:
+        # TODO: Use API genotype counting function instead, e.g.
+        # https://github.com/pystatgen/sgkit/issues/29#issuecomment-656691069
```
tomwhite, line 172: Can you open an issue to track this please.

sgkit/tests/test_hwe.py
```diff
@@ -0,0 +1,160 @@
...
+    p_expected = df["p"].values
+    p_actual = hwep_vec(*cts.T)
+    np.testing.assert_allclose(p_expected, p_actual)
+    p_actual = hwep_vec_jit(*cts.T)
+    np.testing.assert_allclose(p_expected, p_actual)
+
+
+def test_hwep__raise_on_negative():
+    args = [[-1, 0, 0], [0, -1, 0], [0, 0, -1]]
+    for arg in args:
```
tomwhite, line 68: Nit: might be slighter neater to parameterise the test?

sgkit/stats/hwe.py
```diff
@@ -0,0 +1,178 @@
...
+        mask = ds["call_genotype_mask"].any(dim="ploidy")
+        gtc = xr.where(mask, -1, ds["call_genotype"].sum(dim="ploidy"))  # type: ignore[no-untyped-call]
```
tomwhite, line 174: Style nit/question: we are being a bit inconsistent on case for these variables. E.g. in `count_alleles` the variables are `G`, `CTS`, etc. Also, in the [summary stats PR](https://github.com/pystatgen/sgkit/pull/102/files) I used `G` for genotypes, but `n_het` for heterogeneous counts. 

Are there some rules we can use? E.g. use lowercase except when we are replicating a paper which uses `X`, etc?

sgkit/stats/association.py
eric-czech: It was duplicated for some reason.

sgkit/stats/hwe.py
eric-czech: https://github.com/pystatgen/sgkit/pull/76/commits/466859e584c2dbe716f92d417c87c150ffa1d278#diff-b7dd01040721465bb47249dc10bb65f2R28

sgkit/stats/hwe.py
eric-czech: https://github.com/pystatgen/sgkit/issues/115

sgkit/tests/test_hwe.py
eric-czech: https://github.com/pystatgen/sgkit/pull/76/commits/466859e584c2dbe716f92d417c87c150ffa1d278#diff-8778a759770b6a59711bea6328271b0bR66

sgkit/stats/hwe.py
eric-czech: https://github.com/pystatgen/sgkit/pull/76/commits/466859e584c2dbe716f92d417c87c150ffa1d278#diff-2ec013d1f086d9558934c38ac90411d1R173

I like the capital convention so array names don't conflict with scalars when there are a lot of both.  When there aren't a lot of either, switching to lower case name sounds good and is what I would prefer too.  I don't know how to make a threshold for that clear, so I've been trying to err on the side of sticking with the capital letter convention for consistency (thanks for flagging this one).

I filed https://github.com/pystatgen/sgkit/issues/117 to make a record of this.

Issue #94, opened 2020-08-05
Single site popgen stats following tskit


jeromekelleher:
It would be useful to have a set of basic single site popgen statistics implemented to help guide overall API design and provide us with easy benchmarks to run. Having put a fair bit of thought and effort into this recently for [this paper](https://www.genetics.org/content/215/3/779), I'm hoping that we can borrow some of the API design and semantic choices from the [tskit stats API](https://tskit.readthedocs.io/en/latest/stats.html).

My suggestion would be to implement sgkit functions that correspond to:

- [diversity](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.diversity)
- [divergence](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.divergence)
- [Fst](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.Fst)
- [Tajima's D](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.Tajimas_D)
- [Patterson's F4](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.f4)

This would be a good overview of the basic popgen stats, and I think give us a good sample of code to generalise from in terms of coming up with multiple dispatch backends, etc.

Obviously I'm biased in suggesting that we follow the tskit definitions closely, but we did spend a long time thinking through how best to arrange the API and semantics, and the implementations are *very* thoroughly tested. It would also make our lives a bit easier here if we don't have so many choices to make, but of course that's up for discussion. A big advantage is that it would make testing the implementations straightforward, since we could simply generate some simulated test data using [msprime](https://msprime.readthedocs.io/en/latest/), and verify that we get the same answer as tskit.

What do we think of this as a general strategy for putting PopGen stats into the library? I'm not saying we have to slavishly copy the function and parameter names etc, but rather use them as a guide.

eric-czech:
> I think give us a good sample of code to generalise from in terms of coming up with multiple dispatch backends, etc.

These generalizations would include distributed processing in your mind right?

> since we could simply generate some simulated test data using msprime, and verify that we get the same answer as tskit

Sounds great to me, it's a similar approach to what I took with https://github.com/pystatgen/sgkit/issues/31 and scikit-alelle for verification so we'd align nicely.

> What do we think of this as a general strategy for putting PopGen stats into the library?

[...]

PR #100, opened 2020-08-08
[WIP] Popgen stats

daletovar:
1. adds @jeromekelleher's code for converting from tskit trees to sgkit datasets, diversity, and divergence.
2. adds some minimal tests for the popgen stats.

Going forward in this PR I'd like to add:
- a test for ts_to_dataset
- windowing for popgen stats

I might save Fst and the others for a different PR.

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -11,6 +11,7 @@ from .io.vcfzarr_reader import read_vcfzarr
 from .stats.aggregation import count_call_alleles, count_variant_alleles, variant_stats
 from .stats.association import gwas_linear_regression
 from .stats.hwe import hardy_weinberg_test
+from .stats.popgen import Fst, Tajimas_D, divergence, diversity
 from .stats.regenie import regenie
 
 __all__ = [
@@ -28,4 +29,8 @@ __all__ = [
     "regenie",
     "hardy_weinberg_test",
     "variant_stats",
+    "diversity",
+    "divergence",
+    "Fst",
+    "Tajimas_D",
 ]
--- a/sgkit/api.py
+++ b/sgkit/api.py
@@ -24,6 +24,7 @@ def create_genotype_call_dataset(
     variant_id: Any = None,
 ) -> xr.Dataset:
     """Create a dataset of genotype calls.
+
     Parameters
     ----------
     variant_contig_names : list of str
@@ -45,10 +46,12 @@ def create_genotype_call_dataset(
         omitted all calls are unphased.
     variant_id: array_like, str or object, optional
         The unique identifier of the variant.
+
     Returns
     -------
     :class:`xarray.Dataset`
         The dataset of genotype calls.
+
     """
     check_array_like(variant_contig, kind="i", ndim=1)
     check_array_like(variant_position, kind="i", ndim=1)
@@ -112,10 +115,12 @@ def create_genotype_dosage_dataset(
         missing value.
     variant_id: array_like, str or object, optional
         The unique identifier of the variant.
+
     Returns
     -------
     xr.Dataset
         The dataset of genotype calls.
+
     """
     check_array_like(variant_contig, kind="i", ndim=1)
     check_array_like(variant_position, kind="i", ndim=1)
@@ -144,31 +149,3 @@ def create_genotype_dosage_dataset(
         data_vars["variant_id"] = ([DIM_VARIANT], variant_id)
     attrs: Dict[Hashable, Any] = {"contigs": variant_contig_names}
     return xr.Dataset(data_vars=data_vars, attrs=attrs)
-
-
-def ts_to_dataset(ts, samples=None):
-    """
-    Convert the specified tskit tree sequence into an sgkit dataset.
-    Note this just generates haploids for now. With msprime 1.0, we'll be
-    able to generate diploid/whatever-ploid individuals easily.
-    """
-    if samples is None:
-        samples = ts.samples()
-    tables = ts.dump_tables()
-    alleles = []
-    genotypes = []
-    for var in ts.variants(samples=samples):
-        alleles.append(var.alleles)
-        genotypes.append(var.genotypes)
-    alleles = np.array(alleles).astype("S")
-    genotypes = np.expand_dims(genotypes, axis=2)
-
-    df = create_genotype_call_dataset(
-        variant_contig_names=["1"],
-        variant_contig=np.zeros(len(tables.sites), dtype=int),
-        variant_position=tables.sites.position.astype(int),
-        variant_alleles=alleles,
-        sample_id=np.array([f"tsk_{u}" for u in samples]).astype("U"),
-        call_genotype=genotypes,
-    )
-    return df
--- a/sgkit/api.py
+++ b/sgkit/api.py
@@ -24,7 +24,6 @@ def create_genotype_call_dataset(
     variant_id: Any = None,
 ) -> xr.Dataset:
     """Create a dataset of genotype calls.
-
     Parameters
     ----------
     variant_contig_names : list of str
@@ -46,12 +45,10 @@ def create_genotype_call_dataset(
         omitted all calls are unphased.
     variant_id: array_like, str or object, optional
         The unique identifier of the variant.
-
     Returns
     -------
     :class:`xarray.Dataset`
         The dataset of genotype calls.
-
     """
     check_array_like(variant_contig, kind="i", ndim=1)
     check_array_like(variant_position, kind="i", ndim=1)
@@ -115,12 +112,10 @@ def create_genotype_dosage_dataset(
         missing value.
     variant_id: array_like, str or object, optional
         The unique identifier of the variant.
-
     Returns
     -------
     xr.Dataset
         The dataset of genotype calls.
-
     """
     check_array_like(variant_contig, kind="i", ndim=1)
     check_array_like(variant_position, kind="i", ndim=1)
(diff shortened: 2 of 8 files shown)


sgkit/__init__.py
```diff
@@ -5,10 +5,12 @@
     DIM_VARIANT,
     create_genotype_call_dataset,
     create_genotype_dosage_dataset,
+    ts_to_dataset,
```
jeromekelleher, line 8: I don't think we want this in core library as it's introducing a dependency on tskit and we'd probably want to do it a bit differently anyway. Let's just consider ``ts_to_dataset`` a testing utility for now.

sgkit/stats/popgen.py
```diff
@@ -0,0 +1,47 @@
...
+    # the behaviour we want, but it's a starting point. Note that
+    # this is different to the tskit default behaviour where we
+    # normalise by the size of windows so that results
+    # in different windows are comparable. However, we don't have
+    # any information about the overall length of the sequence here
+    # so we can't normalise by it.
+    return pi.sum()
+
+
+def divergence(
```
jeromekelleher, line 33: For consistency we should make the parameters ``ds1`` and ``ds2``, which have the same setup for ``allele_counts``. Either this, or we have one dataset and provide two selectors so specify the two subsets - but, then how would we cache the ``allele_counts`` computation? I think ``ds1``, ``ds2`` is the right approach.

What do you think @eric-czech ?

sgkit/tests/test_popgen.py
```diff
@@ -0,0 +1,57 @@
+import pytest
+import msprime
+import numpy as np
+
+from sgkit import ts_to_dataset, diversity, divergence
+
```
jeromekelleher, line 6: Let's bring the definition of ``ts_to_dataset`` in here for now - we can move it into a better location later once we get an idea of where and how its used.

sgkit/api.py
```diff
@@ -24,7 +24,6 @@ def create_genotype_call_dataset(
     variant_id: Any = None,
 ) -> xr.Dataset:
     """Create a dataset of genotype calls.
-
```
jeromekelleher, line 27: We should probably put  any docstring formatting changes in a separate PR/issue.

sgkit/stats/popgen.py
eric-czech: `ds1`, `ds2` sounds good, though I'd have a slight preference for a signature like:

```python
def divergence(ds: Dataset, allele_counts: str, ...): ...
```

This would make the variable a required parameter for a single array and imply that the segregation must live formally as an axis in that array.  The advantages of this are:

- You don't have to deal with checking that `ac1`/`ac2` or that the variables in `ds1`/`ds2` have equal numbers of variants or alleles since that would be implicit in creating an array with both of them
- It would be more natural in a workflow that segregates more than two populations
- It would fit a little better in a pipeline:

```python
from sgkit import count_alleles
from sgkit.testing import simulate_genotype_call_dataset

ds = simulate_genotype_call_dataset(n_variant=100, n_sample=20, seed=0)
ds['sample_population'] = np.repeat([0, 1], ds.dims['samples'] // 2)
res = (
    ds
    .assign(
        allele_counts=ds
        .groupby('sample_population')
        .map(count_alleles)
        .rename_dims(dict(sample_population=populations))
    )
    .pipe(divergence, allele_counts='allele_counts')
)
```

The disadvantages are:

- The function would need to check that `ds.dims["populations"] == 2`, which is a little clunky 
- We should formalize the notion of a "populations" dimension if you go this route (i.e. we should use that same dimension name everywhere else)
- It assumes users are a bit more Xarray savvy 

I think finding common variables/dimension names for populations is inevitable so I'm sure that will come up again even if it's not here.  Otherwise, I don't feel strongly one way or the other but that little bit of extra generality could be worth it.

jeromekelleher: Good ideas @eric-czech. The term "population" is problematic for lots of reasons, but I guess if we can come up with a term for "group of samples we want to aggregate together" I'm all for your approach.

I think at this point we should use xarray to its maximum potential/follow the idioms as closely as possible to establish the core functionality. Perhaps later then we can see how we can make another interface that builds on this that's simpler/easier for novice users, etc.

hammer: >"group of samples we want to aggregate together"

`sample_group`, `sample_subset`, `subsample`?

jeromekelleher: I like ``sample_group`` - seem unambiguous and flexible.

daletovar: For right now I did this:
`divergence(ds1: Dataset, ds2: Dataset,...):`

If/once we formalize a "groups" dimension, we can change it.

jeromekelleher: OK, sounds good. We should probably create an issue for the ``group`` dimension, seems like a pretty fundamental aspect of the design.

hammer:
@daletovar thanks for your work on this PR so far! In the future, could you be sure to mention the issue in the PR? In this case, https://github.com/pystatgen/sgkit/issues/94. We should update our contributing docs with this request...

daletovar:
@hammer, thanks. Yes, I'll be sure to do that in the future.

jeromekelleher:
@daletovar, can you rebase this so that it's up to date with upstream/master please?

jeromekelleher:
This branch looks pretty messed up @daletovar - I think it would be simplest to rebase against upstream/master and force push.

daletovar:
@jeromekelleher, sorry for the radio silence. I'll see what I can do about that. I think for now I'll leave the Patterson F stats for another time. I've had a hard time finding where they're implemented in tskit and there are differences between the tskit and scikit allele outputs.

jeromekelleher:
> @jeromekelleher, sorry for the radio silence. I'll see what I can do about that. I think for now I'll leave the Patterson F stats for another time. I've had a hard time finding where they're implemented in tskit and there are differences between the tskit and scikit allele outputs.

I'm still not able to see a clean diff here @daletovar making it very difficult to see what's being done - can you rebase against ``upstream/master`` please?

jeromekelleher:
@daletovar - there's more issues with linting here. Can you clean up please?

daletovar:
@jeromekelleher, when it's installing dependencies. it looks like there's an issue building msprime. I'm not sure what's causing it. Do you know of a fix or is there something else we can use to test against tskit?

jeromekelleher:
I see - we just need to install ``libgsl`` via apt I think.

sgkit/stats/popgen.py
jeromekelleher: Issue opened: #224

daletovar:
Okay, is that just a matter of adding a requirement?

jeromekelleher:
> Okay, is that just a matter of adding a requirement?

No, it's done as part of the CI setup. I guess we'd add a line around [here](https://github.com/pystatgen/sgkit/blob/5e044477502fee70fd141c0c6233ce104f1f4528/.github/workflows/build.yml#L20) that looks something like 

```
sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
```

daletovar:
Thanks so much, @jeromekelleher. That worked. I'll look into this Fst issue.

daletovar:
It looks like the errors are caused by the fact that tskit returns `nan `from `ts.diversity` when the dataset has a single sample, as is the case when we set the size to 2 or 3 and split it in half. What I have returns 0 from `sgkit.diversity` when there's a single sample, which has downstream effects on Fst. I can make `sgkit.diversity` return `nan` on single sample datasets to be in line with tskit. Is this what we want? 
@jeromekelleher

jeromekelleher:
Ah, ok, thanks @daletovar. Let's make this agree with tskit's output for now and reexamine these corner cases later.

daletovar:
@tomwhite, I think this is ready for review.

jeromekelleher (dismissed):
The call signatures and return types aren't quite what we're going to want, but I think this is a step in the right direction and we should merge. I'll open some follow-up issues to track the changes that'll need to be made.

sgkit/stats/popgen.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Returns
+    -------
+    DataArray
+        diversity value.
+    """
+    if len(ds.samples) < 2:
+        return xr.DataArray(np.nan)
+    if allele_counts not in ds:
+        ds = count_variant_alleles(ds)
```
tomwhite, line 40: I'm not sure we should modify the passed-in dataset. Instead we should return a dataset (like in #217) and add the allele counts to that dataset.

tomwhite (approved):
LGTM. I agree - the functions should return datasets, and we should discuss if the pattern of being able to override the variable name is how we want it (i.e. `allele_counts`) - but fine to merge then address these in follow ups.

tomwhite:
Also - the new functions should be added to `api.rst`.

jeromekelleher (dismissed):
LGTM, thanks @daletovar. I'll follow up with issues needed once this is merged.

tomwhite (approved):

jeromekelleher:
@Mergifyio update

jeromekelleher:
@Mergifyio refresh

jeromekelleher:
I don't know why mergify doesn't like this - we seem to be missing a tick for ``- [ ] `status-success=build```, but it looks like all the checks are passing. Any idea what's happening here @ravwojdyla?

ravwojdyla:
@jeromekelleher sorry, for late reply (missed this comment). Yea, it's bizarre, also when I try to validate this PR in the [simulator](https://dashboard.mergify.io/installation/11295598/simulator?repositoryName=sgkit&pullRequestNumber=100), it's timing out:
> Oops, looks like https://dashboard.mergify.io is unreachable, try refreshing.

jd:
Your rule specifies `build` as the status check success though there are two checks named `build (3.8)` and `build (3.7)`. You might want to fix your rules.

ravwojdyla:
@jd thanks for the hint, we have made that change recently #218. Btw, is there something wrong with the simulator, it's timing out for me.

jd:
The simulator should work fine — at least it does here. I've sent you an email so we can figure this out. :)

ravwojdyla:
Got an update from @jd about the mergify simulator:

> Your user in particular seems to trigger a 403 on GitHub API side when we try to check user permission. We don't know why exactly yet, but we've a bug opened on our side to track that down and fix it. :)

Soooo, if someone could check the configuration from #244 on this PR and report back, that would be helpful. Simulator [link](https://dashboard.mergify.io/installation/11295598/simulator?repositoryName=sgkit&pullRequestNumber=100). @jeromekelleher ?

jeromekelleher:
@Mergifyio update

jeromekelleher:
Looks like this is breaking because of #245

tomwhite:
I just merged #246 which should fix the build.

@Mergifyio update

tomwhite:
@jeromekelleher @ravwojdyla how do we get mergify to actually merge this?!

ravwojdyla:
@tomwhite we need to merge #244 first, but I can't validate that configuration because https://github.com/pystatgen/sgkit/pull/100#issuecomment-691276250.

jeromekelleher:
Trying another update here, to see if the changes in #244 will unplug things.

jeromekelleher:
@Mergifyio update

jeromekelleher:
@Mergifyio rebase

jeromekelleher (approved):

jeromekelleher (approved):

jeromekelleher (approved):

jeromekelleher:
Github workflows have been modified, so this has be to be merged manually.

PR #101, opened 2020-08-10
Add messages to exceptions in check_array_like. Fixes #96

### LINKED ISSUE #96: Add messages to check_array_like exceptions
The [check_array_like](https://github.com/pystatgen/sgkit/blob/master/sgkit/utils.py#L8) function currently raises exceptions without any error messages, which is unhelpful when trying to figure out what's wrong with your input.

I'd suggest we avoid this practise in general - I'm not sure there's a good reason for not explaining why you're raising an exception?

tomwhite:


eric-czech:
Thank you @tomwhite!  Those have been frustrating the few times I've hit them so far.

Issue #105, opened 2020-08-12
Add zarr to setup.cfg


eric-czech:
In https://github.com/pystatgen/sgkit/pull/40, zarr should have been added to [setup.cfg#L27](https://github.com/pystatgen/sgkit/blob/dce23d09dcac61d6cd5f7b71fa59f65eb4ff3176/setup.cfg#L27) as well as 
[requirements.txt#L5](https://github.com/pystatgen/sgkit/blob/dce23d09dcac61d6cd5f7b71fa59f65eb4ff3176/requirements.txt#L5).  This seems to be breaking the build in the IO repos.

ravwojdyla:
Related #65

PR #106, opened 2020-08-12
Adding zarr dep

eric-czech:
https://github.com/pystatgen/sgkit/issues/105

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,4 +30,4 @@ repos:
     hooks:
       - id: mypy
         args: ["--strict", "--show-error-codes"]
-        additional_dependencies: ["numpy", "xarray", "dask[array]", "scipy"]
+        additional_dependencies: ["numpy", "xarray", "dask[array]", "scipy", "zarr"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ install_requires =
     xarray
     dask[array]
     scipy
+    zarr
     setuptools >= 41.2  # For pkg_resources
 setup_requires =
     setuptools >= 41.2


eric-czech:
For some reason no.  I have done that with several dependencies now (forgotten to add them there) and it doesn't seem to matter, but I don't understand why.  

Should we add it even if mypy runs without it?

ravwojdyla:
@eric-czech afaiu, we should add them to mypy, [doc](https://github.com/pre-commit/mirrors-mypy):

> Because pre-commit runs mypy from an isolated virtualenv (without your dependencies) you may also find it useful to add the typed dependencies to additional_dependencies so mypy can better perform dynamic analysis

eric-czech:
That makes it sound like we should only add them if we're not going to add `ignore_missing_imports = True` for the whole package.

ravwojdyla:
@eric-czech yea, I think for consistency I would still add it, I don't really know all the different ways `mypy` works exactly, and how it uses different dependencies, thus would just lean towards being safe and just install all of them, but if you feel strongly against it - feel free to ignore.

eric-czech:
Mm well my arguments against it would be:

- It bloats the pre-commit environment and makes it take longer to come up (with no benefit)
- It's a 3rd place to remember to add the deps
- Xarray doesn't do it: https://github.com/pydata/xarray/blob/master/.pre-commit-config.yaml

I can't really think of any arguments in favor of it assuming it would break if we didn't add the dependency and then removed the corresponding `ignore_missing_imports = True`.  Do you know if that was true?  Was that perhaps why you added Xarray when you defined it early on?

All that said I added it but now I'm thinking it might make more sense to surface this in the developer docs and say that only typed dependencies need to go in there.  Are there any other advantages to it I'm missing?

ravwojdyla:
@eric-czech I don't know enough about the internals of `mypy` to have an opinion about this one, that's way I tend to lean on being safe and just adding all of it. Re your points:
 * pre-commit caches envs so it shouldn't be an issue after the env is initially created (this includes caching on github actions)
 * hopefully this gets fixed as part of #39 
 * I'm no sure why xarray doesn't do it
 * not sure about your question regarding `ignore_missing_imports`, would require some research
 * +1 to documentation: https://github.com/pystatgen/sgkit/issues/109

eric-czech:
What I mean is that there are only two situations that ever seem to come up after experimenting with mypy through all these changes:

1. The dependency is a package with no type hints so adding `ignore_missing_imports=True` to `setup.cfg` for it means mypy ignores it completely (and it doesn't need to be in `additional_dependencies`)
2. The dependency has type hints so if you forget to add it to `additional_dependencies`, the mypy hook won't pass until you do (I just tried it by removing Xarray from the list).

Basically it appears to be sufficiently safe that way.

eric-czech:
but 👍 to https://github.com/pystatgen/sgkit/issues/39.  For those not very familiar with mypy like myself I think that doc you linked to is good to know in https://github.com/pystatgen/sgkit/pull/106#issuecomment-673377754.

ravwojdyla:
@eric-czech yea, I don't know enough about mypy either, I use it as a black box, and thus prefer to stay on the safe side of things. To elaborate, from [mypy doc](https://mypy.readthedocs.io/en/stable/running_mypy.html):

> If you are unable to find any existing type hints nor have time to write your own, you can instead suppress the errors. All this will do is make mypy stop reporting an error on the line containing the import: the imported module will continue to be of type Any.

> To suppress all missing import imports errors from a single library, add a section to your mypy config file for that library setting ignore_missing_imports to True. For example, suppose your codebase makes heavy use of an (untyped) library named foobar. You can silence all import errors associated with that library and that library alone by adding the following section to your config file:
> ```
> [mypy-foobar.*]
> ignore_missing_imports = True
> ```
> Note: this option is equivalent to adding a # type: ignore to every import of foobar in your codebase. For more information, see the documentation about configuring import discovery in config files. The .* after foobar will ignore imports of foobar modules and subpackages in addition to the foobar top-level package namespace.

So when I read this one, it looks like the key part is this: `this option is equivalent to adding a # type: ignore to every import of foobar in your codebase`, I don't know if that is equivalent to completely not having that library installed, it could be, I'm just not sure. One thing that comes to me mind, if we do NOT install the dependencies in mypy, even if there are some parts of them that are typed, they will always be untyped/missing in our mypy test. Additionally I don't know if mypy can do some sort of type inference in case of missing stubs (if the module is available, for example of the statically available variables/classes) - and it might not do that at all, I'm just no sure. It seems like an overhead to check if each library has typing/stubs (even partial) and only add those to mypy check is unnecessary. Anyway, when not sure I prefer to stay on the safe side, and if you believe that `ignore_missing_imports` is equivalent to not having the library installed, I'm good :)

eric-czech:
> if you believe that ignore_missing_imports is equivalent to not having the library installed

I do believe it for most our dependencies at the moment since "even if there are some parts of them that are typed" wouldn't be a factor, but totally see your point and who knows if/when these libraries start adopting typing.  Having a 3rd place to add all the deps puts me on the fence and I'll just stop caring since it shouldn't matter after #39 (staying on the safe side sounds good there).  Thanks for thinking about it some more.

Issue #107, opened 2020-08-13
Investigate mergify


ravwojdyla:
I will close this issue, and create a separate one for `mergify`.

_Originally posted by @ravwojdyla in https://github.com/pystatgen/sgkit/issues/26#issuecomment-673381620_

Related: https://github.com/pystatgen/sgkit/issues/26#issuecomment-655656832

ravwojdyla:
FYI, I have done a quick review of mergify features, we could automate a lot of PR related actions, see full [doc](https://doc.mergify.io/actions.html). Mergify itself is now enabled for pystatgen org (thanks @hammer). Since this is just the beginning of the project, I wonder if it's better to hold off automatic merge for a bit until we crystallise the patterns/groups/reviewers etc around PRs, for example: 
 * how many reviewers do we requires and from which groups
 * do we reset reviews on PR updates
 * do we bypass reviews in certain cases (small PRs, ex: new dependency)

We could try to have this discussion right now(?), alternatively we could give ourselves some more time to naturally build up those patterns and then codify them via mergify.

In the meantime I will submit a minimal mergify config to:
 * ping authors on conflicts
 * delete merged branches

Btw, a simple rebase config for use could look like [this](https://github.com/Mergifyio/mergify-engine/blob/master/.mergify.yml#L1-L17).

How does that sound? (@tomwhite @eric-czech @jeromekelleher)

eric-czech:
Sounds good to me.  Codifying the review process as part of the minimal config now sounds good too, though I would vote that we start start conservatively with:

- Require 1 reviewer (from anywhere)
- Don't reset reviews on PR updates
- No bypass on reviews

That's been the process so far in my head anyhow.

ravwojdyla:
@eric-czech thanks. I added a simple mergify script that has a auto merge op-in option by labelling PR `auto-merge`, see in #122

jeromekelleher:
[...]

Issue #26, opened 2020-07-08
Git merge strategy - disable merge commits?


ravwojdyla:
I would like to suggest to disable merge commits for PRs and essentially enforce rebase/squash only - the advantage of this strategy is that the git (graph) history is a lot cleaner. Would that be fine with everyone?

jeromekelleher:
Merge commits make the history a bit more readable, IMO so I'd be mild -1. We tried out the "force linear history" thing on GitHub for a bit and ended up reverting.

Have you considered using [mergify](https://mergify.io/) instead? You can set this up to automatically rebase, rerun CI and then merge if it passes. It was life changing for me, I love it!

eric-czech:
+1 to mergify instead

ravwojdyla:
@jeromekelleher in this issue I don't worry about automation (`mergify`), just the git history. I have been working with "linear history", and just prefer that for cleaner history and easier rebasing. I see `mergify` as an orthogonal question, we could use `mergify` with rebase/squash only method, right?

jeromekelleher:
So I totally agree with having a linear history. I've just found that having "squash" commits from the GitHub interface isn't actually that helpful as 

a. You lose the merge commit that serves as a summary of the commits that went into a PR (and it's a bit too strict to day all PRs must be squashed down to one commit - that's not how you get to a clear history, IMO). You also lose the nice branchy view of history on the GitHub UI, which gives a good sense of who did what, when.
b. You still have to get people to rebase before you click the "merge/squash/whatever" button on GitHub, because the CI they have run might be out of date. (GitHub just checks for merge conflicts).

Mergify takes the drudgery out of this, and lets you keep a nice linear Git history, and means you have to ask contributors to rebase a *lot* less.

ravwojdyla:
@jeromekelleher yea, I see. So I'm +1 `mergify` but that is separate question :)

So to be clear, there are really 3 options regarding merge strategy (all of which are supported by `mergify`):
1. merge: we keep all the separate commits from the PR via merge commit -> branch in git graph and non-linear history
2. rebase: we keep all the separate commits from the PR via rebase -> linear history
3. squash: we get a single squashed commit per PR -> linear history

In this issue I'm advocating specifically to avoid `merge` and instead use `rebase` (or `squash` if people prefer that one, though it's a bit against git/linux philosophy). So we still keep each commit, keep linear history (making it easier to use git bisect for example).

jeromekelleher:
[...]

PR #108, opened 2020-08-13
Add initial dev conventions (for tests)

ravwojdyla:
Related: #75

PR #110, opened 2020-08-13
Add pr/git ops doc

ravwojdyla:
Related: #26 and #107

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -206,6 +206,28 @@ To run a specific tool (``black``/``flake8``/``isort``/``mypy`` etc)::
 You can omit ``--all-files`` to only check changed files.
 
 
+PR/Git ops
+~~~~~~~~~~
+
+We currently use ``rebase`` or ``squash`` PR merge strategies. This means that
+following certain git best practices will make your development life easier.
+
+
+1. Try to create isolated/single issue PRs
+
+   This makes it easier to review your changes, and should guarantee
+   a speedy review.
+
+2. Try to push meaningful small commits
+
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -244,3 +244,24 @@ Use double underscore to organize tests into groups, for example:
 
    def test_foo__accepts_strings():
      ...
+
+PR/Git ops
+~~~~~~~~~~
+
+We currently use ``rebase`` or ``squash`` PR merge strategies. This means that
```
eric-czech, line 251: Am I missing something or does this strategy not imply several specific changes to the dev workflow beyond general suggestions like keeping commits small so that they make more sense when rebased rather than merged into one?  I get that the global repo setting only applies to how PRs are merged now, but what happens if there is a merge commit in a branch for a PR?

ravwojdyla: @eric-czech the answer to that is: that it depends. If you follow the recommendations from the links below and use rebase/squash you should not end with merge commits in your feature branch. If you feature branch nevertheless ends up with merge commit inside, `rebase` merge of that specific PR **might** (it's not guarantee) not be possible due to conflicts, and instead you either have to use `squash` merge (which would automatically squash all changes into a single commit for you), or locally rebase and push (as suggested by the official git docs).

ravwojdyla:
moved this section inside the `Contributing to code` re: https://github.com/pystatgen/sgkit/pull/108#issuecomment-673455196

docs/contributing.rst
eric-czech: 👍

Issue #74, opened 2020-07-29
mypy support for hypothesis decorators


ravwojdyla:
@ravwojdyla do you know of a way to avoid doing this everywhere in tests:

```python
@given(st.integers(0, 25))  # type: ignore[misc]
@settings(max_examples=10)  # type: ignore[misc]
def test_something(n: int) -> None: ...
```

I get mypy errors w/o a return type (only for decorated functions) and I'm adding `ignore[misc]` for `Untyped decorator makes function "test_something" untyped`.  Is there some way to tell mypy to ignore type checking for any function wrapped by hypothesis decorators globally?

_Originally posted by @eric-czech in https://github.com/pystatgen/sgkit/pull/69#issuecomment-665541975_

PR #111, opened 2020-08-13
Allow untyped decorators in tests

ravwojdyla:
Related: #74. This allows to use untyped decorators in tests, ex: hypothesis decorators.


eric-czech:
oo yea, much better.

Issue #85, opened 2020-08-03
Add allele counting function for individual calls


eric-czech:
We need a function to count alleles for individual calls in addition to one counting alleles across samples.  I propose we call it `count_call_alleles` and rename `count_alleles` to `count_variant_alleles`.

This would be akin to [GenotypeAlleleCountsArray](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#genotypeallelecountsarray) and [AlleleCountsArray](https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#allelecountsarray) in scikit-allele.

eric-czech:
An important concern with this is whether or not multiple alleles should be counted separately.  

If this function was to generate arrays of shape `(n_variant, n_sample)`, then it clearly implies that some reduction was performed across the known alleles.  The most obvious reduction is to count any non-reference allele, but this can drop important information for downstream tasks.  It could instead return an array of shape `(n_variant, n_sample, n_allele)` as @timothymillar mentions in https://github.com/pystatgen/sgkit/issues/21#issuecomment-671119651, and we should start with this assumption.

Implementing an efficient function that returns counts as an array of shape `(n_variants, n_samples, n_alleles)` with values equal to the number of times each allele appears across chromosomes is not trivial and should likely involve something like a `bincount` across the ploidy dimension.  The "I only care about non-reference allele counts" case on the other hand is much closer to trivial, e.g. `call_allele_counts = (ds['call_genotype'] > 0).sum(dim='ploidy')` or if a user didn't want to count partially missing calls then `xr.where(ds['call_genotype_mask'], -1, (ds['call_genotype'] > 0).sum(dim='ploidy'))`.  We could and probably should wrap this functionality in the API with some docs, but it's also a degenerate case from the `(n_variants, n_samples, n_alleles)` result where you could do the same thing with `call_allele_counts[..., 1:].sum(dim='alleles')` (to get non-reference allele counts).  This would not be as efficient though.

**tl;dr** This function should return a count array of shape `(n_variants, n_samples, n_alleles)` and missing values should be excluded from the counts.

tomwhite:
Fixed in #114

Issue #49, opened 2020-07-16
Track and improve the performance of allele counting method


eric-czech:
The solution to https://github.com/pystatgen/sgkit/issues/3 in https://github.com/pystatgen/sgkit/pull/36 is naive and possibly unacceptably slow.  This will be true if Dask does not optimize the loop over allele indexes to a single pass on the genotypes array (which it probably won't).

The extension to this proposed in https://github.com/pystatgen/sgkit/pull/36#issuecomment-656611356 would definitely solve the problem in a single pass if Dask supported counting rows like numpy does, but it currently doesn't.  

There may be some other efficient ways to do it without dropping down to writing custom kernels but in any case, we should track the performance of this implementation (and others) as part of a benchmark suite like @alimanfoo mentioned in https://github.com/pystatgen/sgkit/pull/36#issuecomment-658893949 so we can measure the impact of future iterations more passively and prevent regressions.

alimanfoo:
Thanks for picking this up @eric-czech, I think I started writing something but then got distracted, and you summarised it much better than I would've.

hammer:
>if Dask supported counting rows like numpy does, but it currently doesn't.

Is there an upstream issue we can track?

eric-czech:
I added one: https://github.com/dask/dask/issues/6423

tomwhite:
Can this be closed now?

eric-czech:
I added https://github.com/pystatgen/sgkit/issues/68 to track the benchmarking discussion so I think this can be closed now.

eric-czech:
Reopening largely to revisit numba, re: https://github.com/pystatgen/sgkit/pull/114.

I was generally hoping to avoid dropping down to numba as much as possible because it then means users have no freedom to choose array backends that the computations run on.  I think this case, counting alleles for variants, is a bit of a grey area for that since it is supported somewhat well with any implementations of `bincount` (as in https://github.com/pystatgen/sgkit/pull/36) applied to larger arrays.

Counting alleles for individual calls (https://github.com/pystatgen/sgkit/issues/85) is different though and afaik, there really is no way to utilize `bincount` even remotely efficiently.  All the vectorization would be in python since it doesn't support reduction across an axis.  If numba is going to be necessary for that (which I think it is), then it would likely make sense to start thinking of counting alleles for variants as a sum of the counts for individual calls, i.e.:

```python
def count_call_alleles(ds) -> DataArray:
[...]

PR #114, opened 2020-08-16
[WIP] count_allele_calls

timothymillar:
See issue #85 

This implements an additional jitted function `count_call_alleles_ndarray_jit` for ndarrays only rather than doing it all in dask.
This approach seems to be inline with the goals outlined [here](https://discourse.pystatgen.org/t/development-and-dispatch-for-genetics-methods-over-pydata-libraries/46/2) but I'm happy to replicate the approach of the original `count_alleles` function if that is preferred.

Likewise I can re-write `count_alleles` using this approach which should improve performance (mainly on chucked arrays due to `njit(..., nogil=True)`)

~~I haven't added `numba` to `requirements.txt` or `setup.py` Because that is done in #76~~

I guess add numba now for CI and fix conflict later


eric-czech:
Thanks for picking this up @timothymillar!

I've been hesitant to wade into numba for things like counting (put a few thoughts at https://github.com/pystatgen/sgkit/issues/49#issuecomment-674561391) but I think it's probably the right choice here.  One of the things we accounted for in picking the variables for the datasets was that ultimately a sentinel value would be very useful in JIT functions, as opposed to using the separate mask.  

It's a shame that there doesn't seem to be a good way to do this with Dask/Xarray, but if we do lose the freedom in array backends it would be nice to still support CPU/GPU easily.  Have you thought at all about using a gufunc for this, which would make it easier to compile to either?  I think the whole module could collapse down to something like this in that case, where using the sentinel also simplifies things:

```python
@numba.guvectorize(['void(numba.int8[:], numba.int32[:], numba.int32[:])'], '(n),(k)->(k)')
def count_alleles(x, _, out):
    out[:] = 0
    for v in x:
        if v >= 0:
            out[v] += 1
            
def count_call_alleles(ds) -> DataArray:
    G = da.asarray(ds.call_genotype)
    # This array is only necessary to tell dask/numba what the 
    # dimensions and dtype are for the output array
    O = da.zeros(G.shape[:2] + (ds.dims['alleles'],), dtype='int32')
    O = O.rechunk(G.chunks[:2] + (-1,))
    return xr.DataArray(
        count_alleles(G, O),
        dims=('variants', 'samples', 'alleles'),
        name='call_allele_count'
    )

def count_variant_alleles(ds) -> DataArray:
    return (
        count_call_alleles(ds)
        .sum(dim='samples')
        .rename('variant_allele_count')
    )
```

I'd be a proponent of simply removing the `count_variant_alleles` function too and leaving it up to users in the future if there isn't a big performance difference between a sum following per-call allele counts and a custom kernel/function.

timothymillar:
> Have you thought at all about using a gufunc for this

Good idea, I'll give it a go

> where using the sentinel missing value also simplifies things vs using the separate mask variable

I was a little uncertain of the application of the mask as their seems to be two options when it comes to a user masking out additional values (which I assume is an intended feature at some point):

1. The user updates the mask (via some API) but the genotype arrays are unchanged hence any function has to use the mask itself either by updating a copy of the genotype calls (as in the current implementation of `count_alleles`) or using the mask itself as in this PR.
2. The user updates the mask (via some API) and then applies it to their DataSet which replaces the (alleles within) genotype calls with the sentinel value in the original DataSet (or in a copy).

Essentially should functions working on a DataSet trust the mask or the sentinel values?

eric-czech:
> Essentially should functions working on a DataSet trust the mask or the sentinel values?

The mask is only a convenience on `ds.call_genotype < 0` so I think it's best to use the sentinel values until both are needed.  

> when it comes to a user masking out additional values (which I assume is an intended feature at some point)

We haven't talked about that yet (feel free to file an issue), but afaik Dask still doesn't support assignment so I see that process as:

1. An operation defines a transformation of the data array that makes additional values equal to the missing sentinel
2. The mask is redefined as `mask = arr < 0`

timothymillar:
@eric-czech The gufunc version is working nicely but I'm having some issues satisfying the CI with the use of `guvectorize`.
Do you know whats causing the Sphinx doc issue?

A couple of notes about the implementation:

- The gufunc is returning an array with dtype `uint8`, I don't think this will be an issue but if it is then the dtype of the dummy array can be used to indicate the output dtype.
- I'm not sure on the docstring style preference as there is a bit of variation in the repo.
- I reversed your signature from `'(n),(k)->(k)'` to `'(k),(n)->(n)'` because the genotype vector has a length of ploidy and `k` is commonly used for ploidy in (polyploid) literature.

eric-czech:
Nice @timothymillar!  This looks great.

> The gufunc is returning an array with dtype uint8, I don't think this will be an issue but if it is then the dtype of the dummy array can be used to indicate the output dtype.

Perfect, makes sense. 

> I'm not sure on the docstring style preference as there is a bit of variation in the repo.

Me neither.  I started using the Dask style like that but then switched to referencing our `ArrayLike` type once that was added and @tomwhite  did too.  I was waiting until Sphinx was up and running before trying to overhaul the signatures, but https://github.com/pystatgen/sgkit/pull/124 would change a lot of it anyhow.  I'm not sure what the standard should be until some more dust settles.  Let us know if you have any thoughts on it.

> I reversed your signature from '(n),(k)->(k)' to '(k),(n)->(n)' because the genotype vector has a length of ploidy and k is commonly used for ploidy in (polyploid) literature.

Good call!

sgkit/stats/aggregation.py
```diff
@@ -1,10 +1,103 @@
...
+           [[1, 1],
+            [1, 1]],
+    <BLANKLINE>
+           [[2, 0],
+            [2, 0]]], dtype=uint8)
+    """
+    n_alleles = ds.dims["alleles"]
+    G = da.asarray(ds["call_genotype"])
+    shape = (G.chunks[0], G.chunks[1], n_alleles)
+    N = da.empty(n_alleles, dtype=np.uint8)
```
eric-czech, line 92: Good idea using this instead!  For the subsequent map_blocks call it might be a good idea to ensure this has only a single chunk so that the blocks broadcast.  I can't imagine why anyone would ever configure the default chunk size to be so small that `n_alleles` items would result in multiple chunks, but guarding against it is probably a good idea.

sgkit/stats/aggregation.py
```diff
@@ -1,10 +1,103 @@
...
+    return xr.DataArray(
+        da.map_blocks(count_alleles, G, N, chunks=shape, drop_axis=2, new_axis=2),
```
eric-czech, line 94: Out of curiosity, what keeps this from working as `count_alleles(G, N)` instead?

timothymillar:
Here's a minimal example explaining the use of `map_blocks`.

setup:
```python
import numpy as np
import dask.array as da

from sgkit.stats.aggregation import count_alleles

n_alleles = 2
genotypes = np.array(
    [[[ 0,  0],
      [ 0,  1],
      [ 1,  0]],
     [[-1,  0],
      [ 0, -1],
      [-1, -1]]], dtype=np.int8)

N = da.empty(n_alleles, dtype=np.uint8)
G = da.asarray(genotypes).rechunk((1,1,1))  # unlikely chunking
```
(All of the below options work correctly if G is not chunked in dimension 2 (ploidy))

**Option 1:** calling `count_alleles` directly:
```python
count_alleles(G, N).compute()
```
Results in error:
```
ValueError: Core dimension `'k'` consists of multiple chunks. To fix, rechunk into a single chunk along this dimension or set `allow_rechunk=True`, but beware that this may increase memory usage significantly.
```
I didn't actually explore the use of `allow_rechunk=True` but even if it achieves the same as below I prefer the more explicit use of `map_blocks`.

**Option 2:** naive use of `map_blocks` (my first attempt):
```python
shape = (G.chunks[0], G.chunks[1], n_alleles)
da.map_blocks(count_alleles, G, N, chunks=shape).compute()
```
Results in incorrect array dimensions: 
```
array([[[1, 0, 1, 0],
        [1, 0, 0, 1],
        [0, 1, 1, 0]],

       [[0, 0, 1, 0],
        [1, 0, 0, 0],
        [0, 0, 0, 0]]], dtype=uint8)
```

**Option 3:**  Using `map_blocks` kwargs to "forget" the ploidy dimensions shape:
```python
shape = (G.chunks[0], G.chunks[1], n_alleles)
da.map_blocks(count_alleles, G, N, chunks=shape, drop_axis=2, new_axis=2).compute()
```
Result:
```
array([[[2, 0],
        [1, 1],
        [1, 1]],

       [[1, 0],
        [1, 0],
        [0, 0]]], dtype=uint8)
```

sgkit/stats/aggregation.py
timothymillar: >  I can't imagine why anyone would ever configure the default chunk size to be so small that `n_alleles` items would result in multiple chunks

Neither but this edge case should be handled correctly by this PR, see below.

sgkit/stats/aggregation.py
timothymillar: `count_alleles(G, N)` results in an error if G is chunked in the ploidy dimension. See comment below for an example.

timothymillar:
> Does it go away if you do from numba import guvectorize and use the decorator that way? I'm not sure if the issue is with that or the numba.int8[:] signature annotations. Perhaps it will work with strings like int8[:] instead?

One of those seems to have fixed it, thanks for the help!

sgkit/stats/aggregation.py
eric-czech: Ah I see.  Thoughts on a `da.asarray(ds["call_genotype"].chunk(chunks=dict(ploidy=-1)))` then?  I think ploidy chunks of size one could actually be pretty common when concatenating haplotype arrays.

I could also see the argument in leaving it up to the user to interpret those error messages and rechunk themselves so that they have to think through the performance implications, but I had already started being defensive about that in the GWAS regression methods.  My rationale was that it's better to optimize for a less frustrating user experience than for making performance concerns prominent (and we could always add warnings for that later).

sgkit/stats/aggregation.py
eric-czech: 👍

sgkit/stats/aggregation.py
eric-czech: Ohh nm, I see why it doesn't matter now (and the test for it).  Ignore that.

sgkit/stats/aggregation.py
```diff
@@ -1,10 +1,103 @@
...
+        "void(numba.int8[:], numba.uint8[:], numba.uint8[:])",
+        "void(numba.int16[:], numba.uint8[:], numba.uint8[:])",
+        "void(numba.int32[:], numba.uint8[:], numba.uint8[:])",
+        "void(numba.int64[:], numba.uint8[:], numba.uint8[:])",
+    ],
+    "(k),(n)->(n)",
+    nopython=True,
+)
+def count_alleles(g: ArrayLike, _: ArrayLike, out: ArrayLike) -> None:
+    """Generaliszed U-function for computing per sample allele counts.
```
eric-czech, line 21: nit: spelling

sgkit/stats/aggregation.py
```diff
@@ -1,10 +1,103 @@
...
+        corresponding to the number of non-missing occurrences
+        of each allele.
+
+    Examples
+    --------
+
+    >>> import sgkit as sg
+    >>> from sgkit.testing import simulate_genotype_call_dataset
+    >>> ds = simulate_genotype_call_dataset(n_variant=4, n_sample=2, seed=1)
+    >>> ds['call_genotype'].to_series().unstack().astype(str).apply('/'.join, axis=1).unstack() # doctest: +NORMALIZE_WHITESPACE
```
eric-czech, line 68: This is a good place for @tomwhite's code in https://github.com/pystatgen/sgkit/pull/58 now.

sgkit/stats/aggregation.py
timothymillar: Thanks, not my strong point

eric-czech:
This is good to go as far as I'm concerned.  @tomwhite / @ravwojdyla / @jeromekelleher could one of you take a look as well?  Two+ approvals seems appropriate for this one.

timothymillar:
Just for reference the following is a strait forward implementation only using `xarray` and `dask` but the performance is much worse that the gufunc version

```python
def count_call_alleles(ds):
    G = ds["call_genotype"]
    M = ds["call_genotype_mask"]
    n_variant, n_allele = G.shape[0], ds.dims["alleles"]
    max_allele = n_allele + 1
    G = xr.where(M, n_allele, G)
    G = G.expand_dims(dim="alleles", axis=-1)
    I = da.arange(max_allele, dtype='int8')
    A = G == I  # one-hot encoding of alleles
    AC = A.sum(axis=-2)
    return AC[..., 0:n_allele]
```

jeromekelleher (commented):
LGTM too, although I don't understand the details of how this is interacting with numba.
 
@tomwhite, I think you should have the final vote here.

tomwhite (approved):
+1 this looks great @timothymillar.

eric-czech:
> Just for reference the following is a strait forward implementation only using `xarray` and `dask`

Clever!  I bet that's still much faster than `da.apply_along_axis(np.bincount, axis=2)`.

Thanks again for picking this up @timothymillar, nicely done.  I'll set it to merge.

ravwojdyla:
@Mergifyio refresh

Issue #104, opened 2020-08-12
Read data from VCF files


hammer:
We can read data from VCF Zarr files with #40.

Ultimately we will need to read data from VCF files directly.

https://github.com/brentp/cyvcf2 looks to be a good option for this purpose.

Understanding our options here will also help us reach a conclusion on #65.

tomwhite:
I spent some time this week prototyping what this would look like.

The first piece is a function to read a (bgzip compressed) VCF file using cyvcf2 and convert it into sgkit xarray format. It does this by reading a chunk of N variants (N=10,000 for my experiment), creating an xarray dataset for those variants, then calling `to_zarr` on the dataset with `append_dim="variants"` to append to a Zarr store on disk. This is very similar to the way scikit-allel's `vcf_to_zarr` function is structured.

So far, everything is serial. To run in parallel, there needs to be a way to i) divide the VCF(s) into parts to be processed, and ii) merge the resulting Zarr stores.

For i), since VCF files are commonly tabix indexed, we could use genomic regions to partition the VCF(s) into roughly equal parts, each of which would be processed independently to create a Zarr store. We'd have to come up with some heuristics for doing a decent job on partitioning (e.g. [like this](https://davemcg.github.io/post/split-vcf-into-n-pieces-by-coordinate/)).

Then, for ii), the problem is that each Zarr store will have a different number of variants, and even though the chunk sizes are the same, they need to be rechunked in the variant dimension.

For the sake of example, if the chunk size was 10, and the first Zarr store had (3, ) variants, and the second had (10, 6) variants, then they would be merged and rechunked as (3+7, 3+6) = (10, 9) variants.

This can be achieved in Xarray with

```python
ds = xr.concat(datasets, dim="variants")
ds = ds.chunk({"variants": chunk_size})
ds.to_zarr(output, mode="w")
```

However, there may be some caveats with running at scale, as discussed in https://github.com/dask/dask/issues/5105 (by @alimanfoo!), which may need work (rechunker may be an option too). I haven't looked at performance, but VCF parsing should be very fast with cyvcf2, since it is a thin interface to htslib (see the paper: https://academic.oup.com/bioinformatics/article/33/12/1867/2971439).

This two step-scheme should scale well (using Dask), since it is basically two embarrassingly parallel steps, mediated by a set of intermediate Zarr stores.

There are some build issues we'd probably want to address too:
[...]

PR #116, opened 2020-08-17
Allow variant_id to be 'O' dtype to support variable length strings

### LINKED ISSUE #98: Support fixed and dynamic length allele strings in IO readers
We've made special cases in the plink/bgen readers to coerce alleles to a fixed length string type, e.g. https://github.com/pystatgen/sgkit-plink/issues/12.  This requires a full scan of the alleles to get the max length so it adds a minor delay to loading times, but makes storage in-memory or otherwise more compact if the allele lengths don't vary widely.  Apart from the delay for in building what would otherwise be a fully lazily defined data structure, it can also be a poor default.  The maximum allele size in UKB PLINK data is 256 so the fixed-length strings are suboptimal.

This has also come up with vcf here: https://github.com/pystatgen/sgkit/pull/40#issuecomment-669948502.

I propose that we change [create_genotype_call_dataset](https://github.com/pystatgen/sgkit/blob/master/sgkit/api.py#L58) to allow fixed or object types for alleles (as @alimanfoo suggested above) and make object type the default for alleles in IO readers.  As far as I know it will be more common to start a workflow with a mix of variant types and then possibly filter to SNVs so any conversion to fixed-length should take place after that.  Perhaps we can broaden the scope of https://github.com/pystatgen/sgkit/issues/90 a bit to make that function a part of the top-level API.

tomwhite:
This fixes #98 for `variant_id`, which I need for #104.

Should we change `sample_id` too, or do we always have the list of sample IDs up front?

### DIFF
--- a/sgkit/api.py
+++ b/sgkit/api.py
@@ -44,7 +44,7 @@ def create_genotype_call_dataset(
     call_genotype_phased : array_like, bool, optional
         A flag for each call indicating if it is phased or not. If
         omitted all calls are unphased.
-    variant_id: array_like, str, optional
+    variant_id: array_like, str or object, optional
         The unique identifier of the variant.
 
     Returns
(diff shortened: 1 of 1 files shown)


eric-czech:
> Should we change sample_id too, or do we always have the list of sample IDs up front?

I don't see why we wouldn't want to support both for IO functions (+1 to changing for sample_id too).

PR #214, opened 2020-08-27
Add missing API docs to Sphinx

### LINKED ISSUE #212: Published API docs are missing recent functions
They need to be added manually to `docs/api.rst`.

tomwhite:
We need to make sure we add new functions to `docs/api.rst`. Perhaps a note in the contributor guide?

This fixes the build, but some of the formatting is not quite right. We should also visually check the docs whenever we add new functions to check that links, formatting etc are OK.

Fixes #212

### DIFF
--- a/sgkit/stats/hwe.py
+++ b/sgkit/stats/hwe.py
@@ -145,7 +145,7 @@ def hardy_weinberg_test(
     Dataset
         Dataset containing (N = num variants):
         variant_hwe_p_value : (N,) ArrayLike
-            P values from HWE test for each variant as float in [0, 1].
+        P values from HWE test for each variant as float in [0, 1].
 
     References
     ----------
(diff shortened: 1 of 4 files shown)


sgkit/stats/hwe.py
```diff
@@ -156,8 +156,9 @@ def hardy_weinberg_test(
     Raises
     ------
     NotImplementedError
-        * If ploidy of provided dataset != 2
-        * If maximum number of alleles in provided dataset != 2
+        If ploidy of provided dataset != 2
+    NotImplementedError
```
eric-czech, line 160: I had originally been breaking multiple causes of the same error out like this but @ravwojdyla suggested collapsing them in a past PR.  @ravwojdyla did you have a reason for that?

tomwhite: Sphinx doesn't render the bullets correctly, which is why I changed this.

Issue #95, opened 2020-08-06
PCA User Story


jerowe:
I'm going to start working on the [PCA user story](https://github.com/pystatgen/sgkit-requirements/tree/master/stories/pop-structure-pca).

I've been back and forth with this on a few calls and internal discussion, but the general consensus seems to be that exploratory statistics and filtering should be separated out from the PCA functionality itself. It's already separated out in the codebase, but is grouped together in the docs as they would be together in a real world analysis.

Resources:

[Scikit Allel PCA Docs](https://scikit-allel.readthedocs.io/en/stable/stats/decomposition.html)
[Tour of Scikit Allel](http://alimanfoo.github.io/2016/06/10/scikit-allel-tour.html)
[Fast PCA Blog Post](http://alimanfoo.github.io/2015/09/28/fast-pca.html)

PCA analysis would be after LD pruning https://github.com/pystatgen/sgkit/issues/31

eric-czech:
Hey @jerowe by upstream did you mean after LD pruning?  We might have the wrong "stream" direction in mind but I wanted to verify that LD pruning would come first.  It's a minor point for the PCA story though if you were going to start with real data, it might be a good idea to use [locate_unlinked](https://scikit-allel.readthedocs.io/en/stable/stats/ld.html?highlight=locate_unlinked#allel.locate_unlinked) or sample with big fixed step windows (1 Mbp would do it if you wanted to try human data like 1000 genomes).

jerowe:
@eric-czech yes I had the wrong stream. Those always get me. ;-)

PCA is *after* LD pruning.

jerowe:
@alimanfoo @eric-czech (and whoever else wants to provide an opinion ;-) ) Questions!

I was digging into the PCA today. At first I was thinking just drop in dask_ml with some fancy doc string examples for those that are used to the , but then as I was looking at it I was wondering is that what we want? 

Do the genotypes ever get big enough that we wouldn't want to just calculate in memory?

Also, it was decided on the last call that everyone likes the interface over PCA that allele provides, so keep that, right?

jeromekelleher:
I think the consensus was that the interface that ``scikit-allel`` uses is pretty good, so we should just copy that in the first instance (@eric-czech - correct me if I'm wrong here!), with a straightforward implementation using upstream tools.  Getting that and some unit test coverage into the library would be a great step in the right direction I think. Is this something you'd like to work into a PR @jerowe?

eric-czech:
[...]

PR #227, opened 2020-09-02
PCA 

jerowe:
I couldn't push to pystatgen/sgkit so I'm reopening this PR here.

Continues - https://github.com/pystatgen/sgkit/pull/123

Linked Issue - https://github.com/pystatgen/sgkit/issues/95


jerowe:
Throwing this back in here to show that we still care.

https://github.com/dask/dask-ml/issues/731

https://github.com/pystatgen/sgkit/blob/6dcf446583bc52de5e15bae65f4aa63fcc5bb9b7/sgkit/tests/test_decomposition.py#L239

I have some tests demonstrating that dask-ml PCA fails based on the shape of the data.

eric-czech:
> I have some tests demonstrating that dask-ml PCA fails based on the shape of the data.

There are two problems I see with dask-ml PCA so far:

1. Sign determinacy support is broken when an array is wider than it is tall (https://github.com/dask/dask-ml/issues/732)
2. Dask deterministic svd doesn't support column chunking (https://github.com/dask/dask/issues/6590)

Both of those are problems we should solve regardless of whether or not we subclass, vendor in, or omit code in sgkit.  It's not looking like they'll be hard to fix upstream at the moment.

To me, the simplest way to support PCA is to:

1. Remove the dependency on dask-ml added in this PR
2. Remove everything but the PattersonScaler (and have it subclass from scikit-learn instead)
3. Add documentation and/or notebooks that show how to do PCA with these other tools

I can imagine that we might eventually find some reason to want to do PCA differently than how dask-ml and dask currently support it, but I can't see any reason why those things would only apply to genetics.  Let me know if you disagree @jeromekelleher / @jerowe.

I guess I wouldn't be too surprised if we one day vendor in a bunch of external code like this, though I'm struggling to find a reason why it's necessary now.

jerowe:
I'll go with whatever the vote is, but my personal vote is that we should support PCA. The GenotypePCA class 74 lines of code, with 9 of those being a call to super in the init function so we have the same defaults as the GenotypePCA. The rest is just porting over docstrings so it's obvious to users where the differences are.  I'd consider PCA to be a baseline in terms of core functionality myself. 

We can still subclass dask-ml PCA and just change the `_fit` to transpose the data the way scikit-allel does it. The problems in the dask-ml PCA lie where they run the `svd_flip` function.

eric-czech:
> We can still subclass dask-ml PCA and just change the _fit to transpose the data the way scikit-allel does it. The problems in the dask-ml PCA lie where they run the svd_flip function.

Transposing the input and supporting sign determinacy aren't specific to our domain though.  Fixing svd_flip (https://github.com/dask/dask-ml/issues/732) won't take much except a little time to figure out where https://github.com/dask/dask/issues/3576#issuecomment-685924337 is at.  I put in https://github.com/dask/dask/pull/6591 yesterday to try to address the deeper svd issue with column chunkings (unrelated to svd_flip).

If we would like to wrap this all up for users now, then this interface would be consistent with the rest of the API:

```python
from sklearn.base import BaseEstimator as Estimator

def pca(ds: Dataset, n_components: int, est: Optional[Estimator] = None, ...) -> Tuple[Dataset, Estimator]:
     # Make sure allele count is present
     if est is None:
         est = ... # Create and fit scaling / PCA pipeline estimator (sklearn or Dask-ML, depending on chunking)
    # Attach results to Dataset
    return ds, est
```

Then we could wrap the utilities rather than extending them in a way that involves copy+pasting so much code in order to work around domain-agnostic upstream limitations.

hammer:
Closing this in favor of #262

Issue #80, opened 2020-07-30
Create bitpacking filter for biallelic diploid datasets


eric-czech:
We should add a bitpacking numcodecs filter for biallelic diploid data since it makes for a substantial improvement over zarr's default compression.  Here's an example: https://nbviewer.jupyter.org/github/related-sciences/gwas-analysis/blob/master/notebooks/platform/xarray/io.ipynb (at the end, results are ~%40 smaller).

That same filter won't work since calls are in a 3D array now, so we'll have to rethink the packing scheme.

hammer:
@alimanfoo @daletovar we thought this task might be a good one for Quansight to pick up. Any interest?

jeromekelleher:
Sounds like a good idea to me. This does sound like something that should be in ``numcodecs``, though, right?

Also, it doesn't feel like this is on the critical path for getting an initial release done - it will make our on-file storage a bit more compact, but won't affect anything else much, right?

eric-czech:
> This does sound like something that should be in numcodecs, though, right?

Possibly?  An alternative that made less sense when we were working with PLINK data directly (as alternate allele counts) would be to encode our calls as booleans and use [PackBits](https://numcodecs.readthedocs.io/en/stable/packbits.html) as is.  This is pretty straightforward and putting a little thought into the syntax would be something like:

```python
# Convert  unphased diploid biallelic calls to 2-bit bools
hap0, hap1 = ds.call_genotype[..., 0], ds.call_genotype[..., 1]
bit0 = (hap0 < 0) | (hap0 != hap1) # Missing or heterozygous
bit1 = hap0 == 0 # Homozygous major or minor
ds['call_genotype_bits'] = xr.concat([bit0, bit1], dim='bits')
xr.to_zarr(ds, encoding={'call_genotype_bits': {'filters': [PackBits()]}})
```

I suspect there's a way to do it in a single pass without creating two in-memory arrays and one more generic interface that numcodecs could support to cover this would be to have a bit packing filter optimized for 8-bit integers in a pre-specified range (something like -1 to 2 in this case for the 4 PLINK call states).  We could call a function like https://github.com/pystatgen/sgkit/issues/85 first, then feed it to this filter.  Decoding back to the 3D format would still be non-trivial but we'll probably need a function for that at some point anyhow.

> Also, it doesn't feel like this is on the critical path for getting an initial release done - it will make our on-file storage a bit more compact, but won't affect anything else much, right?

[...]

PR #228, opened 2020-09-02
Add PC Relate

ravwojdyla:
Need to open a new PR to move to forked repo. Original #133 .

Changes:
 * add `pc_relate` method
 * add tests for `pc_relate`
 * add validation data generator for `pc_relate`
 * add `missing_pct` to `simulate_genotype_call_dataset`, to support missing data

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -11,6 +11,7 @@ from .model import (  # noqa: F401
 from .stats.aggregation import count_call_alleles, count_variant_alleles, variant_stats
 from .stats.association import gwas_linear_regression
 from .stats.hwe import hardy_weinberg_test
+from .stats.pc_relate import pc_relate
 from .stats.popgen import Fst, Tajimas_D, divergence, diversity
 from .stats.regenie import regenie
 
@@ -33,4 +34,5 @@ __all__ = [
     "divergence",
     "Fst",
     "Tajimas_D",
+    "pc_relate",
 ]
--- a/sgkit/stats/pc_relate.py
+++ b/sgkit/stats/pc_relate.py
@@ -1,0 +1,152 @@
+from typing import Tuple
+
+import dask.array as da
+import xarray as xr
+
+from sgkit.typing import ArrayLike
+from sgkit.utils import merge_datasets
+
+
+def gramian(a: ArrayLike) -> ArrayLike:
+    """Returns gramian matrix of the given matrix"""
+    return a.T.dot(a)
+
+
+def _impute_genotype_call_with_variant_mean(
+    call_g: xr.DataArray, call_g_mask: xr.DataArray
+) -> xr.DataArray:
+    call_g_present = ~call_g_mask  # type: ignore[operator]
+    variant_mean = call_g.where(call_g_present).mean(dim="samples")
+    imputed_call_g: xr.DataArray = call_g.where(call_g_present, variant_mean)
+    return imputed_call_g
+
+
+def _collapse_ploidy(ds: xr.Dataset) -> Tuple[xr.DataArray, xr.DataArray]:
+    call_g_mask = ds["call_genotype_mask"].any(dim="ploidy")
+    call_g = xr.where(call_g_mask, -1, ds["call_genotype"].sum(dim="ploidy"))  # type: ignore[no-untyped-call]
+    return call_g, call_g_mask
+
+
+def pc_relate(ds: xr.Dataset, maf: float = 0.01, merge: bool = True) -> xr.Dataset:
+    """Compute PC-Relate as described in Conomos, et al. 2016 [1].
+
+    This method computes the kinship coefficient matrix. The kinship coefficient for
+    a pair of individuals ``i`` and ``j`` is commonly defined to be the probability that
+    a random allele selected from ``i`` and a random allele selected from ``j`` at
+    a locus are IBD. Several of the most common family relationships and their
+    corresponding kinship coefficient:
+
+    +--------------------------------------------------+---------------------+
+    | Relationship                                     | Kinship coefficient |
+    +==================================================+=====================+
+    | Individual-self                                  | 1/2                 |
+    +--------------------------------------------------+---------------------+
+    | full sister/full brother                         | 1/4                 |
+    +--------------------------------------------------+---------------------+
+    | mother/father/daughter/son                       | 1/4                 |
+    +--------------------------------------------------+---------------------+
+    | grandmother/grandfather/granddaughter/grandson   | 1/8                 |
+    +--------------------------------------------------+---------------------+
+    | aunt/uncle/niece/nephew                          | 1/8                 |
+    +--------------------------------------------------+---------------------+
+    | first cousin                                     | 1/16                |
+    +--------------------------------------------------+---------------------+
+    | half-sister/half-brother                         | 1/8                 |
+    +--------------------------------------------------+---------------------+
+
+    Parameters
+    ----------
+    ds : Dataset
+        Dataset containing (S = num samples, V = num variants, D = ploidy, PC = num PC)
+
+        - genotype calls: "call_genotype" (SxVxD)
+        - genotype calls mask: "call_genotype_mask" (SxVxD)
+        - sample PCs: "sample_pcs" (PCxS)
+    maf : float
+        individual minor allele frequency filter. If an individual's estimated
+        individual-specific minor allele frequency at a SNP is less than this value,
+        that SNP will be excluded from the analysis for that individual.
+        The default value is 0.01. Must be between (0.0, 0.1).
+    merge : bool, optional
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset. Output variables will
+        overwrite any input variables with the same name, and a warning
+        will be issued in this case.
+        If False, return only the computed output variables.
+
+    Warnings
+    --------
+    This function is only applicable to diploid, biallelic datasets.
+
+    Returns
+    -------
+    Dataset
+        Dataset containing (S = num samples):
+        pc_relate_phi: (S,S) ArrayLike
+        pairwise recent kinship coefficient matrix as float in [-0.5, 0.5].
+
+    References
+    ----------
+    [1] - Conomos, Matthew P., Alexander P. Reiner, Bruce S. Weir, and Timothy A. Thornton. 2016.
+    "Model-Free Estimation of Recent Genetic Relatedness."
+    American Journal of Human Genetics 98 (1): 127–48.
+
+    Raises
+    ------
+    ValueError
+        If ploidy of provided dataset != 2
+    ValueError
+        If maximum number of alleles in provided dataset != 2
+    ValueError
+        Input dataset is missing any of the required variables
+    ValueError
+        If maf is not in (0.0, 1.0)
+    """
+    if maf <= 0.0 or maf >= 1.0:
+        raise ValueError("MAF must be between (0.0, 1.0)")
+    if "ploidy" in ds.dims and ds.dims["ploidy"] != 2:
+        raise ValueError("PC Relate only works for diploid genotypes")
+    if "alleles" in ds.dims and ds.dims["alleles"] != 2:
+        raise ValueError("PC Relate only works for biallelic genotypes")
+    if "call_genotype" not in ds:
+        raise ValueError("Input dataset must contain call_genotype")
+    if "call_genotype_mask" not in ds:
+        raise ValueError("Input dataset must contain call_genotype_mask")
+    if "sample_pcs" not in ds:
+        raise ValueError("Input dataset must contain sample_pcs variable")
+
+    call_g, call_g_mask = _collapse_ploidy(ds)
+    imputed_call_g = _impute_genotype_call_with_variant_mean(call_g, call_g_mask)
+
+    # 𝔼[gs|V] = 1β0 + Vβ, where 1 is a length _s_ vector of 1s, and β = (β1,...,βD)^T
+    # is a length D vector of regression coefficients for each of the PCs
+    pcs = ds["sample_pcs"]
+    pcsi = da.concatenate([da.ones((1, pcs.shape[1]), dtype=pcs.dtype), pcs], axis=0)
+    # Note: dask qr decomp requires no chunking in one dimension, and because number of
+    # components should be smaller than number of samples in most cases, we disable
+    # chunking on components
+    pcsi = pcsi.T.rechunk((None, -1))
+
+    q, r = da.linalg.qr(pcsi)
+    # mu, eq: 3
+    half_beta = da.linalg.inv(2 * r).dot(q.T).dot(imputed_call_g.T)
+    mu = pcsi.dot(half_beta).T
+    # phi, eq: 4
+    mask = (mu <= maf) | (mu >= 1.0 - maf) | call_g_mask
+    mu_mask = da.ma.masked_array(mu, mask=mask)
+    variance = mu_mask * (1.0 - mu_mask)
+    variance = da.ma.filled(variance, fill_value=0.0)
+    stddev = da.sqrt(variance)
+    centered_af = call_g / 2 - mu_mask
+    centered_af = da.ma.filled(centered_af, fill_value=0.0)
+    # NOTE: gramian could be a performance bottleneck, and we could explore
+    #       performance improvements like (or maybe sth else):
+    #       * calculating only the pairs we are interested in
+    #       * using an optimized einsum.
+    assert centered_af.shape == call_g.shape
+    assert stddev.shape == call_g.shape
+    phi = gramian(centered_af) / gramian(stddev)
+    # NOTE: phi is of shape (S x S), S = num samples
+    assert phi.shape == (call_g.shape[1],) * 2
+    new_ds = xr.Dataset({"pc_relate_phi": (("sample_x", "sample_y"), phi)})
+    return merge_datasets(ds, new_ds) if merge else new_ds
--- a/sgkit/tests/test_pc_relate.py
+++ b/sgkit/tests/test_pc_relate.py
@@ -1,0 +1,188 @@
+import numpy as np
+import pandas as pd
+import pytest
+import xarray as xr
+from hypothesis import given, settings
+from hypothesis.extra.numpy import arrays
+from sklearn.decomposition import PCA
+
+from sgkit.stats.pc_relate import (
+    _collapse_ploidy,
+    _impute_genotype_call_with_variant_mean,
+    gramian,
+    pc_relate,
+)
+from sgkit.testing import simulate_genotype_call_dataset
+
+
+def test_pc_relate__genotype_inputs_checks() -> None:
+    g_wrong_ploidy = simulate_genotype_call_dataset(100, 10, n_ploidy=3)
+    with pytest.raises(ValueError, match="PC Relate only works for diploid genotypes"):
+        pc_relate(g_wrong_ploidy)
+
+    g_non_biallelic = simulate_genotype_call_dataset(100, 10, n_allele=3)
+    with pytest.raises(
+        ValueError, match="PC Relate only works for biallelic genotypes"
+    ):
+        pc_relate(g_non_biallelic)
+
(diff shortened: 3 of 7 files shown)


ravwojdyla:
@eric-czech no implementation/logic changes since #133, just added more tests. Could you please take a look at the tests, and if there is anything more needed?

I added validation code to produce validation dataset BUT I have an issue, I would really like to use a "production-ish" dataset for validation, I noticed that zarr zip store, produced from sgkit_plink is actually almost twice the size of original uncompressed plink data. Do you have recommendations to keep the size for tests as small as possible?

eric-czech:
> Could you please take a look at the tests, and if there is anything more needed from test logic perspective?

Ok I'll take another look.  Can you zip up `test_pc_relate/zarr_data` to get the file count down some more?

> I noticed that zarr zip store, produced from sgkit_plink is actually almost twice the size of original uncompressed

I'm not surprised, something like https://github.com/pystatgen/sgkit/issues/80 was necessary in the past for me to produce zarrs smaller than uncompressed PLINK given how wasteful our representation is (we uses 8 times as many bytes as PLINK does for calls).

I would be surprised though if you couldn't set Blosc/ztsd compressors and get down to the same size with a higher clevel.  I have no sense of how much the zip compression works by default.

> this almost smells like a bug, but I'm not familiar with the sgkit_plink codebase

How so?  I can't see how they'd be connected unless the data written has a ton more entropy somehow (and is completely wrong) or the data types are getting promoted.

sgkit/stats/pc_relate.py
```diff
@@ -0,0 +1,141 @@
...
+
+    References
+    ----------
+    - [1] Conomos, Matthew P., Alexander P. Reiner, Bruce S. Weir, and Timothy A. Thornton. 2016.
+        "Model-Free Estimation of Recent Genetic Relatedness."
+        American Journal of Human Genetics 98 (1): 127–48.
+
+    Raises
+    ------
+    ValueError
```
eric-czech, line 89: These should be broken out like https://github.com/pystatgen/sgkit/pull/214#discussion_r478502734 (or tested on sphinx with a different bullet -- I'm not sure what would make it happy).

sgkit/stats/pc_relate.py
```diff
@@ -0,0 +1,141 @@
...
+        The default value is 0.01. Must be between (0.0, 0.1).
+
+
+    This method computes the kinship coefficient matrix. The kinship coefficient for
+    a pair of individuals ``i`` and ``j`` is commonly defined to be the probability that
+    a random allele selected from ``i`` and a random allele selected from ``j`` at
+    a locus are IBD. Several of the most common family relationships and their
+    corresponding kinship coefficient:
+
+    +--------------------------------------------------+---------------------+
```
eric-czech, line 52: 👍

eric-czech (approved):
Looks great @ravwojdyla!

Except for the minor docstring change for exceptions raised and saving the test data in a ZipStore, I think it's good to go.

ravwojdyla:
I wonder what you think about this solution @eric-czech, @tomwhite, @jeromekelleher, @alimanfoo, @hammer . To properly validate PR Relate, apart from unit tests, I want to use production public data but I don't want to check it into git repo (since it's 100s of MB), I also don't want to store the data/results in external systems (GCS/S3 etc, which likely would go stale), also I would like to "track" changes in the upstream reference R implementation. So here I suggest adding a "Validation suite" github workflow, that runs weekly or by manual trigger. 

This allows me to:
 * fetch production HapMap data from upstream
 * run up to date R reference implementation of PC Relate
 * compare our implementation against up to date results

Some points:
 * the workflow implementation is super [simple](https://github.com/pystatgen/sgkit/pull/228/commits/93206bb2d99caca587f1ffaa470733e9f0bb4eb8)
 * to fully run the validation (which includes building docker image, building R packages, installing up to date sgkit-plink, downloading HapMap data, R PC Relate, our PC Relate) we need around ~40 minutes, but since this is not a unit test, and it runs weekly, we don't need to worry about it much
 * right now the validation workflow, only includes PC Relate, but we could include more steps obviously, we could even use this pattern for benchmarks as well, an exciting validation could be to run full GWAS using Hail and sgkit and comparing them on weekly basis
 * the validation codebase requires only bash and docker, and runs the same on github action and locally: `./validation/gwas/method/pc_relate/run.sh` (which build the docker image, and runs itself inside the docker)
 * the validation tests, is a regular [pytest](https://github.com/pystatgen/sgkit/pull/228/files#diff-fcbefc8007ed54c68202f4521800d2d6), I used to work with a project where we had a nice integration between pytest and docker to run specific pytest inside specific docker images, if we like this pattern, I could reimplement it
 * I believe this would be a good signal for external people that we care about quality and that they can expect similar/same results as the tools they already use

Downsides:
 * there's a lot of moving steps involved in the process (as mentioned above), so in the future, there might be some failures in the validation of PC Relate, that are not related to changes in our repo, but for example change in R package, this is problematic BUT there is a bright side of it, it prevents the validation codebase from going stale (which is great!)

I added 7775e58 to show you have the validation run looks like, see the outputs here: https://github.com/pystatgen/sgkit/pull/228/checks?check_run_id=1075441878 (this commit would obviously not be merged i, I just wanted to validate the execution).

tomwhite:
This is great @ravwojdyla!

Validating the code in sgkit is probably the most important job this script does at the moment, since sgkit's code is changing rapidly (not PC relate specifically - more the other parts). So I think even having that by itself would be valuable. Tracking the R implementation may become onerous, and may not necessarily be something we want to do longer term - but we could easily freeze the version of the package we validate against.

> I also don't want to store the data/results in external systems (GCS/S3 etc, which likely would go stale)

The input data should be fixed, shouldn't it? So going stale isn't a concern. I think it might be preferable to host a copy that we control, for various reasons (reliability, not overloading the upstream source, etc).

ravwojdyla:
@tomwhite 

> but we could easily freeze the version of the package we validate against.

yep, I have not done that, because I actually would like to get an error if the upstream gets updated and fail. If we agree that we prefer to freeze versions, that's fine too, tho it would not fail on actual changes in the upstream implementation (without periodic manual update of versions).

>> I also don't want to store the data/results in external systems (GCS/S3 etc, which likely would go stale)

> The input data should be fixed, shouldn't it? So going stale isn't a concern. I think it might be preferable to host a copy that we control, for various reasons (reliability, not overloading the upstream source, etc).

Yes, the staleness I'm referring to is the codebase, in the past I have seen two ways of doing this:
 * you freeze data in say GCS (or similar) in known paths, and sometimes update (overwrite) that data - which makes the new code base work, but old might fail (this is a short term solutions usually)
 * you essentially reinvent some for of external resource tracking within git, that could be lightweight, but nevertheless still requires work. Essentially why LFS exists.

Normally if this was not an open source project, I would also lean towards hosting our own data, but since this is the open source realm, I have slight preference to use vanilla upstream as another signal that we test on actual public datasets etc. 

wdyt?

ravwojdyla:
And btw, I would also like to point out an interesting use of `git bisect` (already mentioned in #26) in the context of weekly validation codebase. Say we have this history:
 * on Sunday the validation runs - it's all green
 * during the following week we have bunch of changes to different parts of sgkit (all unit tests are green)
 * next Sunday, validation fails due to combo or one of the changes introduced throughout the week

Question: how do we know which change introduced the bug? `git bisect` (with linear git history that we keep since #26)  + validation suite comes to the rescue:
 * you mark the current master as `bad` by running validation suite and getting return code != 0
 * you mark previous Sunday as `good` by running validation suite and getting return code == 0
 * you let git automatically binary search for the commit that introduced the bug

You can read more about git bisect [here](https://stackoverflow.com/a/22592593/1661491). We could document this flow, when we get the validation flow in.

hammer:
I agree with Tom: we don't want to be in the business of tracking upstream code and data. It would be nice for us to just validate against a particular release or hash of an upstream tool and some data fixture produced by that tool.

ravwojdyla:
@hammer @tomwhite ok, if we want point in time validation, we can freeze all the versions and push the input data to public GCS bucket (or do you prefer LFS)? My preference is GCS (but there might be some need to configure access cost).

hammer:
I also prefer GCS.

ravwojdyla:
Also going forward if we like the Validation Suite approach, I will split this PR into PC Relate + unit tests (this PR), and a separate PR with validation suite + PC Relate validation.

ravwojdyla:
As mentioned above I am splitting this PR.

sgkit/stats/pc_relate.py
```diff
@@ -0,0 +1,144 @@
+from typing import Tuple
+
```
eric-czech, line 2: Add to top level api init?

sgkit/stats/pc_relate.py
```diff
@@ -0,0 +1,144 @@
...
+
+def pc_relate(ds: xr.Dataset, maf: float = 0.01) -> xr.Dataset:
+    """Compute PC-Relate as described in Conomos, et al. 2016 [1].
+
+    Parameters
+    ----------
+    ds : `xr.Dataset`
+        Dataset containing (S = num samples, V = num variants, D = ploidy, PC = num PC):
+        * genotype calls: "call_genotype" (SxVxD)
```
eric-czech, line 36: I'm not sure how this will render in sphinx, but it would probably be a good idea to use the same convention as in the results section where the variables are listed the way parameters are (but with more indentation).  That seems to render at least somewhat nicely for https://pystatgen.github.io/sgkit/generated/sgkit.gwas_linear_regression.html#sgkit-gwas-linear-regression.

sgkit/tests/test_pc_relate.py
```diff
@@ -0,0 +1,127 @@
...
+    b = gramian(a)
+    assert np.allclose(b, b.T)
+
+
+def test_collapse_ploidy() -> None:
+    g = simulate_genotype_call_dataset(1000, 10, missing_pct=0.1)
+    assert g.call_genotype.shape == (1000, 10, 2)
+    assert g.call_genotype_mask.shape == (1000, 10, 2)
+
+    # Sprinkle some tests data, this is a bit verbose, but in tests verbosity is not bad:
```
eric-czech, line 64: nit: "Test individual cases" perhaps since "sprinkle some tests data" doesn't make sense (grammar-wise)?  Not sure you need the verbosity disclaimer, I certainly agree it's not bad.

sgkit/tests/test_pc_relate.py
```diff
@@ -0,0 +1,127 @@
...
+    assert call_g.isel(variants=2, samples=2) == 1
+    assert call_g.isel(variants=3, samples=3) == -1
+    assert call_g_mask.isel(variants=1, samples=1) == 0
+    assert call_g_mask.isel(variants=3, samples=3) == 1
+
+
+def test_impute_genotype_call_with_variant_mean() -> None:
+    g = simulate_genotype_call_dataset(1000, 10, missing_pct=0.1)
+    call_g, call_g_mask = _collapse_ploidy(g)
+    # Sprinkle some tests data
```
eric-czech, line 93: nit: Like above, not quite sure what that's getting at

sgkit/tests/test_pc_relate.py
```diff
@@ -0,0 +1,127 @@
...
+    assert imputed_call_g.isel(variants=2, samples=1) == 2
+    assert (imputed_call_g.isel(variants=2, samples=slice(2, 9)) == 1).all()
+    assert (imputed_call_g.isel(variants=2, samples=[0, 9]) == (7 + 2) / 8).all()
+
+
+def test_pc_relate__values_within_range() -> None:
+    n_samples = 100
+    g = simulate_genotype_call_dataset(1000, n_samples)
+    call_g, _ = _collapse_ploidy(g)
+    pcs = PCA(call_g, ncomp=2).loadings
```
eric-czech, line 108: I took a look at the satsmodels code/docs and this would give the eigenvectors from a decomposition on input data where the samples are features rather than observations.  It happens to have the same dimensions but is not the same as the projection of the samples into the PC space implied by a decomposition on the transposed inputs (i.e. samples x variants rather than variants x samples like `call_g`).  It's confusing because sometimes people refer to those projections as "loadings" as well, but "scores" is more common.  And the `pcs` should be scores rather than loadings -- I haven't checked GENESIS but I took a look at Hail to be certain.  It also doesn't really make sense to use scaled eigenvectors from the transposed problem for anything unless those eigenvectors come from left-singular vectors instead of right-singular vectors (they'll be right singular when coming from `PCA`).

I would suggest using either one of these to generate scores instead:

```python
# `x` is should have shape (samples, variants) for us

def pc_scores_numpy(x, k):
    x = x - x.mean(axis=0, keepdims=True)
    u, s, v = np.linalg.svd(x, full_matrices=False)
    r = u * s[np.newaxis]
    return r[:, :k]

def pc_scores_sklearn(x, k):
    return PCA(n_components=k, svd_solver='full').fit_transform(x)

# Then in the test:
# pcs = pc_scores(call_g.T.values, 2)
# ds["sample_pcs"] = (("components", "samples"), pcs.T)
```

This was all fresh since I was doing a lot of work with SVD this week, but I wanted to double check the above as well as the usual "loadings" definitions so I verified in this gist: https://gist.github.com/eric-czech/52a04007f4cfb8f05b312079ca770add

eric-czech: Oh also I found that the statsmodels.PCA was super slow even with small data, so those definitely speed things up a lot.

sgkit/tests/test_pc_relate.py
```diff
@@ -0,0 +1,127 @@
...
+    g["sample_pcs"] = (("components", "samples"), pcs.T)
+    phi = pc_relate(g)
+    assert phi.pc_relate_phi.shape == (n_samples, n_samples)
+    data_np = phi.pc_relate_phi.data.compute()  # to be able to use fancy indexing below
+    upper_phi = data_np[np.triu_indices_from(data_np, 1)]
+    assert (upper_phi > -0.5).all() and (upper_phi < 0.5).all()
+
```
eric-czech, line 115: I think a useful test here would be to try to at least pick out some parent/child relationships since they're fairly easy to simulate.  I was tinkering with this a bit and it snowballed into this:

```python
# Create a dataset that is 2/3 founds and 1/3 progeny
rs = np.random.RandomState(1)
ds = simulate_genotype_call_dataset(1000, 300, seed=1)
ds['sample_type'] = xr.DataArray(np.repeat(['mother', 'father', 'child'], 100), dims='samples')
sample_groups = ds.groupby('sample_type').groups

def simulate_new_generation(ds):
    # Generate progeny genotypes as a combination of randomly 
    # selected haplotypes from each parents
    idx = sample_groups['mother'] + sample_groups['father']
    gt = ds.call_genotype.isel(samples=idx).values
    idx = rs.randint(0, 2, size=gt.shape[:2])
    # Collapse to haplotype along ploidy dim using indexer
    # * shape = (samples, variants)
    ht = gt[np.ix_(*map(range, gt.shape[:2])) + (idx,)].T
    gt_child = np.stack([ht[sample_groups[t]] for t in ['mother', 'father']]).T
    ds['call_genotype'].values = np.concatenate((gt, gt_child), axis=1)
    return ds

# Redefine the progeny genotypes
ds = simulate_new_generation(ds)

# Infer kinship
call_g, _ = _collapse_ploidy(ds)
pcs = pc_scores(call_g.T.values, 1)
ds["sample_pcs"] = (("components", "samples"), pcs.T)
ds["pc_relate_phi"] = pc_relate(ds)["pc_relate_phi"].compute()

# Check that all coefficients are in expected ranges
cts = (
    ds["pc_relate_phi"].to_series().reset_index()
    .pipe(lambda df: df.loc[df.sample_x >= df.sample_y]['pc_relate_phi'])
    .pipe(
        pd.cut, 
        bins=[p for phi in [0, .25, .5] for p in [phi-.1, phi+.1]],
        labels = ['unrelated', 'unclassified', 'parent/child', 'unclassified', 'self'],
        ordered=False
    )
    .value_counts()
)
print(cts)
# unrelated       44650
# self              300
# parent/child      200
# unclassified        0
assert cts['parent/child'] == len(sample_groups['child']) * 2
assert cts['self'] == ds.dims['samples']
assert cts['unclassified'] == 0
```

That produces kinship estimates with pretty well segregated distributions, which you can tighten up a bit by adding more variants to the simulation:

Screen Shot 2020-09-12 at 2 41 15 PM

eric-czech (changes requested):
I pulled the code down to experiment with it a bit and it helped me find a thing or two -- nothing major except for the PCA applications in the tests.  I think that extra test I proposed would be a good addition too, in order to challenge the implementation a bit more without being slow / overly complicated.

sgkit/stats/pc_relate.py
ravwojdyla: added

sgkit/stats/pc_relate.py
ravwojdyla: fixed

sgkit/tests/test_pc_relate.py
ravwojdyla: fixed used "Test individual cases"

sgkit/tests/test_pc_relate.py
ravwojdyla: fixed

sgkit/tests/test_pc_relate.py
ravwojdyla: Oh dang, I remember I meant to validate that, thanks for the catch! Also I did not want to add sklearn as another dependency, but I guess since it's dev/test deps, I guess that was a premature optimisation on my part. Fixed, used sklearn.

ravwojdyla:
Weird `test_read_vcfzarr` is failing 🤔:
> ValueError: Could not convert tuple of form (dims, data[, attrs, encoding]): (['variants'], False) to Variable.

ravwojdyla:
Ah interesting `dask` was released today version: 2.26.0 and it breaks our `test_read_vcfzarr`.

eric-czech (approved):

sgkit/stats/pc_relate.py
tomwhite: @ravwojdyla just wanted to check if you have tried running Sphinx doc for this PR?

sgkit/stats/pc_relate.py
```diff
@@ -0,0 +1,144 @@
...
+    return imputed_call_g
+
+
+def _collapse_ploidy(ds: xr.Dataset) -> Tuple[xr.DataArray, xr.DataArray]:
+    call_g_mask = ds["call_genotype_mask"].any(dim="ploidy")
+    call_g = xr.where(call_g_mask, -1, ds["call_genotype"].sum(dim="ploidy"))  # type: ignore[no-untyped-call]
+    return call_g, call_g_mask
+
+
+def pc_relate(ds: xr.Dataset, maf: float = 0.01) -> xr.Dataset:
```
tomwhite, line 29: We should add the `merge=True` parameter, but that can be done separately.

sgkit/__init__.py
```diff
@@ -1,7 +1,7 @@
-from .api import (  # noqa: F401
+from .api import DIM_SAMPLE  # noqa: F401
+from .api import (
     DIM_ALLELE,
     DIM_PLOIDY,
-    DIM_SAMPLE,
```
tomwhite, line 4: Nit: why did this change?

tomwhite (dismissed):

sgkit/stats/pc_relate.py
ravwojdyla: added

sgkit/__init__.py
ravwojdyla: must have been isort doing something weird, reverted to the original

ravwojdyla:
@tomwhite also added `pc_relate` to the doc api, and fixed some rst issues.

ravwojdyla:
@tomwhite is this good to merge?

tomwhite:
From my point of view, yes. Did @eric-czech approve this one?

eric-czech:
> Did @eric-czech approve this one?

I did but it looks like mergify dismissed it.  Re-approving now.

eric-czech (approved):

PR #236, opened 2020-09-07
Rename api.py

### LINKED ISSUE #91: Rename api.py
The API is not just in `api.py` anymore so we should rename it to `dataset.py` or similar.

tomwhite:
Fixes #91


jeromekelleher:
I'm not blown away by the new name, but I can't think of a better one either, so +0.

eric-czech (approved):

hammer:
This also appears stuck on Mergify. @jeromekelleher is Mergify causing problems for your other projects? I'm inclined to remove it from `sgkit` as we've had to kick it multiple times already.

jeromekelleher:
> @jeromekelleher is Mergify causing problems for your other projects? I'm inclined to remove it from sgkit as we've had to kick it multiple times already.

No, it's working great and has been a major time saver. Let's give it another few weeks, hopefully the config will settle down now.

Issue #24, opened 2020-07-08
PC-Relate implementation


ravwojdyla:
Relevant discussion: https://discourse.smadstatgen.org/t/pc-relate-experiment/
Code: https://github.com/related-sciences/gwas-analysis/tree/rav-pc_relate

Potential plan could be:
 1. move the code+tests over here with adjustments for the sgkit API
 2. improve upon the implementation (this could be tracked as a separate issue(s))

We can use this issue to track the work necessary to move it over here.

ravwojdyla:
@tomwhite I know you started reviewing the PC-Relate code - what's the state right now?

tomwhite:
My focus was on the locating the performance bottlenecks in the PC-Relate code (brief notes here: https://github.com/related-sciences/gwas-analysis/blob/rav-pc_relate-tw/notebooks/platform/xarray/pc_relate_perf.py#L3-L7).

I think your plan makes sense - move the code over to this repo once the sgkit API becomes a bit clearer.

ravwojdyla:
Thanks @tomwhite I see. Regarding perf/gramian issue, I thought it might also be interesting to try to represent it with optimised `einsum`, plus you've have mentioned potentially computing only the parts we are interested in, which is equally interesting! I bet there is lot we can improve - but yea, let's first move it in, and then optimise.

ravwojdyla:
FYI: I will go ahead and create a PR for the current state.

ravwojdyla:
Closing this via #228

PR #242, opened 2020-09-08
Validation Suite (PC Relate)

ravwojdyla:
Depends on #228 (related: #24)

See: https://github.com/pystatgen/sgkit/pull/228#issuecomment-687607928 for context.

Todo:
 - [x] merge #228 
 - [x] freeze all versions of R packages etc
 - [x] stage public data to a GCS bucket
 - [x] configure github actions to allow to read from the GCS bucket

This validation job currently tests sgkit_plink's `read_plink` + sgkit's `pc_relate` on HapMap data.

This is blocked on pystatgen/sgkit-plink#32.

### DIFF
--- a/validation/gwas/method/pc_relate/README.md
+++ b/validation/gwas/method/pc_relate/README.md
@@ -2,7 +2,7 @@ This runs test to validate our implementation gets the same
 exact results as the reference R implementation for production
 HapMap data.
 
-This code gets executed weekly as part of the Github Actions CI.
+This code is scheduled as part of the Github Actions CI.
 
 To run manually:
 
--- a/validation/gwas/method/pc_relate/README.md
+++ b/validation/gwas/method/pc_relate/README.md
@@ -4,7 +4,11 @@ HapMap data.
 
 This code is scheduled as part of the Github Actions CI.
 
-To run manually:
+To run manually, you need to first download the test data
+from `gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip`,
+keep in mind that `sgkit-data` GCS bucket uses
+[Requester Pays config](https://cloud.google.com/storage/docs/requester-pays),
+and the file size is about 32MB.
 
 ```bash
 gsutil -u $BILLING_PROJECT cp gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip /tmp/
--- a/validation/gwas/method/pc_relate/README.md
+++ b/validation/gwas/method/pc_relate/README.md
@@ -1,0 +1,19 @@
+This runs test to validate our implementation gets the same
+exact results as the reference R implementation for production
+HapMap data.
+
+This code gets executed weekly as part of the Github Actions CI.
+
+To run manually:
+
+```bash
+gsutil -u $BILLING_PROJECT cp gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip /tmp/
+./run.sh /tmp/hapmap_JPT_CHB_r23a_filtered.zip
+```
+
+`run.sh` will:
+ * convert plink data to GDS
+ * run reference [R PC-Relate implementation](pc_relate.R)  
+ * run [our PC-Relate and compare results](validate_pc_relate.py)
+
+The only requirement is that you have Docker and Bash installed.
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -1,0 +1,23 @@
+name: Validation suite
+
+on:
+  schedule:
+    # Run at the end of every day
+    - cron: "0 0 * * *"
+  # manual trigger
+  workflow_dispatch:
+
+jobs:
+  validation_suite:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+      with:
+        project_id: ${{ secrets.GCP_PROJECT_ID }}
+        service_account_key: ${{ secrets.GCP_SA_KEY }}
+        export_default_credentials: true
+    - name: Download public test data (real HapMap data)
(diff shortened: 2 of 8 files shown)


ravwojdyla:
Validated that the workflow *should* work via [act](https://github.com/nektos/act):

```
act -sGCP_SA_KEY -sGCP_PROJECT_ID -P ubuntu-latest=nektos/act-environments-ubuntu:18.04 -j validation_suite
```

jeromekelleher (approved):
LGTM. Pretty solid validation.

eric-czech:
Looks good to me too, that gcloud github action is very cool.

It is nagging me though that the last part isn't a unit test.  I can imagine that if we adopt this approach more broadly, we will have bugs we don't know about for up to a week and that it will be hard to debug them since it would take some work to set this up to be interactive.  Or at least I'm imagining that if we accidentally make a change that alters our pc-relate (far more likely than a meaningful change in GENSIS no doubt), I can see myself copying/pasting the `test_same_as_the_reference_implementation` code locally with a copy of that csv to figure out what the difference is.

Do you think it would make more sense @ravwojdyla / @jeromekelleher if this process simply produced the data we validate in unit tests on a schedule rather than being our only automated way of checking against it too?

If we did the same with say REGENIE, I would definitely want an easy way to fire up the debugging context since the comparable `test_same_as_the_reference_implementation` test for it is the most useful one by far and also much more likely to break than the others.

jeromekelleher:
> Do you think it would make more sense @ravwojdyla / @jeromekelleher if this process simply produced the data we validate in unit tests on a schedule rather than being our only automated way of checking against it too?

Saving the CSVs makes sense, I guess, but we do have to take into account how long things need to run for unit tests. There's a balance here between making the validation realistic and the sub-second run times we need for unit tests. On the whole, I think I'd be in favour of more realistic validations even if that does mean we have a lag between some problems showing up. There's also the size of the repo to consider, which already seems to be getting pretty lardy.

There should be some simple way of running these validations locally though, I agree.

ravwojdyla:
> Or at least I'm imagining that if we accidentally make a change that alters our pc-relate (far more likely than a meaningful change in GENSIS no doubt), I can see myself copying/pasting the test_same_as_the_reference_implementation code locally with a copy of that csv to figure out what the difference is.

@eric-czech @jeromekelleher to clear, to run this manually you don't need to copy anything, just follow this [doc](https://github.com/pystatgen/sgkit/blob/c5bc07ae9309e0c004eeb10fa5c2538ef70f421f/validation/gwas/method/pc_relate/README.md):

```
gsutil -u $BILLING_PROJECT cp gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip /tmp/
./run.sh /tmp/hapmap_JPT_CHB_r23a_filtered.zip
```

I think there is space for unit tests and there is space for more integration/validation style tests (for example right now this test is catching the bug that `sgkit_plink` is broken via pystatgen/sgkit-plink#32). And I would argue that we should have both. Unit tests should catch any bugs introduced in `pc_relate`, and this validation is just an extra safety net (+ documentation, that should not go stale because it's being run regularly). Plus if we want nothing stops us from running this more often than weekly (daily? etc)

> Do you think it would make more sense @ravwojdyla / @jeromekelleher if this process simply produced the data we validate in unit tests on a schedule rather than being our only automated way of checking against it too?

The test data in this case is in 100s of MB, we can't stage that in the git repo, we would need to use LFS (which bunch of people, including me, had bad experience with), so we would probably need to stage that on GCS, then we need to fetch data from GCS in unit test, so we probably need to build caching for it as well, and so on, I'm not sure if that is worth it.

> If we did the same with say REGENIE, I would definitely want an easy way to fire up the debugging context since the comparable test_same_as_the_reference_implementation test for it is the most useful one by far and also much more likely to break than the others.

If you run this locally (as documented above, you can definitely connect debugger there). Also I think in the case of REGENIE, your test data is relatively small, so that is fine, but maybe we could add a bigger validation test for REGENIE  and test against GLOW or the cpp implementation.

What do you think @eric-czech @jeromekelleher ?

eric-czech:
> The test data in this case is in 100s of MB, we can't stage that in the git repo

Ah ok I see.

> Plus if we want nothing stops us from running this more often than weekly (daily? etc)

Probably better than randomly sampling to a relatively tiny subset of the same data for a unit test?  I would definitely have a preference for that over trying to fetch big datasets from GCS in the build.  I suppose we can cross that bridge later if setting up a local process ever becomes time consuming.

eric-czech (approved):

ravwojdyla:
> Probably better than randomly sampling to a relatively tiny subset of the same data for a unit test?

I believe so, and even with sampling, it's not enough to sample the output data in this case, the input is large as well, and if we sample input, we no longer actually validating against "real world data" anymore. I'm happy to change to schedule to daily, will push an update to this PR.

validation/gwas/method/pc_relate/README.md
```diff
@@ -0,0 +1,19 @@
...
+HapMap data.
+
+This code is scheduled as part of the Github Actions CI.
+
+To run manually:
+
+```bash
+gsutil -u $BILLING_PROJECT cp gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip /tmp/
+./run.sh /tmp/hapmap_JPT_CHB_r23a_filtered.zip
+```
```
tomwhite, line 12: Might be worth adding a comment about requester-pays and the data size to the README.

tomwhite (approved):
Excellent +1

validation/gwas/method/pc_relate/README.md
ravwojdyla: Fixed

ravwojdyla:
Good to merge?

jeromekelleher:
Three +1s here, so marking for merge.

ravwojdyla:
@jeromekelleher I'm sorry you need to go through this, but his one also changes github workflows -> needs manual merge please.

jeromekelleher:
@mergifyio update

jeromekelleher:
Ok, thanks @ravwojdyla - I'll merge once CI passes again after update

Issue #245, opened 2020-09-12
Dask 2.26.0 breaks test_read_vcfzarr


ravwojdyla:
`dask` was released yesterday version: [2.26.0](https://pypi.org/project/dask/) and it breaks our `test_read_vcfzarr`. That same test works fine on `2.25.0`.

_Originally posted by @ravwojdyla in https://github.com/pystatgen/sgkit/pull/228#issuecomment-691538352_

The issue seems to be that:

```
    ds["variant_id_mask"] = (
        [DIM_VARIANT],
        variants_id == ".",
    )
```

returns `(['variants'], False)` instead of array of bool as the 2nd element.

ravwojdyla:
Further:

```
import dask.array as da

a = da.from_array(["a", "b", ".", "e"])
r = a == "."
print(r)

# 2.26.0: False
# 2.25.0: dask.array<eq, shape=(4,), dtype=bool, chunksize=(4,), chunktype=numpy.ndarray>
```

That's rather big change.

ravwojdyla:
Dask issue: https://github.com/dask/dask/issues/6631

ravwojdyla:
was fixed via #246 closing

PR #246, opened 2020-09-12
Workaround dask#6631

ravwojdyla:
Re: #245


eric-czech (approved):

eric-czech:
heh I have to believe that's not a change they intended.

ravwojdyla:
@eric-czech this is definitely a bug, see https://github.com/dask/dask/issues/6631 for an analysis, this essentially broke all dunder ops between Dask Array and non-scalar ...

eric-czech:
Crazy that it didn't break a test somewhere!

tomwhite:
We could also exclude Dask version 2.26.0, but this works too.

I'm going to merge this manually now so we can move forward.

PR #249, opened 2020-09-14
Remove "remove outdated reviews" from mergify.

### LINKED ISSUE #248: Remove "remove outdated reviews" mergify option
I think we can remove the "remove outdated reviews" mergify option as it's very twitchy. Every time someone rebases it invalidates reviews. I think this will cause many more problems than the rare occasion that someone radically changes a PR after it has been approved. I've just reapproved several PRs several times now, and it's pretty tedious.

jeromekelleher:
Closes #248

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -11,11 +11,6 @@ pull_request_rules:
       merge:
         strict: smart+fasttrack
         method: rebase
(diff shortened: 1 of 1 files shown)


ravwojdyla (approved):
sure

PR #250, opened 2020-09-14
Add merge parameter to method functions

### LINKED ISSUE #103: Append output variables from functions to input dataset

tomwhite:
Fixes #103

### DIFF
--- a/sgkit/stats/association.py
+++ b/sgkit/stats/association.py
@@ -8,7 +8,7 @@ from dask.array import Array, stats
 from xarray import Dataset
 
 from ..typing import ArrayLike
-from ..utils import merge_datasets
+from ..utils import conditional_merge_datasets
 from .utils import concat_2d
 
 
@@ -218,4 +218,4 @@ def gwas_linear_regression(
             "variant_p_value": (("variants", "traits"), res.p_value),
         }
     )
-    return merge_datasets(ds, new_ds) if merge else new_ds
+    return conditional_merge_datasets(ds, new_ds, merge)
--- a/sgkit/stats/association.py
+++ b/sgkit/stats/association.py
@@ -152,10 +152,9 @@ def gwas_linear_regression(
         Add intercept term to covariate set, by default True.
     merge : bool, optional
         If True (the default), merge the input dataset and the computed
-        output variables into a single dataset. Output variables will
-        overwrite any input variables with the same name, and a warning
-        will be issued in this case.
-        If False, return only the computed output variables.
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
 
     Warnings
     --------
--- a/sgkit/stats/association.py
+++ b/sgkit/stats/association.py
@@ -8,6 +8,7 @@ from dask.array import Array, stats
 from xarray import Dataset
 
 from ..typing import ArrayLike
+from ..utils import merge_datasets
 from .utils import concat_2d
 
 
@@ -116,6 +117,7 @@ def gwas_linear_regression(
     covariates: Union[str, Sequence[str]],
     traits: Union[str, Sequence[str]],
     add_intercept: bool = True,
+    merge: bool = True,
 ) -> Dataset:
     """Run linear regression to identify continuous trait associations with genetic variants.
 
@@ -148,6 +150,12 @@ def gwas_linear_regression(
         and concatenated to any 1D traits along the second axis (columns).
     add_intercept : bool, optional
         Add intercept term to covariate set, by default True.
+    merge : bool, optional
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset. Output variables will
+        overwrite any input variables with the same name, and a warning
+        will be issued in this case.
+        If False, return only the computed output variables.
 
     Warnings
     --------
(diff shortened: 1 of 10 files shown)


sgkit/stats/association.py
```diff
@@ -203,10 +211,11 @@ def gwas_linear_regression(
...
     res = linear_regression(G.T, X, Y)
-    return xr.Dataset(
+    new_ds = xr.Dataset(
         {
             "variant_beta": (("variants", "traits"), res.beta),
             "variant_t_value": (("variants", "traits"), res.t_value),
             "variant_p_value": (("variants", "traits"), res.p_value),
         }
     )
+    return merge_datasets(ds, new_ds) if merge else new_ds
```
jeromekelleher, line 221: I wonder if it's worth having a function that does the conditional merge? I can see it being a bit error-prone having this logic in every function, maybe:
```python
```

sgkit/stats/hwe.py
```diff
@@ -137,6 +139,12 @@ def hardy_weinberg_test(
         where `N` is equal to the number of variants and the 3 columns contain
         heterozygous, homozygous reference, and homozygous alternate counts
         (in that order) across all samples for a variant.
+    merge : bool, optional
+        If True (the default), merge the input dataset and the computed
```
jeromekelleher, line 143: I don't like the idea of duplicating this block of text in every function. Maybe we could refer to a section of the docs discussing what merge is, something like "If True (the default) merge the input dataset and the computed output variables into a single dataset. See :ref:`sec_dataset_merge` for more details".

Or, we could try to define a sphinx substitution or something?

jeromekelleher (approved):
Some minor comments.

sgkit/stats/association.py
tomwhite: See https://github.com/pystatgen/sgkit/pull/250/commits/b634bd71dd70aaa3199399bf6e2c179c9eb4e6d2

sgkit/stats/hwe.py
tomwhite: I prefer having a reference. Implemented in https://github.com/pystatgen/sgkit/pull/250/commits/cfc5e2f00d16d282e581ce6b707f08d43e161c63.

jeromekelleher:
LGTM - @eric-czech, @ravwojdyla, it'd be good to get your eyes on this before we start duplicating the pattern.

eric-czech (approved):
LGTM

ravwojdyla (changes requested):
@tomwhite could you please also update `pc_relate`?

ravwojdyla:
@tomwhite thanks, `pc_relate` still has the old docstring, could you please update that as well?

tomwhite:
Oops - should be fixed now.

ravwojdyla (approved):

PR #253, opened 2020-09-14
Add missing typing_extensions to setup.cfg

ravwojdyla:
Example of failure (from sgkit_vcf): https://github.com/pystatgen/sgkit-vcf/pull/11/checks?check_run_id=1112654920, reason `typing_extensions` are not in the ~requirements.~ setup.cfg.

Re: #255

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ install_requires =
     scipy
     zarr
     numba
+    typing-extensions
     setuptools >= 41.2  # For pkg_resources
 setup_requires =
     setuptools >= 41.2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,4 +30,4 @@ repos:
     hooks:
       - id: mypy
         args: ["--strict", "--show-error-codes"]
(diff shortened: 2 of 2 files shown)


ravwojdyla:
Looks like py3.7 doesn't have Literal, I might need to add `typing_extensions` to requirements in the end.

setup.cfg
```diff
@@ -31,6 +31,7 @@ install_requires =
     scipy
     zarr
     numba
+    typing-extensions
```
eric-czech, line 34: This doesn't need to go into requirements.txt or the mypy requirements?

eric-czech (approved):

setup.cfg
ravwojdyla: @eric-czech it's already in the `requirements.txt` (thus this bug), but good point about `mypy requirements`, missed that one. We should really automated this via https://github.com/pystatgen/sgkit/issues/39

ravwojdyla: Fixed

jeromekelleher:
Seems uncontroversial so merging.

Issue #240, opened 2020-09-08
Update Fst and Tajima's D functions to use grouping convention


jeromekelleher:
#100 added functions ``Fst`` and ``Tajimas_D`` which take two datasets as input. However, in #224  we discuss how the convention should be to use a grouping variable within the dataset to do this. Once the convention has been established, update these functions to use it.

Related: #239, #238, #224, #226, #225

tomwhite:
Fixed in #260

Issue #224, opened 2020-09-01
Method for grouping samples which is understood by library functions


jeromekelleher:
Population structure analysis depends on us being able to assign labels to samples. In [this thread](https://github.com/pystatgen/sgkit/pull/100#discussion_r467895101) we discussed the possibility of using individual datasets, one for each group, but grouping by a categorical variable is much more flexible and idiomatic.

We need to develop some conventions which allows the user to do some standard grouping of samples, and functions which use these groupings (``Fst``, or ``divergence``, for example), should understand these conventions and update the output dataset accordingly.

It may also be worth thinking about how we might group the variants dimension while we're thinking about it. For example, we might want to get the pairwise Fst values for all pairs of populations, for all chromosomes in a dataset. It would be super-nice if we had an idiomatic way of running these calculations.

@eric-czech - you had some concrete ideas on this in the earlier thread, what are your thoughts?

eric-czech:
> @eric-czech - you had some concrete ideas on this in the earlier thread, what are your thoughts?

Using `sample_group` as the default variable name along with a `groups` dimension as the convention would align nicely with our other naming conventions.  I don't see too much harm in appropriating the "groups" name for the samples dimension since it will be less common with variants.  It would be nice if there was something more specific like "populations" since up until now dimension names have been fairly unambiguous.  Using "groups" as the dimension name would strengthen the case for making both variable and dimension names configurable.

Renaming dimensions to meet the requirements of our functions isn't hard either so this choice isn't that important IMO, but I would mildly favor something like `cohorts` (variable = `sample_cohort`) rather than `groups`.  Do you see any issues with that @jeromekelleher?

As far as actual computations are concerned, I think statistics that can be computed for cohorts independently from one another are a great fit for the `.groupby.map` idiom.  

For something like computing pairwise stats for sample groups by chromosome, this is the most idiomatic approach I can think of if the sample sizes aren't equal and there is no guarantee on sorting:

```python
import itertools
import xarray as xr
import numpy as np
from sgkit.testing import simulate_genotype_call_dataset

# Simulate 2 contigs
ds = simulate_genotype_call_dataset(n_variant=100, n_sample=30, n_contig=2, seed=0)
# Create 3 arbitrary cohorts
ds['sample_cohort'] = xr.DataArray(np.repeat([0, 1, 2], ds.dims['samples'] // 3), dims='samples')

def fn2(g):
    g1, g2 = g
[...]

PR #260, opened 2020-09-15
Cohort grouping for popgen

tomwhite:
This is an initial attempt at #240, which I'm posting as a draft for discussion.

The main idea (following @eric-czech's https://github.com/pystatgen/sgkit/issues/224#issuecomment-685061129) is to pass a single dataset to the `diversity`, `divergence`, `Fst` functions that includes a `sample_cohort` variable that indicates which cohort each sample belongs to (using an index, 0 = first cohort, 1 = second cohort, etc). Then, there is a new `count_cohort_alleles` function that counts the number of each allele for each variant and cohort, and which is used by `diversity` etc to perform computations within/between cohorts.

A few questions for discussion:
* What's a good implementation for `count_cohort_alleles`? I tried one using `groupby`, and one that should perform better using Dask's `map_blocks` (but that could also be much improved).
* How does `sample_cohort` get specified? The tests just set it manually at the moment.
* Do we need to consider more than two cohorts yet?

Also, there's still more to do to get some of the tests passing, and to convert `Tajimas_D` to use the same approach.


jeromekelleher:
Looks good, thanks @tomwhite. Pinging @aktech for input.

eric-czech:
> What's a good implementation for count_cohort_alleles? I tried one using groupby, and one that should perform better using Dask's map_blocks (but that could also be much improved).

Hm I haven't tested this but it's what came to mind first as a possibility, borrowing some tricks from regenie and Tim's [PR](https://github.com/pystatgen/sgkit/pull/114/files):

```python
# n = samples, c = cohorts, k = alleles
@numba.guvectorize(['void(numba.int8[:, :], numba.int32[:], numba.int32[:], numba.int32[:,:])'], '(n, k),(n),(c)->(c,k)')
def count_cohort_alleles(ac, cohorts, _, out):
    out[:,:] = 0 # This is (cohorts, alleles)
    for i in range(ac.shape[0]):
        out[cohorts[i]] += ac[i]

ds = count_call_alleles(ds)
AC, SC = ds.call_allele_count, ds.sample_cohort
C = da.empty(n_cohorts, dtype=np.uint8)

# Call directly if there is no chunking in core dimensions

count_cohort_alleles(AC.chunk((None, -1, -1)), SC.chunk((-1,)), C)
# shape = (variants, cohorts, alleles)

# OR run on individual blocks for better scalability #

AC = da.map_blocks(count_cohort_alleles, AC, SC, C, chunks=(n_cohorts, n_alleles), ...)
# shape = (variants, cohorts * AC.numblocks[0], alleles * AC.numblocks[1])

# Stack the blocks and sum across them 
# (which will only work because each chunk is guaranteed to have same size)
AC = da.stack([AC.blocks[i] for i in list(np.ndindex(x.numblocks))]).sum(axis=0)
# shape = (variants, cohorts, alleles)
```

I would guess that this is a good bit better than any pure dask/xarray method if it works.

tomwhite:
That's great @eric-czech - thank you. I've incorporated the vectorized function in the latest update.

tomwhite:
In [today's call](https://discourse.pystatgen.org/t/20200917-developer-call-notes/97) we discussed what should happen for the n > 2 case. One option would be to have a new variable for each combination (named like `Fst_cohort0_cohort1`, etc), or just have a data array indexed by n cohort dimensions. The latter was thought to be preferable since you can use standard indexing to manipulate it, and xarray coordinates should allow indexing using cohort (population) labels, rather than integers. We should explore this in this PR.

jeromekelleher:
Thanks @tomwhite - I've written up my thoughts on it [here](https://github.com/pystatgen/sgkit/issues/224#issuecomment-694350208)

tomwhite:
Thanks @jeromekelleher. I have updated this PR with approach 2 from [your suggestion](https://github.com/pystatgen/sgkit/issues/224#issuecomment-694350208). Fst returns a square matrix, where the entries above the diagonal are the Fst values for cohort i and j.

Using xarray coordinates, we can refer to the values using cohort names. In this example the names are `co_0` and `co_1`, and the dimensions are `cohorts_a` and `cohorts_b`. (In general there could be any number of cohort names, but always just two dimensions for Fst.)

```python
>>> cohort_names = ["co_0", "co_1"]
>>> ds = ds.assign_coords({"cohorts_a": cohort_names, "cohorts_b": cohort_names})
>>> ds = Fst(ds) # compute Fst and return a new dataset
>>> ds["stat_Fst"]
<xarray.DataArray 'stat_Fst' (cohorts_a: 2, cohorts_b: 2)>
array([[       nan, 0.02538071],
       [       nan,        nan]])
Coordinates:
  * cohorts_a  (cohorts_a) object 'co_0' 'co_1'
  * cohorts_b  (cohorts_b) object 'co_0' 'co_1'
```

You can use the cohort names to get a particular Fst value:

```python
>>> ds["stat_Fst"].sel(cohorts_a="co_0", cohorts_b="co_1").values
0.02538071065989833
```

Similarly, `diversity` returns a 1D matrix, with dimension `cohorts`.

sgkit/stats/aggregation.py
```diff
@@ -167,6 +185,33 @@ def count_variant_alleles(ds: Dataset, merge: bool = True) -> Dataset:
...
+        AC, n_variants, n_cohorts * AC.numblocks[0], n_alleles * AC.numblocks[1]
+    )
+
+    # Stack the blocks and sum across them
+    # (which will only work because each chunk is guaranteed to have same size)
+    AC = da.stack([AC.blocks[:, i] for i in range(AC.numblocks[1])]).sum(axis=0)
+    assert_array_shape(AC, n_variants, n_cohorts, n_alleles)
+
+    new_ds = Dataset({"cohort_allele_count": (("variants", "cohorts", "alleles"), AC)})
+    return conditional_merge_datasets(ds, new_ds, merge)
```
aktech, line 212: @tomwhite A quick question, I am trying to understand what this function (`count_cohort_alleles`) is trying to do mathematically, is there a reference implementation or formula, which I can look at to understand this better? Or is it just a an attempt to do a better rewrite of an old implementation?

jeromekelleher:
This is very beautiful @tomwhite, I don't really have any substantive input. My only question really is, should we fill the lower diagonal of the output matrix also? ``Fst[0, 1]`` and ``Fst[1, 0]`` should be equal, and we're not saving any memory by not filling out the lower diagonal. It's great that xarray also fulfils the need for labelled cohorts.

sgkit/stats/aggregation.py
tomwhite: @aktech Good question (and this function definitely needs documentation!). It's needed because `count_call_alleles` and `count_variant_alleles` return allele counts for each sample and for all samples (respectively), whereas in this PR we need something between the two since we now have the concept of "cohort" (a set of samples): so we need per-cohort counts.

The counts array computed here is of shape `(n_variants, n_cohorts, n_alleles)`, so we have a count for each variant/cohort/allele combination. Have a look at the tests (e.g. `test_count_cohort_alleles__multi_variant_multi_sample`), and perhaps try calling the allele counting functions in a Python repl to get a feel for what they are doing.

tomwhite:
Thanks @jeromekelleher. I agree on the point about filling the lower diagonal - the only reason I haven't is because I'm not sure if xarray/dask would do the computations twice.

jeromekelleher:
> Thanks @jeromekelleher. I agree on the point about filling the lower diagonal - the only reason I haven't is because I'm not sure if xarray/dask would do the computations twice.

What about using numpy tricks like [these](https://stackoverflow.com/questions/16444930/copy-upper-triangle-to-lower-triangle-in-a-python-matrix)? That should avoid the idea of redoing the computations.

I guess as a general pattern we should just run the calculations again though - for some functions, ``F[a, b]`` will not be equal to ``F[b, a]``, and so maybe it's simpler to just do all N^2 calculations, and make the optimisation for symmetric functions as a second pass?

tomwhite:
This latest update uses a generalized ufunc for computing divergence, which should be more efficient than using `itertools`.

I think this is ready for review - changing the version of Fst (to hudson) can be done separately, as can being able to specify subsets of cohorts.

We might want to discuss the naming convention for cohort dimensions:
* `cohorts_a`, `cohorts_b` (current PR)
* `cohorts_1`, `cohorts_2` (natural alternative)
* `cohort_a`, `cohort_b` (but other dimensions are plural, e.g. `variants`)
* `cohort_1`, `cohort_2` (ditto)

jeromekelleher (approved):
Looks great!

sgkit/stats/aggregation.py
```diff
@@ -167,6 +200,53 @@ def count_variant_alleles(ds: Dataset, merge: bool = True) -> Dataset:
...
+    Dataset containing variable `call_allele_count` of allele counts with
+    shape (variants, cohorts, alleles) and values corresponding to
+    the number of non-missing occurrences of each allele.
+    """
+
+    n_variants = ds.dims["variants"]
+    n_alleles = ds.dims["alleles"]
+
+    ds = count_call_alleles(ds)
+    AC, SC = da.asarray(ds.call_allele_count), da.asarray(ds.sample_cohort)
```
eric-czech, line 229: Should we standardize on using `ds. call_allele_count` vs `ds['call_allele_count']`?  I've been trying to use the former in examples only and stick to the latter in code, but not for great reasons other than consistency and hedging against potentially having variable names that aren't legal attributes or conflict with something in the xarray namespace.  Presumably not using attributes becomes easier to standardize on with https://github.com/pystatgen/sgkit/pull/276.

sgkit/stats/aggregation.py
```diff
@@ -167,6 +200,53 @@ def count_variant_alleles(ds: Dataset, merge: bool = True) -> Dataset:
...
+def count_cohort_alleles(ds: Dataset, merge: bool = True) -> Dataset:
+    """Compute per cohort allele counts from genotype calls.
+
+    Parameters
+    ----------
+    ds
+        Genotype call dataset such as from
+        `sgkit.create_genotype_call_dataset`.
+    merge
+        (optional)
```
eric-czech, line 212: What should this `(optional)` indicate exactly?  I haven't been including it in many cases, but I don't know if we want it to mean:

1. It's a keyword-only argument
2. It's an argument with some default value
3. It can be none, like typing.Optional

sgkit/stats/aggregation.py
```diff
@@ -167,6 +200,53 @@ def count_variant_alleles(ds: Dataset, merge: bool = True) -> Dataset:
...
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
+
+    Returns
+    -------
+    Dataset containing variable `call_allele_count` of allele counts with
+    shape (variants, cohorts, alleles) and values corresponding to
+    the number of non-missing occurrences of each allele.
```
eric-czech, line 222: I'm wondering if this shouldn't also return a (variants, cohorts) array that contains the number of present samples used in each aggregation.  It doesn't do it now, but if `count_call_alleles` returned -1 for all alleles when any (or potentially all) input calls were missing then I can see a count of the samples with non-negative allele counts being a useful downstream denominator.

Is that not important with the stats functions in `popgen.py` now?  It may make sense to put a warning on those functions and log an issue related to supporting it.  It seems like only ignoring the missing values during the original allele count must introduce some kind of bias through all those aggregations.

sgkit/stats/popgen.py
```diff
@@ -28,98 +31,168 @@ def diversity(
...
+    ac = ds[allele_counts]
+    an = ac.sum(axis=2)
+
+    n_variants = ds.dims["variants"]
+    n_cohorts = ds.dims["cohorts"]
+    ac = da.asarray(ac)
+    an = da.asarray(an)
+    shape = (ac.chunks[0], n_cohorts, n_cohorts)
+    d = da.map_blocks(_divergence, ac, an, chunks=shape, dtype=np.float64)
+    assert_array_shape(d, n_variants, n_cohorts, n_cohorts)
```
eric-czech, line 125: 👍  -- I'll try to remember to start doing that too after all `map_blocks` calls.

sgkit/stats/popgen.py
```diff
@@ -28,98 +31,168 @@ def diversity(
...
+    d_sum = d.sum(axis=0)
+    assert_array_shape(d_sum, n_cohorts, n_cohorts)
+
+    new_ds = Dataset({"stat_divergence": (("cohorts_0", "cohorts_1"), d_sum)})
+    return conditional_merge_datasets(ds, new_ds, merge)
+
+
+# c = cohorts
+@guvectorize(  # type: ignore
+    ["void(float64[:], float64[:,:])"],
```
eric-czech, line 136: I think float16 and float32 overloads would make sense here.  I haven't tested it, but it looks like it should be possible to get a float32 result out of diversity by feeding it allele counts that are float32 (or float16).  That seems like a good way for users to control floating point precision so I think we should try to propagate those types like numpy does where we can, especially since np.sum does that and this function is similar.

sgkit/tests/test_popgen.py
```diff
@@ -37,39 +40,70 @@ def ts_to_dataset(ts, samples=None):
 def test_diversity(size):
```
eric-czech, line 40: Given that map_blocks is underlying these stats, I think chunking on variants and samples dimensions would be useful parameters for these tests.  I'm imagining something like:

```python
@pytest.mark.parametrize("size", [2, 3, 10, 100])
@pytest.mark.parametrize("chunks", [(-1, -1), (10, 10)])
def test_diversity(size, chunks):
    ts = msprime.simulate(size, length=100, mutation_rate=0.05, random_seed=42)
    ds = ts_to_dataset(ts).chunk(dict(zip(["samples", "variants"], chunks)))
```

eric-czech (commented):
Very nice @tomwhite!  Those will be great examples to follow.

I had a few suggestions/questions so let me know when you'd like me to take a look again.

sgkit/stats/aggregation.py
tomwhite: I'm not sure. It's certainly shorter to say `ds.call_allele_count`, but this form can't be used in assignments:

```python
ds.call_allele_count = ...
```

I've changed the accessors in this PR, but we should probably discuss a bit more about our code standards, and what we recommend for user code.

sgkit/stats/aggregation.py
tomwhite: This shouldn't have been there, so I've removed it.

sgkit/stats/aggregation.py
tomwhite: Good point. I haven't thought about missingness with respect to the meaning of the functions though. Opened #290

sgkit/stats/popgen.py
tomwhite: Added a float32 overload. Numba doesn't have float16 as far as I can see: https://numba.pydata.org/numba-doc/dev/reference/types.html#numbers

sgkit/tests/test_popgen.py
tomwhite: Good call. Adding this test uncovered two bugs: one to do with the wrong `numblocks` index being used, and another where the `guvectorize` function if the input is chunked along the samples dimension. I haven't been able to figure out how fix the latter yet, so I'm going to leave it for the moment. (Popgen doesn't need to chunk along the samples dimension just yet.)

tomwhite: I've added warnings to the documentation about not supporting chunking in the samples dimension.

tomwhite:
Thanks for the review @eric-czech. I have addressed your comments (and left some until later), so it should be ready for another look.

eric-czech (approved):

sgkit/stats/popgen.py
eric-czech: Ah right, ty

jeromekelleher:
Looks like we're ready to merge?

Issue #125, opened 2020-08-19
Improve contributing docs


hammer:
A few things missing:
- [x] information about Mergify #107 
- [x] file an issue before submitting a PR and be sure to mention the issue in the PR comments
- [ ] clarify docstring style https://github.com/pystatgen/sgkit/pull/114#issuecomment-676584075

hammer:
@jerowe @daletovar @timothymillar would love your input on what we need to add/remove/change in our contributing docs.

tomwhite:
> file an issue before submitting a PR and be sure to mention the issue in the PR comments

On https://pystatgen.github.io/sgkit/contributing.html#issues it already says,

> Before starting work, make sure there is an issue covering the feature or bug you plan to produce a pull request for.

But it should be changed to say to mention the issue in the PR comments.

hammer:
@jeromekelleher mentioned in the [last developer call](https://discourse.pystatgen.org/t/20200827-developer-call-notes/91) that he had a [guide](https://stdpopsim.readthedocs.io/en/latest/development.html#github-workflow) from another project that could be useful to link from our documentation.

There's also been some discussion in #129 and #140 on the use of branches in developer forks and how to keep those branches up to date. Do we need to add some text to our contributing docs to reflect the consensus from those issues?

hammer:
Also I believe https://github.com/pystatgen/sgkit/issues/216 addresses the docstring style task above.

hammer:
@tomwhite do we need to augment the contributing docs to ensure we don't miss functions when building the Sphinx docs as in https://github.com/pystatgen/sgkit/issues/212?

aktech:
I see #212 is closed, but I still don't see `stats` and `io` modules docs, is that intentional?

> do we need to augment the contributing docs to ensure we don't miss functions when building the Sphinx docs as in #212?

The instructions for the same does seems to be fixing here: https://pystatgen.github.io/sgkit/contributing.html#docstrings

> Also I believe #216 addresses the docstring style task above.

I have made a comment here on my understanding of the same: https://github.com/pystatgen/sgkit/issues/216#issuecomment-688964428

Let me know, if these makes sense, I will open a PR for the same.

aktech:
@tomwhite @hammer Can this be closed now? as #261 is merged

PR #261, opened 2020-09-16
Standardize docstrings and improve contributing docs

### LINKED ISSUE #216: Standardize array type annotations in docstrings
Here are a few conventions floating around the repo with minor differences in how arrays are documented:

From https://github.com/pystatgen/sgkit/blob/master/sgkit/stats/association.py#L26:

```
Parameters
----------
XL : (M, N) array-like
    "Loop" covariates for which N separate regressions will be run
XC : (M, P) array-like
    "Core" covariates included in the regressions for each loop
    covariate. All P core covariates are used in each of the N
    loop covariate regressions.
Y : (M, O)
    Continuous outcomes
```

From https://github.com/pystatgen/sgkit/blob/master/sgkit/stats/regenie.py#L610:

```
    Parameters
    ----------
    G : (M, N) ArrayLike
        Genotype data array, `M` samples by `N` variants.
    X : (M, C) ArrayLike
        Covariate array, `M` samples by `C` covariates.
    Y : (M, O) ArrayLike
        Outcome array, `M` samples by `O` outcomes.
    contigs : (N,) ArrayLike
```

From https://github.com/pystatgen/sgkit/blob/1a5885734020cc52847a3fdd327a35f4426aeb73/sgkit/api.py:

```
Parameters
----------
variant_contig_names : list of str
    The contig names.
variant_contig : array_like, int
    The (index of the) contig for each variant.
variant_position : array_like, int
    The reference position of the variant.
```

There are some situations where a data type for the array and shape information is relevant and other times that it's not.  I propose we standardize ```array_name : [shape_info] array_like[, dtype]```.  As an example, ```x : (M, N) array_like, int```.

When the "array_like" type needs to be more specific, e.g. a Dask array, I think we can still use "array_like" in the docstring and let a more specific type hint indicate the true bounds on inputs.

For reference, here is a random Dask [docstring snippet](https://github.com/dask/dask/blob/bdb7e906051b686207d8e1005f13962593be9797/dask/array/linalg.py#L954):

```
Parameters
----------
a : (M, M) array_like
    A triangular matrix
b : (M,) or (M, N) array_like
```

aktech:
Fixes #216 #125 

- [x] Standardize docstrings
- [x] Uses `scanpydoc` and `sphinx_autodoc_typehints`
- [x] Removes type information from all docstrings (except array dimensions)
- [x] Fix formatting for list view in Return
- [x] Adds custom extension from scanpy too parse return docs to fix this issue: https://github.com/theislab/scanpydoc/issues/7
- [x] Adds numpy, python and sklearn for intersphinx linking
- [x] Fix issue with xarray objects not linked (workaround)

@tomwhite @eric-czech @hammer


sgkit/model.py
```diff
@@ -15,43 +16,46 @@
...
-    sample_id: Any,
-    call_genotype: Any,
-    call_genotype_phased: Any = None,
-    variant_id: Any = None,
+    variant_contig: ArrayLike,
+    variant_position: ArrayLike,
+    variant_alleles: ArrayLike,
+    sample_id: ArrayLike,
+    call_genotype: ArrayLike,
+    call_genotype_phased: Optional[ArrayLike] = None,
```
tomwhite, line 24: I noticed that this is shown in the rendered doc as `Union[ndarray, Array, None]`. Not necessarily a problem, but it means we probably should keep the "optional" marker in the description (as you have done).

sgkit/model.py
```diff
@@ -15,43 +16,46 @@
...
     *,
     variant_contig_names: List[str],
-    variant_contig: Any,
-    variant_position: Any,
-    variant_alleles: Any,
-    sample_id: Any,
-    call_genotype: Any,
-    call_genotype_phased: Any = None,
-    variant_id: Any = None,
+    variant_contig: ArrayLike,
```
tomwhite, line 19: I don't think this is under our control, but the way it displays in the HTML is as `ndarray, Array`, then when you mouse over it changes to `Union[ndarray, Array]`.

sgkit/model.py
```diff
@@ -15,43 +16,46 @@
...
-    call_genotype : array_like, int
-        Genotype, encoded as allele values (0 for the reference, 1 for
-        the first allele, 2 for the second allele), or -1 to indicate a
-        missing value.
-    call_genotype_phased : array_like, bool, optional
-        A flag for each call indicating if it is phased or not. If
-        omitted all calls are unphased.
-    variant_id: array_like, str or object, optional
+    call_genotype
+        array_like (element type: int) Genotype, encoded as allele values
```
tomwhite, line 46: I think this might be a bit more readable as 

[array_like, element type: int] Genotype, encoded as allele values

but would be interested to hear other opinions.

tomwhite (commented):
This looks great @aktech! Thanks for sorting out the intersphinx linking too - that's working very well now.

I've made a couple of minor stylistic suggestions.

sgkit/model.py
aktech: Yeah, I was thinking its even better that it shows explicitly what array like mean.

sgkit/model.py
aktech: Yes, indeed. `sphinx_autodoc_typehints` does that to make it more readable, as `ndarray, Array` is easier on eyes than `Union[ndarray, Array]`

aktech: and moreover, just displaying `ArrayLike` doesn't tells the user explicitly what it means.

jeromekelleher:
I think the consensus on today's call was that this is a good step in the right direction, if not the final answer for our docstringing needs. 

Are there things needed before we can merge?

aktech:
> Are there things needed before we can merge?

Yes, there is this comment that needs to be addressed: https://github.com/pystatgen/sgkit/pull/261#discussion_r489513526

sgkit/model.py
```diff
@@ -31,26 +31,26 @@ def create_genotype_call_dataset(
...
+        [array_like, element type: str or object]
         The unique identifier of the sample.
     call_genotype
-        array_like (element type: int) Genotype, encoded as allele values
+        [array_like, element type: int] Genotype, encoded as allele values
         (0 for the reference, 1 for the first allele, 2 for the second allele),
         or -1 to indicate a missing value.
     call_genotype_phased
-        array_like (element type: bool), optional A flag for each call indicating if it is
+        [array_like, element type: bool], optional A flag for each call indicating if it is
```
tomwhite, line 50: One last change then I think we're there: 

[array_like, element type: bool, optional]

aktech: Done, thanks!

tomwhite (approved):
Looks great - thanks @aktech!

Issue #23, opened 2020-07-08
Simulate genotypes for unit tests


eric-czech:
I'm trying to simulate some genotype calls for unit tests using a simple model from `msprime`.  At @jeromekelleher's suggestion, I tried this:

```python
# Using 1.x code, not yet released
!pip install git+https://github.com/tskit-dev/msprime.git#egg=msprime

import msprime
model = msprime.Demography.island_model(2, migration_rate=1e-3, Ne=10**4)
samples = model.sample(10, 10) # 10 haploids from each pop
ts = msprime.simulate(
     samples=samples,
     demography=model, # what should this be?
     mutation_rate=1e-8,
     recombination_rate=1e-8
)
list(ts.variants())
[]
```

What is `demography` supposed to be?  I saw a "FIXME" for documenting it in `simulate` but I wasn't sure what to set it to if not the `Demography` instance that comes back from `island_model`.

Also, is this why I'm not getting any variants back out?

I am ultimately trying to simulate hardy weinberg equilibrium.  I'd like to have some fraction of the variants be in or near perfect equilibrium and the remainder way out of it.  If you get a chance, could I get an assist @jeromekelleher?

ravwojdyla:
+1 @eric-czech when we figure out the right toolset for (test) data generation it could be valuable to expose them as strategies in `hypothesis` (this could be a separate test module within this repo). wdyt?

jeromekelleher:
I'm impressed you've gotten this far with the new undocumented APIs @eric-czech! You're just missing a "length" parameter to tell it how much sequence to simulate:
```python
model = msprime.Demography.island_model(2, migration_rate=1e-3, Ne=10**4)
samples = model.sample(10, 10) # 10 haploids from each pop
ts = msprime.simulate(
    samples=samples,
    demography=model,
    length=10**6,
    recombination_rate=1e-8
)
ts = msprime.mutate(ts, model=msprime.JukesCantor(), rate=1e-8, discrete=True)
print(ts.num_sites)
for variant in ts.variants():
    print(int(variant.site.position), variant.alleles, variant.genotypes, sep="\t")
```
This gives ~3000 sites, that look like
```
641872  ('G', 'A')      [0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 1 0 0 0]
641885  ('C', 'G')      [0 1 0 0 1 0 0 0 0 1 0 1 0 0 1 0 1 0 0 1]
[...]

PR #262, opened 2020-09-16
PCA implementation

eric-czech:
Implementation for https://github.com/pystatgen/sgkit/issues/95.

Notes:

### Upstream Fixes

Some upstream fixes necessary for this were:

- Dask SVD broken for short-fat arrays (https://github.com/dask/dask/pull/6591)
- Dask SVD had poor performance on certain in-memory arrays (https://github.com/dask/dask/pull/6616)
- The svd_flip function in dask-ml (used by PCA/TruncatedSVD) would have forced in-memory evaluation, but it doesn't anymore (https://github.com/dask/dask/pull/6613)
  - The fix for this also makes da.linalg.svd results consistent across platforms and input array chunkings 
- Dask-ml PCA classes force eager evaluation (https://github.com/dask/dask-ml/issues/740)
- Randomized SVD broken for small arrays (https://github.com/dask/dask/pull/6622)

### Discussion

- I added https://github.com/pystatgen/sgkit/issues/285 to track adding more extensive examples
- This contains a temporary workaround for https://github.com/pystatgen/sgkit/issues/282 that should eventually be removed
- Dependencies added:
  - matplotlib and scikit-allel as dev dependencies
  - dask-ml as a core dependency
- I opened https://github.com/pystatgen/sgkit/issues/286 to discuss how to handle required variables in a calculation like this (I don't particularly like how it works in this PR at the moment)

### TODO

- [x] More tests
  - [x] f4/f8 dtype preservation
  - [x] More complex chunking tests (this is the source of many problems)
  - [x] Add randomized/deterministic comparison 
  - [x] Ensure lazy evaluation
  - [x] Validate sign stability across random states and chunkings
  - [x] Add missing data warning
  - [x] Test for inputs on different array backends
  - [x] Patterson scaler
    - [x] test vs skallel
    - [x] test with missing data
    - [x] test dtype preservation
- [x] Add return_estimator option (or save scaler properties in dataset)
- [X] Throw informative errors when chunking not compatible with algorithm
- [x] Decide whether or not randomized pca should be a separate function since it is the only option for fully-chunked arrays
- [x] Lift `count_call_alternate_alleles` into issue
- [x] Improve test speed
- [x] Add docs

### DIFF
--- a/sgkit/stats/pca.py
+++ b/sgkit/stats/pca.py
@@ -1,0 +1,343 @@
+from typing import Any, Optional, Union
+
+import dask.array as da
+import numpy as np
+import xarray as xr
+from dask_ml.decomposition import TruncatedSVD
+from sklearn.base import BaseEstimator
+from sklearn.pipeline import Pipeline
+from typing_extensions import Literal
+from xarray import DataArray, Dataset
+
+from sgkit import variables
+
+from ..typing import ArrayLike, DType, RandomStateType
+from ..utils import conditional_merge_datasets
+from .aggregation import count_call_alleles
+from .preprocessing import PattersonScaler
+
+
+def pca_est(
+    ds: Dataset,
+    n_components: int = 10,
+    *,
+    ploidy: Optional[int] = None,
+    scaler: Optional[Union[BaseEstimator, str]] = None,
+    algorithm: Optional[Literal["tsqr", "randomized"]] = None,
+    n_iter: int = 0,
+    random_state: RandomStateType = 0,
+    variable: str = "call_alternate_allele_count",
+) -> BaseEstimator:
+    """ Create PCA estimator """
+    if ploidy is None:
+        if "ploidy" not in ds.dims:
+            raise ValueError(
+                "`ploidy` must be specified explicitly when not present in dataset dimensions"
+            )
+        ploidy = ds.dims["ploidy"]
+    scaler = scaler or "patterson"
+    if isinstance(scaler, str):
+        if scaler != "patterson":
+            raise ValueError(
+                f"Only 'patterson' scaler currently supported (not '{scaler}')"
+            )
+    algorithm = algorithm or "tsqr"
+    if algorithm not in {"tsqr", "randomized"}:
+        raise ValueError(
+            f"`algorithm` must be one of ['tsqr', 'randomized'] (not '{algorithm}')"
+        )
+
+    numblocks = da.asarray(_allele_counts(ds, variable, check_missing=False)).numblocks
+    if all(s > 1 for s in numblocks) and algorithm != "randomized":
+        raise ValueError(
+            "PCA can only be performed on arrays chunked in 2 dimensions if algorithm='randomized'. "
+            "Consider using this algorithm instead or rechunking the alternate allele counts array "
+            "(e.g. ds.call_alternate_allele_count.chunk((None, -1)))."
+        )
+
+    return Pipeline(
+        [
+            ("scaler", PattersonScaler(ploidy=ploidy)),
+            (
+                "pca",
+                TruncatedSVD(
+                    n_components=n_components,
+                    algorithm=algorithm,
+                    n_iter=n_iter,
+                    random_state=random_state,
+                    compute=False,
+                ),
+            ),
+        ]
+    )
+
+
+def pca_fit(
+    ds: Dataset,
+    est: BaseEstimator,
+    *,
+    variable: str = "call_alternate_allele_count",
+    check_missing: bool = True,
+) -> BaseEstimator:
+    """ Fit PCA estimator """
+    AC = _allele_counts(ds, variable, check_missing=check_missing)
+    return est.fit(da.asarray(AC).T)
+
+
+def pca_transform(
+    ds: Dataset,
+    est: BaseEstimator,
+    *,
+    variable: str = "call_alternate_allele_count",
+    check_missing: bool = True,
+    merge: bool = True,
+) -> Dataset:
+    """ Apply PCA estimator to new data """
+    AC = _allele_counts(ds, variable, check_missing=check_missing)
+    projection = est.transform(da.asarray(AC).T)
+    new_ds = Dataset(
+        {variables.sample_pca_projection: (("samples", "components"), projection)}
+    )
+    return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+def _get(est: BaseEstimator, attr: str, fn: Any = lambda v: v) -> Optional[ArrayLike]:
+    try:
+        if hasattr(est, "named_steps"):
+            est = est["pca"]
+        return fn(getattr(est, attr))
+    except Exception:
+        return None
+
+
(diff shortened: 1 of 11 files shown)


eric-czech:
fyi @jeromekelleher I didn't use msprime for any of this because the features you mentioned in https://github.com/pystatgen/sgkit/issues/23#issue-653340614 still aren't released afaik.  Looks like it's been quite a while since the last one (Dec 2019).  Is that likely to change anytime soon?

sgkit/stats/pca.py
```diff
@@ -0,0 +1,338 @@
...
+from .aggregation import count_call_alleles
+from .preprocessing import PattersonScaler
+
+
+def pca_est(
+    ds: Dataset,
+    n_components: int = 10,
+    *,
+    ploidy: Optional[int] = None,
+    scaler: Union[BaseEstimator, str] = "patterson",
```
jeromekelleher, line 23: I tend to prefer having "None" as the default in the signature as it gives a bit more flexibility, both in the context of other parameters and over time as the API evolves. Are you sure that "patterson" will always be the right default value, in every context?

Same for other value here.

jeromekelleher (commented):
LGTM @eric-czech, just a minor point on my aversion to hard-coding default values in the method signature. I know \approx nothing about PCA though, so pinging @alimanfoo for a review.

jeromekelleher:
> fyi @jeromekelleher I didn't use msprime for anything here because the features mentioned in #23 (comment) still aren't released afaik. Looks like it's been quite a while since the last one (Dec 2019). Is that likely to change anytime soon?

There'll be a beta in the coming weeks, but hard to know when we'll ship a final release. I think you made the right choice, we can follow up with more detailed tests later.

sgkit/stats/pca.py
eric-czech: Hm I can't think of a scenario where the default scaler becomes a choice made based on other inputs, but making it more flexible in case one comes up sounds good to me.  It definitely makes sense for the `algorithm` parameter.  I changed both in https://github.com/pystatgen/sgkit/pull/262/commits/28a9b495ffb38f92c526d0c91f62673f76d188bd.

tomwhite (approved):
This looks great @eric-czech. No real substantive comments from me. It's good to see lots of very comprehensive tests here, and also that most of the PCA implementation details are upstream.

sgkit/stats/pca.py
```diff
@@ -0,0 +1,342 @@
...
+        Ratio of variance explained to total variance for each principal component,
+        i.e. ``dask_ml.decomposition.TruncatedSVD.explained_variance_ratio_``.
+
+    Examples
+    --------
+
+    >>> import xarray as xr
+    >>> import numpy as np
+    >>> import sgkit as sg
+    >>> from sgkit.testing import simulate_genotype_call_dataset
```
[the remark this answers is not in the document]
eric-czech: https://github.com/pystatgen/sgkit/pull/262/commits/237bc966beddb34cb5f360b520d75bf9fce72b46

eric-czech:
@jeromekelleher should I wait for @alimanfoo to review or merge?  Should have asked on the call.

jeromekelleher:
> @jeromekelleher should I wait for @alimanfoo to review or merge? Should have asked on the call.

No, if we haven't heard from him in a few days don't block (same goes for me, over the next few weeks). Merge away here I'd say.

sgkit/stats/pca.py
```diff
@@ -0,0 +1,341 @@
...
+    }
+    new_ds = Dataset({k: v for k, v in new_ds.items() if v[1] is not None})
+    if "sample_pca_component" in new_ds and "sample_pca_explained_variance" in new_ds:
+        new_ds["sample_pca_loading"] = new_ds["sample_pca_component"] * np.sqrt(
+            new_ds["sample_pca_explained_variance"]
+        )
+    return conditional_merge_datasets(ds, new_ds, merge)
+
+
+def pca(
```
ravwojdyla, line 135: Should this function use `sgkit.variables`, which would mean:
 * using variable references instead of strings
 * validate input
 * validate output
 * update the docs

sgkit/tests/test_preprocessing.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.mark.parametrize("dtype", ["i1", "i2", "f2", "f4", "f8"])
+@pytest.mark.parametrize("use_nan", [True, False])
+def test_patterson_scaler__missing_data(dtype, use_nan):
+    ac = np.array([[0, -1, -1, -1], [1, 0, -1, -1], [2, 2, 1, -1]], dtype=dtype)
+    if use_nan and ac.dtype.kind != "f":
+        return
+    if use_nan:
+        # Test that nan and negative sentinel values
```
ravwojdyla, line 82: there is something wrong with this comment (?)

sgkit/stats/pca.py
```diff
@@ -0,0 +1,341 @@
...
+    try:
+        if hasattr(est, "named_steps"):
+            est = est["pca"]
+        return fn(getattr(est, attr))
+    except Exception:
+        return None
+
+
+def pca_stats(ds: Dataset, est: BaseEstimator, *, merge: bool = True) -> Dataset:
+    """ Extract attributes from PCA estimator """
```
ravwojdyla, line 112: nit: when you have a single line docstring, you add a space prefix/suffix, why? Is this documented somewhere?

ravwojdyla (changes requested):

sgkit/stats/pca.py
eric-czech: Added the variables (https://github.com/pystatgen/sgkit/pull/262/commits/05fa77df1a6c8f1fe110631dfd8672e526ad170f#diff-cd77642bae81c13ab776800efe4ba498900a284bc2e4741d943b727f60b8b7e1R250-R284) and updated references (https://github.com/pystatgen/sgkit/pull/262/commits/05fa77df1a6c8f1fe110631dfd8672e526ad170f#diff-e66844198337edbcfa8f27fbcbcc58ef011272669fe9295119b0847c1b7cdf69R134).   Squashed one too many commits before rebasing on the current master so I lost a clean delta, but that was basically all I changed.

sgkit/stats/pca.py
```diff
@@ -0,0 +1,341 @@
...
+    variable: str = "call_alternate_allele_count",
+) -> BaseEstimator:
+    """ Create PCA estimator """
+    if ploidy is None:
+        if "ploidy" not in ds.dims:
+            raise ValueError(
+                "`ploidy` must be specified explicitly when not present in dataset dimensions"
+            )
+        ploidy = ds.dims["ploidy"]
+    if scaler is None:
```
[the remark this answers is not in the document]
eric-czech: changed

sgkit/tests/test_preprocessing.py
eric-czech: good call, fixed

sgkit/stats/pca.py
eric-czech: https://github.com/pystatgen/sgkit/issues/325

Issue #247, opened 2020-09-14
Add windows CI


jeromekelleher:
I think it would be a good idea to have windows CI setup before we implement the mono-repo (#65). 

Any takers?

tomwhite:
We now use msprime for testing, so we'll need to install that on Windows. Do you think it's best to use a conda env to do that @jeromekelleher?

jeromekelleher:
Hmm, yes, good point. Conda is really the only option on Windows for msprime  (and life generally easier on windows with conda). I'm guessing a lot of the stuff we want to use will be conda only on windows anyway.

tomwhite:
I'm going to try using this GitHub Action to set up conda: https://github.com/marketplace/actions/setup-miniconda.

@ravwojdyla pointed out two other alternatives (https://github.com/s-weigand/setup-conda and https://github.com/fastai/workflows/tree/master/conda-env), but they don't seem to be as popular.

PR #269, opened 2020-09-17
Windows CI

tomwhite:
This is a first attempt at #247.

* The main change is to use Miniconda (so we can install msprime easily on Windows).
* I couldn't get the pre-commit GH action to work with conda, so I ended up just calling the relevant commands directly.
* Added guards to only perform some actions (linting, doc checking, etc) for one matrix entry (Ubuntu/Python 3.7) 
* I had to skip some doctest checking, since on Windows it would show numpy arrays as `(array([0, 2], dtype=int64)` not `(array([0, 2])`. I'm not sure why this is happening exactly, but I couldn't see a way in numpy to control whether the dtype was displayed or not.
* We probably want to update `docs.yml` to use the same approach for consistency.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,32 +5,44 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-latest]
         python-version: [3.7, 3.8]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+    - name: Set up Miniconda with Python version ${{ matrix.python-version }}
+      uses: goanpeca/setup-miniconda@v1
       with:
+        auto-update-conda: true
+        channels: conda-forge
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
+      shell: bash -l {0}
       run: |
-        sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
-        python -m pip install --upgrade pip
+        conda install msprime
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
-      uses: pre-commit/action@v2.0.0
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.7'
+      shell: bash -l {0}
+      run: |
+        pre-commit install
+        pre-commit run --all-files
     - name: Check for Sphinx doc warnings
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.7'
+      shell: bash -l {0}
       run: |
         cd docs
         make html SPHINXOPTS="-W --keep-going"
     - name: Test with pytest and coverage
+      shell: bash -l {0}
       run: |
         pytest -v --cov=sgkit --cov-report=term-missing
     - name: Upload coverage to Codecov
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.8'
+      shell: bash -l {0}
       env:
         CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,44 +5,32 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
         python-version: [3.7, 3.8]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Miniconda with Python version ${{ matrix.python-version }}
-      uses: goanpeca/setup-miniconda@v1
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
       with:
-        auto-update-conda: true
-        channels: conda-forge
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
-      shell: bash -l {0}
       run: |
-        conda install msprime
+        sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
+        python -m pip install --upgrade pip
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -5,32 +5,44 @@ on: [push, pull_request]
...
+        auto-update-conda: true
+        channels: conda-forge
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
+      shell: bash -l {0}
       run: |
-        sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
-        python -m pip install --upgrade pip
+        conda install msprime
         pip install -r requirements.txt -r requirements-dev.txt
```
ravwojdyla, line 26: Should we install these also via conda?

.github/workflows/build.yml
```diff
@@ -5,32 +5,44 @@ on: [push, pull_request]
...
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
+    - name: Set up Miniconda with Python version ${{ matrix.python-version }}
+      uses: goanpeca/setup-miniconda@v1
       with:
+        auto-update-conda: true
+        channels: conda-forge
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
+      shell: bash -l {0}
```
ravwojdyla, line 23: Why is this necessary?

.github/workflows/build.yml
```diff
@@ -5,32 +5,44 @@ on: [push, pull_request]
...
-        sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
-        python -m pip install --upgrade pip
+        conda install msprime
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
-      uses: pre-commit/action@v2.0.0
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.7'
+      shell: bash -l {0}
+      run: |
+        pre-commit install
```
ravwojdyla, line 31: So this is not using cache for pre commit anymore? ` FYI pre-commit/action` comes with cache built in

.github/workflows/build.yml
```diff
@@ -5,32 +5,44 @@ on: [push, pull_request]
...
     - name: Install dependencies
+      shell: bash -l {0}
       run: |
-        sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
-        python -m pip install --upgrade pip
+        conda install msprime
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
-      uses: pre-commit/action@v2.0.0
+      if: matrix.os == 'ubuntu-latest' && matrix.python-version == '3.7'
```
ravwojdyla, line 28: Is there no better way to do this than having it repeated in each step?

ravwojdyla (changes requested):

.github/workflows/build.yml
tomwhite: It's to ensure the conda env is activated. See https://github.com/marketplace/actions/setup-miniconda#important

I can add a comment to document this.

.github/workflows/build.yml
tomwhite: Good question!

.github/workflows/build.yml
tomwhite: Here's an example of a run where it failed: https://github.com/tomwhite/sgkit/runs/1121992680?check_suite_focus=true

```
##[error]Unable to locate executable file: pre-commit. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.
```

.github/workflows/build.yml
tomwhite: I'm not sure. What's the downside of doing it like this?

ravwojdyla: @tomwhite not sure, but if we already use conda, why not use it all the way?

.github/workflows/build.yml
ravwojdyla: Oh, this is why internally we use:

```
    - uses: s-weigand/setup-conda@v1
      with:
        python-version: 3.8
        conda-channels: conda-forge
```

that makes it easier to use the `base` env.

.github/workflows/build.yml
jeromekelleher: I'd be in favour of using conda by default - it's good to test out CI on the conda infrastructure.

jeromekelleher:
My take would be that it'd be better to write a specific workflow for windows/conda and leave the current Ubuntu workflows as they are. In practise, it never works having complicated stuff work across all three platforms. Having a specific workflow where we just do the bare minimum to make the unit tests work on Windows will make things much easier.

tomwhite:
Thanks for all the feedback. Hows does this look now?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,30 @@
+name: Windows build
```
ravwojdyla, line 1: Do you want to add this as a badge? I assume so, and if we do, this name will appear on the badge, is this the name we want (maybe just `Windows`)?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on: [push, pull_request]
+
+jobs:
+  build:
```
ravwojdyla, line 6: Do we need to update `mergify` config (we might want to rename `build` to say `win_build` to avoid collision)?

ravwojdyla (changes requested):

.github/workflows/windows.yml
jeromekelleher: Be nice to have the badge.

tomwhite: Fixed in https://github.com/pystatgen/sgkit/pull/269/commits/5314a835299b1612a0928e0c33eeb28055afdcfb

.github/workflows/windows.yml
tomwhite: Fixed in https://github.com/pystatgen/sgkit/pull/269/commits/5314a835299b1612a0928e0c33eeb28055afdcfb

ravwojdyla (approved):

jeromekelleher:
LGTM - ready to merge?

jeromekelleher (approved):

PR #272, opened 2020-09-21
Split out docs requirements so they don't need to be installed on Windows

### LINKED ISSUE #247: Add windows CI

tomwhite:
The Windows CI build is [failing](https://github.com/pystatgen/sgkit/runs/1143067740) because it crossed with #261 which introduced dependencies that are not available in conda (for Windows). These are `scanpydoc` and `sphinx_autodoc_typehints`.

Since we don't build docs on Windows, I've split out the docs requirements, and not included them for the Windows build.

Fixes #247


jeromekelleher (approved):
Good solution

PR #271, opened 2020-09-18
Improve/fix the validation a bit

ravwojdyla:
* the order in the dockerfile matters a bit (otherwise `data.table` fails to properly install)
 * fail `run.sh` if any of the cmds fail along the way

Issue #257, opened 2020-09-15
Move sgkit-plink to main sgkit repo


tomwhite:

ravwojdyla:
FYI: will do this as POC for the rest of the repos.

tomwhite:
Thanks @ravwojdyla. We also agreed in today's call that this would include a POC of the "barrier" to give the user control over which IO libraries to install. Plink has a straightforward dependency, but VCF does not so it will be needed there at least.

Issue #65, opened 2020-07-24
Monorepo vs. multi-repo


ravwojdyla:
We have a  solid style check setup in the main repo, IO repos have a bit older setup, it might be hard to keep them in sync plus we also have repos in RS that also have similar setups. I wonder if we should explore some kind of centralisation of this style check setup (akin to common plugin in sbt or root parent in maven). The best I could find for precommit is [this](https://github.com/pre-commit/pre-commit/issues/731#issuecomment-376945745), we could give it a try, one downside I can see is that it might not work well with mypy check (but that already doesn't work well anyway even in solo repo, related: https://github.com/pystatgen/sgkit/issues/39)

An alternative solution to this problem would be to have a single repo with proper separation of modules to prevent dependency hell.

ravwojdyla:
It came up during our weekly sync, so I wanted to document it, we could use `extras_require` to separate dependencies, and even provide a convenient way to install a complete sgkit ex: `sgkit[complete]`.

ravwojdyla:
@jeromekelleher @alimanfoo @tomwhite @eric-czech @hammer any thoughts/objects/-1/+1 to merging IO repos into the main repo and using `extra_requires` to separate dependencies. If you prefer to discuss this during our ~weekly meetings, this can obviously wait until the next next one.

eric-czech:
I'm a +1 to that as long as it's always easy to install the io-related packages in a separate environment from all the other sgkit packages (and not need to share common dependencies).  Something like this seems reasonable to me:

```
conda activate base
conda install sgkit[core]
conda activate io
conda install sgkit[plink,bgen]
```

I think making CI run for everything in a common environment will work for now, but I don't think that would be a good long term solution.  How hard is it to split the build environments like that?

ravwojdyla:
@eric-czech I believe (we should double check), install of any `extra_requires` will always install the root package(s). I'm not sure right now how idiomatic is it to not install root package for any of the extra, and I'm not sure if that is a good idea at the moment, BUT if we do want that, I guess a workaround for this could be to have a dummy root package `sgkit_parent`, and then everything else is in extra packages like `core`, `plink`, `bgen`. Regarding splitting CI, I agree for now we should just keep a single env, but later we can split things up (it shouldn't be too difficult).

ravwojdyla:
And to provide some further reasoning for this:
 * it would lower the overhead of our style check, doc setup etc
[...]

PR #235, opened 2020-09-07
Add VCF docs

tomwhite:
First (draft) attempt to include VCF docs for https://github.com/pystatgen/sgkit-vcf.

The API docs are included, but I needed to install the sgkit-vcf package ~~(with `--no-deps`, to avoid having to install cyvcf2, which is not needed for API signatures)~~.

sgkit-vcf

Refs #54

PR #277, opened 2020-09-23
sgkit-plink IO merger

ravwojdyla:
Re: #65 , #257 

~Depends on: #274, #271~

UX test:

```bash
# ------------------------------------------------------------------
# create a new clean py env
# ------------------------------------------------------------------
(base) ➜  projects conda create -n test_sgkit_blah python=3.7
(base) ➜  projects conda activate test_sgkit_blah

# ------------------------------------------------------------------
# clean install of sgkit (need to install from source since we don't
# publish yet) towards the end of the output you can see packages
# installed, they do NOT include sgkit-plink dependencies
# ------------------------------------------------------------------
(test_sgkit_blah) ➜  ~ pip install projects/sgkit
Processing /Users/rav/projects/sgkit
Collecting numpy
...
Installing collected packages: numpy, six, python-dateutil, pytz, pandas, xarray, pyyaml, toolz, dask, scipy, asciitree, numcodecs, monotonic, fasteners, zarr, llvmlite, numba, typing-extensions, sgkit
Successfully installed asciitree-0.3.3 dask-2.27.0 fasteners-0.15 llvmlite-0.34.0 monotonic-1.5 numba-0.51.2 numcodecs-0.7.2 numpy-1.19.2 pandas-1.1.2 python-dateutil-2.8.1 pytz-2020.1 pyyaml-5.3.1 scipy-1.5.2 sgkit-0.1.dev172+g725c774 six-1.15.0 toolz-0.10.0 typing-extensions-3.7.4.3 xarray-0.16.1 zarr-2.4.0

# ------------------------------------------------------------------
# let's test if sgkit works as expected
# ------------------------------------------------------------------
(test_sgkit_blah) ➜  ~ echo $PYTHONPATH

(test_sgkit_blah) ➜  ~ python3.7
>>> from sgkit.testing import simulate_genotype_call_dataset
>>> simulate_genotype_call_dataset(10, 10)
<xarray.Dataset>
Dimensions:             (alleles: 2, ploidy: 2, samples: 10, variants: 10)
...

# ------------------------------------------------------------------
# looks like the sgkit core works fine, now let's try to use
# sgkit-plink IO:
# ------------------------------------------------------------------
>>> from sgkit.io.plink import read_plink
...
Please install them via pip :

  pip install --upgrade 'sgkit[plink]'

# ------------------------------------------------------------------
# we've got an error there that tells us we need to install missing
# dependencies, and it shows us the commend, again since we don't
# publish sgkit yet, I install from source:
# ------------------------------------------------------------------

(test_sgkit_blah) ➜  ~ pip install --upgrade 'projects/sgkit[plink]'
...
Successfully installed appdirs-1.4.4 bed-reader-0.1.1 chardet-3.0.4 fsspec-0.8.2 idna-2.10 locket-0.2.0 packaging-20.4 partd-1.1.0 pooch-1.2.0 pyparsing-2.4.7 requests-2.24.0 sgkit-0.1.dev172+gff4fc2c urllib3-1.25.10

# ------------------------------------------------------------------
# You can see some extra package got installed, let's try to use plink IO again:
# ------------------------------------------------------------------
(test_sgkit_blah) ➜  ~ python3.7
>>> from sgkit.io.plink import read_plink
>>> read_plink(path="/tmp/plink_sim_10s_100v_10pmiss")
# ------------------------------------------------------------------
# all works now
# ------------------------------------------------------------------
```


setup.cfg
```diff
@@ -36,7 +36,24 @@ install_requires =
...
+# --use-feature 2020-resolver, or installing
+# dask[dataframe] in the install_requires, or just listing
+# the 2 missing dependencies from dataframe, the way we do
+# here, when pip finally gets a resolver, this won't be
+# a problem. Here we opt for listing the 2 dependencies
+# since this is the least user invasive solution.
+plink =
+    partd
+    fsspec
+    bed-reader
```
tomwhite, line 55: I can't see where dask[dataframe] actually gets installed - it looks like only its dependencies get installed here?

tomwhite (commented):
This looks great @ravwojdyla.

We should rename `pysnptools.py` since we no longer use that library. Maybe call it `plink_reader.py`?

I was wondering if it would be possible to have the `read_plink` function at the top-level, but that's not possible due to the guard (right?). I think `from sgkit.io.plink import read_plink` is logical however.

Also, we'll need some documentation about how to install the library, but did you think that would be a separate PR?

setup.cfg
ravwojdyla: @tomwhite see the comment above (https://github.com/pystatgen/sgkit/pull/277/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R41-R51) for the context. Does that make sense?

ravwojdyla:
@tomwhite 

> We should rename pysnptools.py since we no longer use that library. Maybe call it plink_reader.py?

Sure - will push a fixup.

> I was wondering if it would be possible to have the read_plink function at the top-level, but that's not possible due to the guard (right?). I think from sgkit.io.plink import read_plink is logical however.

Exactly.

> Also, we'll need some documentation about how to install the library, but did you think that would be a separate PR?

I think Eric added a bit of that in #278 (plus this PR adds informative error message if plink deps are missing), would you suggest anything more?

setup.cfg
tomwhite: It makes sense, but I'm still missing something: where is the `dask[dataframe]` dependency declared? It's used by `pysnptools.py`, but how does it get installed?

tomwhite:
> I think Eric added a bit of that in #278 (plus this PR adds informative error message if plink deps are missing), would you suggest anything more?

Yes, I think we should add a bit saying what you need to do to read PLINK/BGEN/VCF. I've added something in #235, but that will need changing, so I'm happy to do it after this (and the BGEN, VCF equivalents) have gone in.

setup.cfg
ravwojdyla: `dask[dataframe]` is not an "actual" dependency, it's an extra in `dask`, and since we already install `dask` via `dask[array]` in `install_requires` and since it's an **extra in our extras** (which [this bug](https://github.com/pypa/pip/issues/4957) in pip affects) we can't really list `dask[dataframe]` as a dependency (unless we use the 2020 pip resolver), so we have the options outlined in the comment above, so instead of forcing the extra flag on users, I opts for listing the missing dependencies that come from `dask[dataframe]`. Initial version of this PR actually used the 2020 resolver, but I reverted to listing `dask[dataframe]` deps directly so that we can use plain `pip`.

tomwhite: Ah, got it. I was missing the bit about `dask[dataframe]` not actually providing any Dask code.

When the pip bug is fixed do you think we should switch to just add the `dask[dataframe]` dependency here?

ravwojdyla: @tomwhite quick answer - sure. But that bug will be fixed by the 2020 resolver, right now 2020 resolver is available via feature flag, it is scheduled to be default in October this year, so we can wait until then + a couple of months and we should be good to switch over.

tomwhite (approved):
Please also rename `test_pysnptools.py` to `test_plink_reader.py`.

jeromekelleher (approved):
LGTM, the import guard should work well I think.

Just so I'm clear, the intended usage is

```python
# my user script
from sgkit.io.plink import read_plink  # Fails here if we 

df = read_plink("myplinkfile")
```
So, this will fail at import time if bedtools doesn't exist. I guess this will work well once we're good about library hygiene and make sure we don't ever refer to ``sgkit.io.plink`` elsewhere in sgkit.

I guess the other option would be to fail at *run time* when someone calls ``read_plink``. This would mean we don't have to be quite as strict about how we use ``read_plink`` within the library (see #57, eg). I'm sure there are pros and cons either way, just bringing up the idea for discussion.

Anyway, this is easy to tweak afterwards. Very happy to merge this as is.

ravwojdyla:
@tomwhite @jeromekelleher thanks for the reviews, last update:
 * finished up renaming of the files
 * moved plink tests to the main test directory
 * added API doc and installation doc

@jeromekelleher 

> So, this will fail at import time if bedtools doesn't exist. I guess this will work well once we're good about library hygiene and make sure we don't ever refer to sgkit.io.plink elsewhere in sgkit.

Correct, and thanks for raising this it. And as you point out, we can easily change that in the future if we like (though I prefer it the way it is right now). Regarding CLI, there is many ways we can solve that with import/run-time guards, and we can wait until we have CLI to decide.

jeromekelleher:
Sure @ravwojdyla, let's leave things as they are for now and think about the runtime/import time thing later.

Pinging @eric-czech for a review here too, since this is a big change.

tomwhite:
> decided to moved plink tests to the main test directory (no value in having them separate)

I'm looking at the VCF tests, and they are spread across several files, so in that case I think it is worth putting them in a separate directory (`tests/io/vcf`). So perhaps we should for plink too for consistency.

ravwojdyla:
@tomwhite done.

eric-czech (approved):
Thanks @jeromekelleher.

LGTM.  One very minor concern I have is testing that the core functionality works without the extras installed (as they would be in the build env now).  I suppose that only becomes more of a risk when we have more functionality like `vcfzarr_reader` though, or anything that is IO-related without needing the underlying reader libraries.

ravwojdyla:
@eric-czech thanks for the review! Regarding your concern about testing IOs, in those cases we can always leverage pytest markers, and for example by default skip tests that we expect are hard to test locally (and hide them behind a marker flag), kinda like what we do internally with datastore emulated tests.

Btw - about `auto-merge` this PR has a change in workflows, so you need to merge this manually.

eric-czech:
Hm I tried adding auto-merge but no dice.  @jeromekelleher could you take a look?  Was that supposed to be fixed now or did I attach the wrong label?

jeromekelleher:
> Hm I tried adding auto-merge but no dice. @jeromekelleher could you take a look? Was that supposed to be fixed now or did I attach the wrong label?

as @ravwojdyla pointed out, mergify doesn't do auto merge on things that modify workflows. I've hit the manual merge.

PR #283, opened 2020-09-28
Revert "Workaround dask#6631 (#246)"

ravwojdyla:
Dask 2.18.0 has the fix included via https://github.com/dask/dask/pull/6632

This reverts commit ffd85e0e1f27478f550b188a53b597b71338d364.

### DIFF
--- a/sgkit/io/vcfzarr_reader.py
+++ b/sgkit/io/vcfzarr_reader.py
@@ -60,11 +60,9 @@ def read_vcfzarr(path: PathType) -> xr.Dataset:
     )
 
     # Add a mask for variant ID
-    # Note: we wrap the "." inside a list to workaround https://github.com/dask/dask/issues/6631,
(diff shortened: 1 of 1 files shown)


jeromekelleher:
I assume this is just-released: should we set a dask minimum version to avoid hitting this bug?

ravwojdyla:
thanks for the prompt review @jeromekelleher , I'm don't really have a strong opinion here, since we currently don't specify version, it's somewhat implicit that it's the latest version, I'm ok with keeping it that way since we are in heavy dev mode, but if you (or anyone else) feel strongly about specifying the min version, I can definitely add it.

eric-czech (approved):

jeromekelleher (approved):

jeromekelleher:
I think we'll need to pin a minimum dask version at some point, but I don't have a stong opinion on when.

PR #289, opened 2020-10-01
sgkit-vcf merger

### LINKED ISSUE #258: Move sgkit-vcf to main sgkit repo

tomwhite:
This follows #277, with a couple of differences:
* cyvcf2 is *not* included in requirements-dev.txt, since it does not work on Windows. Instead, it is explicitly installed in the Linux CI build.
* All the VCF tests have import guards so they don't run if cyvcf2 isn't installed.

The build changes for developers will need documenting, but I was planning on doing that in #235

Fixes #258


sgkit/tests/io/vcf/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+from sgkit.typing import PathType
+
+
+def path_for_test(shared_datadir: Path, file: str, is_path: bool = True) -> PathType:
+    """Return a test data path whose type is determined by `is_path`.
+
+    If `isPath` is True, return a `Path`, otherwise return a `str`.
+    """
+    path: PathType = shared_datadir / file
```
ravwojdyla, line 11: A more idiomatic way to say this would be:

```
path: PathType = shared_datadir / file
return path if is_path else str(path)
```

sgkit/io/vcf/vcf_reader.py
```diff
@@ -0,0 +1,425 @@
...
+
+def count_variants(path: PathType, region: Optional[str] = None) -> int:
+    """Count the number of variants in a VCF file."""
+    with open_vcf(path) as vcf:
+        if region is not None:
+            vcf = vcf(region)
+        count = 0
+        for variant in region_filter(vcf, region):
+            count = count + 1
+        return count
```
ravwojdyla, line 425: A more idiomatic way to say this would be:

```
return sum(1 for _ in region_filter(vcf, region))
```

sgkit/io/vcf/vcf_reader.py
```diff
@@ -0,0 +1,425 @@
...
+            regions,
+            temp_chunk_length,
+            chunk_width,
+            tempdir_storage_options,
+        )
+
+        ds = zarrs_to_dataset(paths, chunk_length, chunk_width, tempdir_storage_options)
+
+        # Ensure Dask task graph is efficient, see https://github.com/dask/dask/issues/5105
+        with dask.config.set({"optimization.fuse.ave-width": 50}):
```
ravwojdyla, line 194: `50` should probably be exposed in the method parameter

ravwojdyla (changes requested):
Mostly two requests for change related to how we handle `cyvcf2`, plus I glanced at the rest of the code.

.github/workflows/build.yml
```diff
@@ -18,9 +18,11 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
+        sudo apt update -y
+        sudo apt install -y libcurl4-openssl-dev  # Needed for htslib (via cyvcf2)
         sudo apt install libgsl-dev   # Needed for msprime < 1.0. Binary wheels include GSL for >= 1.0
         python -m pip install --upgrade pip
-        pip install -r requirements.txt -r requirements-dev.txt -r requirements-doc.txt
+        pip install -r requirements.txt -r requirements-dev.txt -r requirements-doc.txt cyvcf2
```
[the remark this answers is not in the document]
tomwhite: Great suggestion! Thanks - that's much better.

sgkit/tests/io/vcf/test_tbi.py
```diff
@@ -0,0 +1,43 @@
+import pytest  # noqa: F401 isort:skip
+
+cyvcf2 = pytest.importorskip("cyvcf2")  # noqa: F401
```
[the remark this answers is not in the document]
tomwhite: Ditto - this works and reduce duplication.

sgkit/tests/io/vcf/utils.py
tomwhite: I will make the changes to functionality in a separate PR , so this one is solely about migrating the existing code.

sgkit/io/vcf/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    from .vcf_reader import vcf_to_zarr, vcf_to_zarrs, zarrs_to_dataset  # noqa: F401
+
+    __all__ = [
+        "partition_into_regions",
+        "vcf_to_zarr",
+        "vcf_to_zarrs",
+        "zarrs_to_dataset",
+    ]
+except ImportError as e:
+    msg = (
```
ravwojdyla, line 12: I wonder if we should also check here if the OS is Windows, and if it is, instead of suggesting `pip install` straight up say that `sgkit-vcf` doesn't support Windows.

ravwojdyla (approved):

sgkit/io/vcf/__init__.py
jeromekelleher: I agree, it will save save people some time if we just tell them sgkit-vcf isn't supported on windows.

jeromekelleher (approved):
LGTM!

sgkit/io/vcf/__init__.py
ravwojdyla: And we could also mentioned that if the user wants to work with VCF data on Windows, they could convert that data into zarr and then still use sgkit on Windows via `read_vcfzarr`, essentially to prevent the impression of sgkit+vcf on Windows being a complete showstopper.

tomwhite: Good suggestion. I've added a Windows-specific message, referring users to the documentation, which is where I'll add something about the `read_vcfzarr` workaround.

Issue #251, opened 2020-09-14
Stop using __all__ for top-level namespace


jeromekelleher:
We seem to have a lot of "git conflict potential" in the current setup. The two main places seem to be in the ``known_third_party`` in setup.cfg and manually listing all API functions in ``__init__.py``. I can see how ``known_third_party`` will settle down after a while, but manually updating  ``__init__.py`` for every new function is going to lead to lots of conflicts and be pretty tedious for maintainers.

Is it worth keeping such a strict hold on the top-level namespace? People can still use undocumented methods by accessing the individual packages, so I've never really seen the point in manually managing the ``__all__`` variable.

ravwojdyla:
I have looked into an alternatives to #294, and I start to lean more towards keeping both `__all__` and imports. The reasoning: without `__all__` global names are not reexported for type checking purposes unless we either specify mypy's `implicit_reexport` or change all imports to `import x as x` style ([doc](https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport)). Initially setting `implicit_reexport` in sgkit might seem like an attractive solution, but it would also force sgkit users (who use mypy) to either also set `implicit_reexport` or they might see mypy errors (missing symbols from sgkit inits). As mentioned above alternative to `__all__` could be the `import x as x` style, one issue with that solution is that the `__all__` approach has a bit better docstring support, for example `help(sgkit)` would include help for `__all__` symbols unlike the `import x as x` solution. Overall, at this point in time, in my mind, the user benefits are worth the sporadic PR conflicts.

jeromekelleher:
OK, sounds good @ravwojdyla. Perhaps it's worth a quick comment in the top-level ``__init__.py`` explaining to developers why we're doing it this way? I feel like we'll be returning to this over and again as new contributors come along otherwise.

ravwojdyla:
Closing via https://github.com/pystatgen/sgkit/pull/359

PR #294, opened 2020-10-01
Proposal to simplify __init__.py

jeromekelleher:
Here's what I'd propose for #251: just import * from all of our component modules. Only non-private functions are exported this way:

```python
import sgkit as sg
# Both these functions are in sgkit.stats.popgen
print(sg.diversity) # Works as before
print(sg._pairwise_sum) # Fails with AttributeError
```

I don't think we're losing much, to be honest, and it avoids a fair bit of fiddly maintenance and merge conflicts. If people use ``from sgkit import *`` then they'll get a few extra things that we import (like zarr, xr, etc) as well into their namespace, but other than that I don't think there's much difference.

There probably is some stuff in there now that we would like to tidy up a bit, so maybe this is a bit extreme, but I thought I'd open the PR to see what people think.

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -1,40 +1,10 @@
-from .display import display_genotypes
-from .io.vcfzarr_reader import read_vcfzarr
-from .model import (  # noqa: F401
-    DIM_ALLELE,
-    DIM_PLOIDY,
-    DIM_SAMPLE,
-    DIM_VARIANT,
-    create_genotype_call_dataset,
-    create_genotype_dosage_dataset,
-)
-from .stats.aggregation import count_call_alleles, count_variant_alleles, variant_stats
-from .stats.association import gwas_linear_regression
-from .stats.hwe import hardy_weinberg_test
-from .stats.pc_relate import pc_relate
-from .stats.popgen import Fst, Tajimas_D, divergence, diversity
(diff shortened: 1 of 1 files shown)


ravwojdyla (changes requested):
-1 to the current version, reasons:

1. it goes against official [python doc](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package):

> Although certain modules are designed to export only names that follow certain patterns when you use import *, it is still considered bad practice in production code.

2. this change would also require changes to all the modules to filtering out all the names that we actually want to surface out, either by renaming or using `__all__` within those submodules. so this PR is somewhat incomplete.

We could still keep the original imports, just remove the `__all__`, this would not prevent the sporadic conflicts tho. I suppose those conflicts come up frequently because we change the public API so much, in the future it shouldn't be much of a problem.

jeromekelleher:
OK, closing this since the only opinion offered is that we shouldn't do it.  I think we really should get rid of the ``__all__`` variable though, as it is pointless writing the same information down twice.

ravwojdyla:
> I think we really should get rid of the __all__ variable though, as it is pointless writing the same information down twice.

@jeromekelleher I'm +1 to that, but we should also keep the imports unchanged.

PR #302, opened 2020-10-05
Add Hudson estimator to Fst, and make it the default

### LINKED ISSUE #292: Implement Hudson estimator for Fst
This is so we can compare Fst results with scikit-allel.

See #225

tomwhite:
Also, compute Fst from single divergence matrix which has diversity values on the diagonal. This idea is from tskit, which [does the same thing](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.divergence). The advantage of this approach will come with windowing, since only one array (divergence) will need to be windowed, rather than two (diversity and divergence).

Fixes #292


tomwhite:
@eric-czech any chance you can take a look at this one?

eric-czech (approved):
sure thing @tomwhite, LGTM

sgkit/tests/test_popgen.py
```diff
@@ -68,18 +69,51 @@ def test_divergence(size, n_cohorts):
...
     for i, j in itertools.combinations(range(n_cohorts), 2):
         ts_div[i, j] = ts.divergence([subsets[i], subsets[j]], span_normalise=False)
-        ts_div[j, i] = ts_div[i, j]
+        ts_div[j, i] = ts.divergence([subsets[j], subsets[i]], span_normalise=False)
     np.testing.assert_allclose(div, ts_div)
 
 
+@pytest.mark.parametrize(
+    "size, n_cohorts",
+    [(2, 2), (3, 2), (10, 2), (100, 2)],
```
eric-czech, line 88: Why is the set of of size / n_cohort pairs so different for the Hudson vs Nei tests?

sgkit/tests/test_popgen.py
```diff
@@ -68,18 +69,51 @@ def test_divergence(size, n_cohorts):
...
+)
+def test_Fst__Hudson(size, n_cohorts):
```
eric-czech, line 90: Worth making chunking a part of the tests?

sgkit/tests/test_popgen.py
tomwhite: Thanks for taking a look @eric-czech!

Hudson is tested by comparing it to scikit-allel, which only allows pairs of cohorts (populations), whereas Nei is compared to tskit, which allows any number of cohorts (and considers them in pairs). I've added a note to the test saying this.

https://github.com/pystatgen/sgkit/pull/302/commits/7cc649303c26b154b112c338d60af7bf91a9f638

sgkit/tests/test_popgen.py
tomwhite: Yes, definitely. Added a chunks parameter to all popgen tests. This exposed an issue in the divergence code, which I've now fixed.

https://github.com/pystatgen/sgkit/pull/302/commits/c9338ff8caa8deb6648aec957e7a277dfe7d293e

Issue #281, opened 2020-09-25
Port function for calculating genomic windows


eric-czech:
The [axis_intervals](https://github.com/related-sciences/gwas-analysis/blob/master/notebooks/platform/xarray/lib/stats/core.py#L102) function may be a good place to start for https://github.com/pystatgen/sgkit/issues/229.  I already started lifting this over on a branch for https://github.com/pystatgen/sgkit/issues/31 at some point, so I'll prioritize getting that in for review soon so we can see if there's anything important missing.

Let me know if this is blocking you at all @tomwhite.  I'm happy to coordinate on it if you're looking to start working on windowing right away.

tomwhite:
Thanks @eric-czech. It would be good to get something in soon. Having said that, a while ago I was looking at how to port scikit-allel's [`moving_statistic`](https://scikit-allel.readthedocs.io/en/stable/stats/window.html#allel.moving_statistic) function to Dask, so I have a [bare-bones implementation of windowing](https://github.com/tomwhite/sgkit/tree/window) that uses Dask's map_overlap, and which I can use for the time being. I think `axis_intervals` is more fully-featured, but it would be interesting to compare the two approaches (Dask delayed vs map_overlap).

eric-czech:
> Having said that, a while ago I was looking at how to port scikit-allel's moving_statistic function to Dask, so I have a bare-bones implementation of windowing that uses Dask's map_overlap, and which I can use for the time being.

👍   Something of an outstanding question for me is whether or not chromosomes should be reflected in chunking, or if they should be passed to these operations and dealt with at a higher level.  Given that it's pretty straightforward to put a chunk boundary at chromosome boundaries, I thought that was a nicer solution though I don't imagine it will ever work well with managed rechunking solutions like those mentioned in https://github.com/dask/dask/issues/6597.   It would have to know which small chunks to merge into larger preceding chunks.

It would be nice if there was an efficient way to do a groupby and then apply chromosome-agnostic functions to those groups, but that performed really terribly in my early gwas-analysis experiments.  Perhaps it's worth pushing upstream on an Xarray grouping operation that is optimized for pre-sorted groups (and probably relies on chunking like we will likely have to)?  Maybe that effort already exists.

> I think axis_intervals is more fully-featured, but it would be interesting to compare the two approaches (Dask delayed vs map_overlap).

[...]

Issue #229, opened 2020-09-02
Windowing along the genome


jeromekelleher:
Computing statistics in windows along the genome is a basic requirement. For example, users will often want to compute Fst values in (say) 100kb chunks along the genome. I think there's two basic approaches we could take here:

## Add a "windows" argument to all functions

Add an argument ``windows`` to all of the functions in which this would make sense, which would follow a similar logic to how [tskit](https://tskit.readthedocs.io/en/latest/stats.html#sec-stats-windows) does it. I think this is better than the approach taken in [scikit-allel](https://github.com/cggh/scikit-allel/blob/master/allel/stats/fst.py#L479), where we have functions like ``windowed_patterson_fst``. 

To make things concrete, here's a rough idea of what Fst would look like:
```python
def parse_windows(ds, windows)
      L = largest_variant_position(ds) + 1 # assuming we can get this
      if windows is None:
          # By default we have one window
           windows = [0, L]    
      elif isinstance(windows, str):
            # Obviously this is stupid and we'd parse out the suffix
            if windows == "100kb":
                step_size = 100_000
           else:
                 raise ValueError("Unrecognised window description")
           windows = np.arange(0, L, step=step_size)
     elif isinstance(windows, int):   # This is brittle and we'd probably do something better in reality
          # Interpret the argument as asking for n windows
          windows = np.linspace(0, L, num=windows)
     # Otherwise, assume that windows is a 1D array of n values describing n intervals along the genome.
     return windows
     

def Fst(ds, *, windows=None):
      windows = parse_windows
      # compute Fst in variants in these windows and return an array/scalar depending on the input.
```   

So, in the most general case, ``windows`` is a 1D array of n coordinates describing n - 1 intervals along the genome. The intervals are half-open (left-inclusive, right-exclusive). We also provide some other handy ways of specifying common types of windows by interpreting different argument types differently. These bells and whistles probably aren't necessary initially, though.

## Store stats and window afterwards

The previous formulation assumes that what we're returning is a single value or numpy array, but this is somewhat at odds with current thinking. In #103 we are assuming that we update the input data set with a variable for each thing that we compute on the dataset. Thus, an alternative way we might do things is that we compute the value of  ``Fst`` for each variant and store the ``Fst`` array in the result dataset. Then, windowing is something that we do *afterwards*, by aggegating values. E.g.,

```
ds = sgkit.Fst(ds)  # ds now contains a "stats_Fst" variable, or similar
[...]

PR #303, opened 2020-10-05
Fst windowing

tomwhite:
This is a proposal for #229, and is also related to #281.

* Adds a new `window` function that adds windowing information to a dataset (a `windows` dimension). This just adds fixed-size windows at the moment, and doesn't take contigs into account - both could be improved in the future.
* Adds a `window_statistic` function (similar scikit-allel's `moving_statistic`) that computes a statistic for each window (using Dask's `map_overlap` function) for use by method functions (i.e. this is not a user-facing function).
* Updates `diversity`, `divergence`, and `Fst` to use windows if present.

Sample usage:

```python
>>> ds = window(ds, size=25, step=25)
>>> fst_ds = Fst(ds, estimator="Nei")
>>> fst_ds
<xarray.Dataset>
Dimensions:              (alleles: 2, cohorts: 2, cohorts_0: 2, cohorts_1: 2, ploidy: 1, samples: 10, variants: 98, windows: 4)
Coordinates:
  * cohorts_0            (cohorts_0) object 'co_0' 'co_1'
  * cohorts_1            (cohorts_1) object 'co_0' 'co_1'
Dimensions without coordinates: alleles, cohorts, ploidy, samples, variants, windows
Data variables:
    stat_Fst             (windows, cohorts_0, cohorts_1) float64 dask.array<chunksize=(4, 2, 2), meta=np.ndarray>
    cohort_allele_count  (variants, cohorts, alleles) int64 dask.array<chunksize=(98, 2, 2), meta=np.ndarray>
    call_allele_count    (variants, samples, alleles) uint8 dask.array<chunksize=(98, 10, 2), meta=np.ndarray>
    window_start         (windows) int64 0 25 50 75
    window_stop          (windows) int64 25 50 75 98
    variant_contig       (variants) int64 0 0 0 0 0 0 0 0 0 ... 0 0 0 0 0 0 0 0
    variant_position     (variants) int64 0 2 6 8 12 19 ... 193 193 194 196 197
    variant_allele       (variants, alleles) |S1 b'0' b'1' b'0' ... b'0' b'1'
    sample_id            (samples) <U5 'tsk_0' 'tsk_1' ... 'tsk_8' 'tsk_9'
    call_genotype        (variants, samples, ploidy) int8 0 0 0 1 1 ... 1 0 0 1
    call_genotype_mask   (variants, samples, ploidy) bool False False ... False
    sample_cohort        (samples) int32 0 0 0 0 0 1 1 1 1 1
>>> fst = fst_ds["stat_Fst"].values
>>> fst
[[[       nan 0.04802022]
  [0.04802022        nan]]

 [[       nan 0.05591201]
  [0.05591201        nan]]

 [[       nan 0.03881701]
  [0.03881701        nan]]

 [[       nan 0.02708124]
  [0.02708124        nan]]]
```

Note the `window_start` and `window_stop` variables of dimension `windows`, and the `stat_Fst` variable whose first dimension is `windows`.

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -8,6 +8,7 @@ from xarray import Dataset
 from sgkit.stats.utils import assert_array_shape
 from sgkit.typing import ArrayLike
 from sgkit.utils import conditional_merge_datasets
+from sgkit.window import has_windows, window_statistic
 
 from .. import variables
 from .aggregation import count_cohort_alleles, count_variant_alleles
@@ -68,15 +69,33 @@ def diversity(
     # replace zeros to avoid divide by zero error
     n_pairs_na = n_pairs.where(n_pairs != 0)
     pi = n_diff / n_pairs_na
-    pi_sum = pi.sum(axis=0, skipna=False)
-    new_ds = Dataset(
-        {
-            variables.stat_diversity: (
-                "cohorts",
-                pi_sum,
-            )
-        }
-    )
+
+    if has_windows(ds):
+        div = window_statistic(
+            pi,
+            np.sum,
+            ds.window_start.values,
+            ds.window_stop.values,
+            dtype=pi.dtype,
+            axis=0,
+        )
+        new_ds = Dataset(
+            {
+                "stat_diversity": (
+                    ("windows", "cohorts"),
+                    div,
+                )
+            }
+        )
+    else:
+        new_ds = Dataset(
+            {
+                "stat_diversity": (
+                    ("variants", "cohorts"),
+                    pi,
+                )
+            }
+        )
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
 
 
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -197,10 +197,32 @@ def divergence(
     d = da.map_blocks(_divergence, ac, chunks=shape, dtype=np.float64)
     assert_array_shape(d, n_variants, n_cohorts, n_cohorts)
 
-    d_sum = d.sum(axis=0)
-    assert_array_shape(d_sum, n_cohorts, n_cohorts)
-
-    new_ds = Dataset({variables.stat_divergence: (("cohorts_0", "cohorts_1"), d_sum)})
+    if has_windows(ds):
+        div = window_statistic(
+            d,
+            np.sum,
+            ds.window_start.values,
+            ds.window_stop.values,
+            dtype=d.dtype,
+            axis=0,
+        )
+        new_ds = Dataset(
+            {
+                "stat_divergence": (
+                    ("windows", "cohorts_0", "cohorts_1"),
+                    div,
+                )
+            }
+        )
+    else:
+        new_ds = Dataset(
+            {
+                "stat_divergence": (
+                    ("variants", "cohorts_0", "cohorts_1"),
+                    d,
+                )
+            }
+        )
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
 
 
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -348,10 +348,11 @@ def Fst(
         ds, allele_counts=allele_counts, call_genotype=call_genotype, merge=False
     ).stat_divergence
     gs = da.asarray(gs)
-    shape = (n_cohorts, n_cohorts)
+    shape = (gs.chunks[0], n_cohorts, n_cohorts)
     fst = da.map_blocks(known_estimators[estimator], gs, chunks=shape, dtype=np.float64)
-    assert_array_shape(fst, n_cohorts, n_cohorts)
-    new_ds = Dataset({variables.stat_Fst: (("cohorts_0", "cohorts_1"), fst)})
+    # TODO: reinstate assert (first dim could be either variants or windows)
+    # assert_array_shape(fst, n_windows, n_cohorts, n_cohorts)
+    new_ds = Dataset({variables.stat_Fst: (("windows", "cohorts_0", "cohorts_1"), fst)})
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
 
 
--- a/sgkit/tests/test_popgen.py
+++ b/sgkit/tests/test_popgen.py
@@ -48,10 +48,10 @@ def ts_to_dataset(ts, chunks=None, samples=None):
     return ds
 
 
-@pytest.mark.parametrize("size", [2, 3, 10, 100])
+@pytest.mark.parametrize("sample_size", [2, 3, 10, 100])
 @pytest.mark.parametrize("chunks", [(-1, -1), (10, -1)])
-def test_diversity(size, chunks):
-    ts = msprime.simulate(size, length=100, mutation_rate=0.05, random_seed=42)
+def test_diversity(sample_size, chunks):
+    ts = msprime.simulate(sample_size, length=100, mutation_rate=0.05, random_seed=42)
     ds = ts_to_dataset(ts, chunks)  # type: ignore[no-untyped-call]
     ds = ds.chunk(dict(zip(["variants", "samples"], chunks)))
     sample_cohorts = np.full_like(ts.samples(), 0)
@@ -63,9 +63,9 @@ def test_diversity(size, chunks):
     np.testing.assert_allclose(div, ts_div)
 
 
-@pytest.mark.parametrize("size", [10])
-def test_diversity__windowed(size):
-    ts = msprime.simulate(size, length=200, mutation_rate=0.05, random_seed=42)
+@pytest.mark.parametrize("sample_size", [10])
+def test_diversity__windowed(sample_size):
+    ts = msprime.simulate(sample_size, length=200, mutation_rate=0.05, random_seed=42)
     ds = ts_to_dataset(ts)  # type: ignore[no-untyped-call]
     sample_cohorts = np.full_like(ts.samples(), 0)
     ds["sample_cohort"] = xr.DataArray(sample_cohorts, dims="samples")
@@ -93,12 +93,12 @@ def test_diversity__windowed(size):
 
 
 @pytest.mark.parametrize(
-    "size, n_cohorts",
+    "sample_size, n_cohorts",
     [(2, 2), (3, 2), (3, 3), (10, 2), (10, 3), (10, 4), (100, 2), (100, 3), (100, 4)],
 )
 @pytest.mark.parametrize("chunks", [(-1, -1), (10, -1)])
-def test_divergence(size, n_cohorts, chunks):
-    ts = msprime.simulate(size, length=100, mutation_rate=0.05, random_seed=42)
+def test_divergence(sample_size, n_cohorts, chunks):
+    ts = msprime.simulate(sample_size, length=100, mutation_rate=0.05, random_seed=42)
     subsets = np.array_split(ts.samples(), n_cohorts)
     ds = ts_to_dataset(ts, chunks)  # type: ignore[no-untyped-call]
     sample_cohorts = np.concatenate(
@@ -124,10 +124,10 @@ def test_divergence(size, n_cohorts, chunks):
     np.testing.assert_allclose(div, ts_div)
 
 
-@pytest.mark.parametrize("size, n_cohorts", [(10, 2)])
+@pytest.mark.parametrize("sample_size, n_cohorts", [(10, 2)])
 @pytest.mark.parametrize("chunks", [(-1, -1), (50, -1)])
-def test_divergence__windowed(size, n_cohorts, chunks):
-    ts = msprime.simulate(size, length=200, mutation_rate=0.05, random_seed=42)
+def test_divergence__windowed(sample_size, n_cohorts, chunks):
+    ts = msprime.simulate(sample_size, length=200, mutation_rate=0.05, random_seed=42)
     subsets = np.array_split(ts.samples(), n_cohorts)
     ds = ts_to_dataset(ts, chunks)  # type: ignore[no-untyped-call]
     sample_cohorts = np.concatenate(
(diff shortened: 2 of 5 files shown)


tomwhite:
BTW this is based on #302, which should be merged first.

alimanfoo:
Hi @tomwhite, one small thought is that I'll often try computing some statistic with a variety of different window sizes. So we might need some way to specify a name or prefix for each set of windows being added to the same dataset.

tomwhite:
@alimanfoo that's good to know. Another way of achieving this would be to set `merge=False` to get a dataset for each set of windows.

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+                for i, j in zip(chunk_window_starts, chunk_window_stops)
+            ]
+        )
+        return out
+
+    if values.ndim == 1:
+        new_chunks = (tuple(windows_per_chunk),)
+    else:
+        # depth is 0 except in first axis
+        depth = tuple([depth] + ([0] * (values.ndim - 1)))
```
ravwojdyla, line 138: I wonder if `map_overlap` would work in this case with `depth = {0: depth}` (without the need to specify the depth on the other dimensions) 🤔

sgkit/tests/test_window.py
```diff
@@ -0,0 +1,46 @@
...
+    stat = moving_statistic(values, sum_cols, size=size, step=step, dtype=values.dtype)
+    stat = stat.compute()
+    if length % size != 0 or size != step:
+        # scikit-allel misses final window in this case
+        stat = stat[:-1]
+
+    values_sa = arr
+    stat_sa = allel.moving_statistic(values_sa, sum_cols, size=size, step=step)
+
+    np.testing.assert_equal(stat, stat_sa)
```
ravwojdyla, line 46: Would it make sense to add more tests here, there are some interesting functions in the `window` package, like `_get_chunked_windows`? Plus test for corner cases like empty arrays etc. And maybe also a test for what happens when you window multiple times with `merge=True`. And maybe a test case for https://github.com/pystatgen/sgkit/pull/303#issuecomment-705682385

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+import dask.array as da
+import numpy as np
+from xarray import Dataset
+
+from sgkit.utils import conditional_merge_datasets
+
+# Window definition (user code)
+
+
+def window(
```
ravwojdyla, line 12: Thinking about the API, I wonder if we should have a simple function to "remove" the windowing from a dataset? Otherwise we might expect people to pop `window_start`/`window_stop` vars?

ravwojdyla (changes requested):
I like the idea of storing the window information in the `Dataset`!

tomwhite:
Thanks for the review @ravwojdyla! I will fix your suggestions given that the overall approach seems to be OK.

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+                window_stops,
+            ),
+        }
+    )
+    return conditional_merge_datasets(ds, new_ds, merge)
+
+
+def _get_windows(start: int, stop: int, size: int, step: int) -> Any:
+    # Find the indexes for the start positions of all windows
+    # TODO: take contigs into account
```
eric-czech, line 42: Worth elevating as an issue?  I'm sure it will come up again in the context of LD estimation.

sgkit/window.py
eric-czech: That should work: https://github.com/dask/dask/blob/master/dask/array/overlap.py#L732 is the relevant bit for that (all others default to 0).

sgkit/tests/test_window.py
```diff
@@ -0,0 +1,46 @@
...
+    stat = stat.compute()
+    if length % size != 0 or size != step:
+        # scikit-allel misses final window in this case
+        stat = stat[:-1]
+
+    values_sa = np.arange(length)
+    stat_sa = allel.moving_statistic(values_sa, np.sum, size=size, step=step)
+
+    np.testing.assert_equal(stat, stat_sa)
+
```
eric-czech, line 25: I think a test for dtype preservation when provided float32 inputs would be good for these kinds of functions.

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+            "window_stop": (
+                "windows",
+                window_stops,
+            ),
+        }
+    )
+    return conditional_merge_datasets(ds, new_ds, merge)
+
+
+def _get_windows(start: int, stop: int, size: int, step: int) -> Any:
```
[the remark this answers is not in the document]
tomwhite: Changed to `ArrayType`.

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+        raise ValueError(
+            f"Minimum chunk size ({min_chunksize}) must not be smaller than size ({size})."
+        )
+    window_starts, window_stops = _get_windows(0, length, size, step)
+    return window_statistic(
+        values, statistic, window_starts, window_stops, dtype, **kwargs
+    )
+
+
+def window_statistic(
```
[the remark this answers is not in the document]
tomwhite: This was from when I was prototyping, and I forgot to fix them before submitting. I have now eliminated them all except `kwargs` and `block_info`, which are harder to provide hints for.

sgkit/stats/popgen.py
```diff
@@ -68,15 +69,33 @@ def diversity(
...
-            )
-        }
-    )
+
+    if has_windows(ds):
+        div = window_statistic(
+            pi,
+            np.sum,
+            ds.window_start.values,
+            ds.window_stop.values,
```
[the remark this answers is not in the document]
tomwhite: No, not at the moment at least.

sgkit/tests/test_popgen.py
```diff
@@ -49,11 +58,40 @@ def test_diversity(size, chunks):
     ds["sample_cohort"] = xr.DataArray(sample_cohorts, dims="samples")
     ds = ds.assign_coords({"cohorts": ["co_0"]})
     ds = diversity(ds)
-    div = ds.stat_diversity.sel(cohorts="co_0").values
+    div = ds.stat_diversity.sum(axis=0, skipna=False).sel(cohorts="co_0").values
     ts_div = ts.diversity(span_normalise=False)
     np.testing.assert_allclose(div, ts_div)
 
 
+@pytest.mark.parametrize("size", [10])
```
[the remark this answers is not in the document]
tomwhite: Good point - especially as we now have sample size and window size in the same test.

sgkit/tests/test_popgen.py
```diff
@@ -49,11 +58,40 @@ def test_diversity(size, chunks):
...
+    ds = window(ds, size=25, step=25)
+    ds = diversity(ds)
+    div = ds["stat_diversity"].sel(cohorts="co_0").compute()
+
+    # Calculate diversity using tskit windows
+    # Find the variant positions so we can have windows with a fixed number of variants
+    positions = [variant.site.position for variant in ts.variants()]
+    windows = [0] + positions[::25][1:] + [ts.sequence_length]
+    ts_div = ts.diversity(windows=windows, span_normalise=False)
+    np.testing.assert_allclose(div, ts_div)
```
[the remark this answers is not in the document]
tomwhite: Good idea. I opened #332

sgkit/tests/test_popgen.py
```diff
@@ -86,20 +124,64 @@ def test_divergence(size, n_cohorts, chunks):
...
+    # Calculate divergence using scikit-allel moving_statistic
+    # (Don't use windowed_divergence, since it treats the last window differently)
+    ds1 = count_variant_alleles(ts_to_dataset(ts, samples=ts.samples()[:1]))  # type: ignore[no-untyped-call]
+    ds2 = count_variant_alleles(ts_to_dataset(ts, samples=ts.samples()[1:]))  # type: ignore[no-untyped-call]
+    ac1 = ds1["variant_allele_count"].values
+    ac2 = ds2["variant_allele_count"].values
+    mpd = allel.mean_pairwise_difference_between(ac1, ac2, fill=0)
+    ska_div = allel.moving_statistic(mpd, np.sum, size=25, step=25)  # noqa: F841
+    # TODO: investigate why numbers are different
+    # np.testing.assert_allclose(div[:-1], ska_div)  # scikit-allel has final window missing
```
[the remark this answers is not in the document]
tomwhite: Done

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+    """Test if a dataset has windowing information."""
+    return "window_start" in ds and "window_stop" in ds
+
+
+def moving_statistic(
+    values: Any,
+    statistic: Callable[..., Any],
+    size: int,
+    step: int,
+    dtype: int,
```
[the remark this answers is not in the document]
tomwhite: Done

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+# Window definition (user code)
+
+
+def window(
+    ds: Dataset,
+    size: int,
+    step: int,
+    merge: bool = True,
+) -> Dataset:
+    """Add windowing information to a dataset."""
```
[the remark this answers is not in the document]
tomwhite: I expanded this, and added new variables.

sgkit/window.py
```diff
@@ -0,0 +1,174 @@
...
+
+    # Dask will raise an error if the last chunk size is smaller than the depth
+    # Workaround by rechunking to combine the last two chunks in first axis
+    # See https://github.com/dask/dask/issues/6597
+    if depth > values.chunks[0][-1]:
+        chunk0 = values.chunks[0]
+        new_chunk0 = tuple(list(chunk0[:-2]) + [chunk0[-2] + chunk0[-1]])
+        # None means don't rechunk along that axis
+        new_chunks = tuple(list([new_chunk0] + ([None] * (len(chunk0) - 1))))  # type: ignore
+        values = values.rechunk(new_chunks)
```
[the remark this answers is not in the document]
tomwhite: Thanks - done

sgkit/window.py
tomwhite: Thanks for checking - fixed

sgkit/window.py
tomwhite: Yes, that would be useful. Opened #333

sgkit/window.py
tomwhite: Opened #335

sgkit/tests/test_window.py
tomwhite: I added an assert to check for dtype preservation here.

tomwhite:
I've addressed all the feedback, except for the comment about adding more tests for the window module. I'll try to write some more, but otherwise this is probably ready to be merged.

sgkit/stats/popgen.py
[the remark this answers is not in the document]
eric-czech: Should it be?  The loose guideline I have in my head is that anything that doesn't run interactively (<1s or so) across a whole genome dataset (~100M variants) should probably be lazy.  Is it worth adding an issue to make this lazy or test that the latter is true?

eric-czech (approved):

sgkit/window.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Windows are defined over the ``variants`` dimension, and are
+    used by some downstream functions to calculate statistics for
+    each window.
+
+    Parameters
+    ----------
+    ds
+        Genotype call dataset.
+    size
```
jeromekelleher, line 30: Hmm, I'm not so sure about this API, it feels a bit inflexible. People are going to want to window and plot in terms of physical distance and not just in terms of the number of variants (see the proposal in #229). I see that using the number of variants aligns much more nicely with chunking. Is this a fundamental restriction - will we always need windows defined in this way to have equal numbers of variants? If we did want to do things like "compute Fst in 100Kb windows", is this incompatible with the way we're doing things in Dask?

It's not obvious what the utility of the ``step`` parameter is either - I'd imagine the vast majority of people will want contiguous windows (unless I'm misunderstanding what this does?). It should at least have a default such that users do get contiguous windows and don't need to specify this value all the time.

jeromekelleher (commented):
This looks good, but I worry that we're painting ourselves into a bit of a corner with the names of things here, given that this is so tied to variant indexes. How will we deal with windowing in physical distance? I think we should have a good idea how this is going to work and how we're going to name things before finalising this API.

sgkit/tests/test_window.py
tomwhite: Added tests for various window functions in https://github.com/pystatgen/sgkit/pull/303/commits/d57aefdb889fd9190b44d4439ba75e91ac8201b9

sgkit/tests/test_popgen.py
```diff
@@ -39,28 +48,57 @@ def ts_to_dataset(ts, chunks=None, samples=None):
...
+    sample_cohorts = np.full_like(ts.samples(), 0)
+    ds["sample_cohort"] = xr.DataArray(sample_cohorts, dims="samples")
+    ds = ds.assign_coords({"cohorts": ["co_0"]})
+    ds = window(ds, size=25, step=25)
+    ds = diversity(ds)
+    div = ds["stat_diversity"].sel(cohorts="co_0").compute()
+
+    # Calculate diversity using tskit windows
+    # Find the variant positions so we can have windows with a fixed number of variants
+    positions = [variant.site.position for variant in ts.variants()]
```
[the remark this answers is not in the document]
tomwhite: Thanks for the suggestion - I've used `ts.tables.sites.position` instead

sgkit/stats/popgen.py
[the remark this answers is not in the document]
tomwhite: Opened #340

sgkit/window.py
jeromekelleher: I really like this idea @tomwhite - how about we create a follow up issue for this so, and get this much merged?

jeromekelleher:
@ravwojdyla - have the changes you requested been made?

ravwojdyla (approved):

jeromekelleher (approved):

jeromekelleher:
Looks great @tomwhite - please add the auto-merge flag when you're happy to merge this (I guess with an issue flagging the current limitations as discussed above)

sgkit/window.py
tomwhite: Filed #341

Issue #241, opened 2020-09-08
Pairwise distance calculation


jeromekelleher:
A fundamental operation that we want to support is efficiently calculating pairwise distances over a number of different metrics across a number of different computational backends. See the [prototype](
https://github.com/scikit-allel/skallel-stats/tree/master/src/skallel_stats/distance) as an example of what we're trying to achieve.

@alimanfoo - I think you had a blog post about this also?

alimanfoo:
Hi @jeromekelleher, yes @jakirkham and I started a blog post last year but afraid I haven't had time to see it through. FWIW it's over here: https://github.com/dask/dask-blog/pull/50

jakirkham:
It would still be cool to finish that blog 😉

jeromekelleher:
Just had a chat with @aktech, and he's going to have a look at this.

aktech:
Some of the potential options that I am looking at for dispatching:

### NEP 37
`__array_module__ `: NEP 37: https://numpy.org/neps/nep-0037-array-module.html
This is in draft, but ready to be used, it will be merged in the numpy by end of this year meanwhile we can use the external package from here: https://github.com/seberg/numpy-dispatch, developed by the author of NEP37 (also a numpy core developer.)

### NEP 18
`__array_function__ `: NEP 18: https://numpy.org/neps/nep-0018-array-function-protocol.html
This is not great for long-term as NEP37 is a better successor due to the following reasons:

- Overrides existing numpy namespace
- Currently requires all or nothing approach to implementing numpy’s API
- No good pathway for incremental adoption
- Would result in breaking changes
- Backward compatibility concerns
- Limitation on what can be overridden

### unumpy

> unumpy builds on top of uarray. It is an effort to specify the core NumPy API, and provide backends for the API.

- https://unumpy.uarray.org/en/latest/generated/unumpy.html
This is a new project incubated by Quansight

alimanfoo:
Just to add a bit more information about some motivating use cases. Prime motivation was to analyse population structure by computing pairwise genetic distance between individuals. From the distance matrix we would typically compute and visualise a neighbour joining tree. Here's what such a tree might look like:

![Screenshot from 2020-09-24 17-24-54](https://user-images.githubusercontent.com/703554/94172832-e76f6280-fe8a-11ea-99d4-fddaea6fdb62.png)

[...]

PR #306, opened 2020-10-05
Implement pairwise distance calculation

aktech:
This is an attempt to implement: #241 

It is based on @eric-czech 's suggestion here: https://github.com/pystatgen/sgkit/issues/241#issuecomment-698878378. It implements Euclidean and Pearson correlation using the Map/reduce idea. I have tried to document the algorithm in the pairwise function. It only calculates the upper triangular matrix, to avoid repeated calculations. For the public API, I have taken inspiration from [`scipy.spatial.distance.pdist`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html). I haven't added it to the `api.rst` yet.

I haven't ran this on GPU, but for the API, I am thinking to use, `target` (`=cuda`) argument of [`guvectorize`](https://numba.pydata.org/numba-doc/latest/cuda/ufunc.html#generalized-cuda-ufuncs). From user's point of view, I think we can let the user set it either via environment variable like say `SGKIT_USE_GPU` or via some kind of sgkit config, similar to dask's config object, and then we can compile the `guvectorize` functions to run on GPU. I am open to opinions on both.

cc @eric-czech @jeromekelleher


eric-czech:
Thanks @aktech!  Looks good, nice work figuring out how to omit the blocks for one of the triangular matrices.

I was feeling a little nervous about the map/reduce recommendation so I did some more research and found a couple key things:

1. Dask array ops like `matmul` (which is a pairwise calculation when multiplying a matrix by itself) are apparently not actually block-scalable in both dimensions.  I was surprised that they work this way, but the `da.blockwise` usage like [this](https://github.com/dask/dask/blob/8a699223982ce8425f59f048d0163090b4a81193/dask/array/routines.py#L313) pairs up whole rows of blocks, concatenates them all together and then runs `np.matmul` on them.  This is effectively like running an `x.rechunk((None, -1))` beforehand, which is a bit disappointing.
2. `guvectorize` definitely does *not* materialize arrays that need to be broadcast together

It is certainly worth considering the fact that the level of scalability we're trying for here isn't something dask does in core operations so I put together a bunch of different options and compared them in [this notebook](https://nbviewer.jupyter.org/gist/eric-czech/eb1913775b20456c4dc9eac6677dab9e).  My takeaways from this:

1. The only way to get performance comparable to functions like `da.corrcoef` is to feed `guvectorize` 2D inputs and apply vectorized functions within that custom function.  This is quite hard though since numba doesn't implement a lot of numpy functions that would make this more usable and realistically, we need to be able to run our own loops to handle things like missing values and metrics that can't be expressed as numpy.
2. Everything is significantly faster than the implementation I suggested **unless** you rechunk to remove column chunks first, in which case it's about the same as most of the others
    - This makes sense since this implementation should scale linearly with the number of column chunks
3. Using `njit`/`cuda.jit` instead of `guvectorize` doesn't help 

**tl;dr** I think it still makes sense to move forward with this implementation, but we need to communicate the fact that performance at least somewhat comparable to functions like `da.corrcoef` is only possible when column chunks are removed.  I would suggest we make this the default behavior.  And if we encounter a metric that won't work in the map/reduce model, this is a nice simple alternative:

```python
@guvectorize(['void(int8[:], int8[:], float64[:])'], '(n),(n)->()', fastmath=True)
def corrcoef(x, y, out):
    # Equivalent to np.corrcoef(x, y)[0, 1] but 2x faster
    cov = ((x - x.mean()) * (y - y.mean())).sum() 
    out[0] = cov / (x.std() * y.std()) / x.shape[0]
    
def pairwise(x, fn):
    return da.blockwise(
        # Lambda wraps reshape for broadcast
        lambda x, y: fn(x[:, None, :], y), 'jk', 
        x, 'ji', x, 'ki', 
        dtype='float64',
        concatenate=True
    )
    
pairwise(x, corrcoef)
```

I'm also thinking that there is a way to express the loop we have now that references `.blocks` as a `blockwise` call like this, though I haven't been able to get it to work yet.

aktech:
@eric-czech Thanks for the analysis, that's really very useful. The option 6 is indeed much faster than 5 with a slight change, let me try to get rid of the column chunks, to incorporate that. Also, in the current implementation since we omit the blocks for one of the triangular matrices, it should be even more faster!

sgkit/distance/core.py
```diff
@@ -0,0 +1,163 @@
...
+    x: ArrayLike,
+    map_fn: np.ufunc,
+    reduce_fn: np.ufunc,
+    n_map_param: int,
+    chunks: typing.Optional[typing.Tuple[int, int]] = None,
+) -> ArrayLike:
+    """A generic pairwise function for any separable distance metric.
+    This calculates the pairwise distance between a set of vectors in the
+    given two-dimensional array, using the map reduce ufuncs. To illustrate
+    the algorithm consider the following (4, 5) two dimensional array:
```
alimanfoo, line 20: The example array is (4, 4), so need to either change the example or the stated shape. Maybe change the example to be non-square, so it's clear whether you are treating rows or columns as vectors to be compared?

sgkit/distance/core.py
```diff
@@ -0,0 +1,163 @@
...
+
+    The (i, j) position in the resulting array (matrix) denotes the distance
+    between vi and vj vectors.
+
+    The given array can be either of type numpy or dask array. For better
+    performance, it is recommended to use dask array with suitable
+    chunking, which makes it highly parallelizable.
+
+    Algorithm: This function works on the philosophy of map/reduce style
+    computation.The map step, which is implemented in the given map_fn
```
alimanfoo, line 48:
```suggestion
    computation. The map step, which is implemented in the given map_fn
```

sgkit/distance/core.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Now instead of passing the whole vectors: v0, v1, v2, v3 for the
+    calculation of the distance, only chunks of them are passed and the map
+    function calculates partial, between the chunks of two or more vectors and
+    this calculation can happen in parallel, hence speeding up the whole process.
+
+    After the partial calculations of the chunks are done in the map step, they
+    are accumulated in the reduce step, via the given reduce function.
+
+    The importance of ``chunks`` can be seen from the mentioned algorith above,
```
alimanfoo, line 71:
```suggestion
    The importance of ``chunks`` can be seen from the mentioned algorithm above,
```

sgkit/distance/core.py
```diff
@@ -0,0 +1,163 @@
...
+           [0.99717646, 0.78571429, 1.        ]])
+
+    >>> x = np.array([[6, 4, 1,], [4, 5, 2], [9, 7, 3]])
+    >>> pairwise(x, correlation_map, correlation_reduce, N_MAP_PARAM['correlation'], chunks=(2, 2)).compute()
+    array([[1.        , 0.73704347, 0.99717646],
+           [0.73704347, 1.        , 0.78571429],
+           [0.99717646, 0.78571429, 1.        ]])
+
+    """
+    if isinstance(x, np.ndarray) and not chunks:
```
alimanfoo, line 113: FWIW I think it would be fine to just raise an error if the input is not a dask array, rather than trying to be helpful.

sgkit/distance/metrics.py
```diff
@@ -0,0 +1,233 @@
...
+    Parameters
+    ----------
+    x
+        An array chunk, a partial vector
+    y
+        Another array chunk,  a partial vector
+    _
+        A dummy variable to map the size of output
+    out
+        The output array, which has the output of pearson correlation.
```
alimanfoo, line 163: "pearson correlation"?

sgkit/distance/core.py
eric-czech: The approach we've taken in other functions is to say `x = da.asarray(x)` instead.  I think it's safe to assume that if a numpy array was provided, there is no reason to worry about it becoming a single chunk dask array.

eric-czech:
> Also, do you have any comments on the proposed approach for running on GPU?

For posterity after talking on our dev call, it would be ideal if a user could initiate GPU computations like this:

```python
(
    ds
    .pipe(sg.count_call_alternate_alleles)
    # Move to gpu as in https://docs.dask.org/en/latest/gpu.html#arrays
    .assign(count_call_alternate_alleles=lambda ds: ds.count_call_alternate_alleles.map_blocks(cupy.asarray))
    # pairwise_distance knows how to pick guvectorize functions based on type of array._meta
    .pipe(sg.pairwise_distance, metric='euclidean') 
)
```

I think that would be best since many other dask functions will work automatically when cupy is the chunk backend.  As far as detecting this goes, `type(ddc._meta) == cupy.core.core.ndarray` does some like a reasonable way to do it (as used in dask [here](https://github.com/dask/dask/blob/bd18820642aba7388455381f872bfbc5437dc9a1/dask/array/tests/test_cupy.py#L152)).  That seems to be the intent of that attribute based on https://github.com/dask/dask/issues/2977.  I'm not sure why it's hidden.

sgkit/distance/metrics.py
aktech: map step of "pearson correlation" rather

aktech:
@eric-czech I was updating the implementation to get rid of column chunks and realised that and the last dimension is not required anymore, i.e. the complete distance calculation can be done in the map step itself, because the map step receives a full pair of vectors which means the reduce step can be performed right after it as there are no chunks for aggregation in the reduce step. Let me know if that doesn't makes sense.

eric-czech:
> I was updating the implementation to get rid of column chunks and realised that and the last dimension is not required anymore, i.e. the complete distance calculation can be done in the map step itself, because the map step receives a full pair of vectors which means the reduce step can be performed right after (together basically) it as there are no chunks for aggregation in the reduce step.

Makes sense, the custom functions can use full vectors if there are no column chunks (which may not necessarily be true).  Is there a question I can help answer with that though?

aktech:
> if there are no column chunks (which may not necessarily be true)

Since, we are getting rid of column chunks (via `x.rechunk((None, -1))`), there won't be any column chunks and it would not be map reduce anymore, there will be just one dispatch function basically (instead of two). I am not sure what you meant by "which may not necessarily be true"?

eric-czech:
> Since, we are getting rid of column chunks (via x.rechunk((None, -1))), there won't be any column chunks

When I suggested that I meant it as default but optional behavior.  There would definitely be no need for map/reduce functions if we made that mandatory.

Two things would make sense here in moving forward:

1. Keep going with the map/reduce functions
  - Make `.rechunk((None, -1))` default but optional behavior, perhaps with a `keep_chunks=False` flag
  - Validate that the performance of loops around `.blocks` references is the same as `da.blockwise` for fairly large arrays with large numbers of blocks
  - Communicate the performance/scalability tradeoff in the docs 
2. Switch to https://github.com/pystatgen/sgkit/pull/306#issuecomment-704294220
  - This wouldn't require benchmarking to prove anything 

In the interest of time/simplicity, I'll recommend switching to 2 now.  We can just get something basic in that way and iterate from there.   Does that make sense @aktech?  Then all the custom functions can work on whole vectors.

aktech:
> In the interest of time/simplicity, I'll recommend switching to 2 now. We can just get something basic in that way and iterate from there. Does that make sense @aktech? Then all the custom functions can work on whole vectors.

Yes, that makes perfect sense, but one quick question are you suggesting by moving forward with 2 now, we should skip 1 for now? i.e. in this PR. Also, things are overlapping in the 1st and 2nd point you made for example (*Make `.rechunk((None, -1))` default but optional behavior*, *Communicate the performance/scalability tradeoff in the docs*), these are mentioned on the https://github.com/pystatgen/sgkit/pull/306#issuecomment-704294220 as well, that's why I am not entirely sure about the suggestion.

eric-czech:
> Yes, that makes perfect sense, but one quick question are you suggesting by moving forward with 2 now, we should skip 1 for now

Yes.

>   Also, things are overlapping in the 1st and 2nd point you made for example (Make .rechunk((None, -1)) default but optional behavior, Communicate the performance/scalability tradeoff in the docs), these are mentioned on the #306 (comment) as well, that's why I am not entirely sure about the suggestion.

Yep -- you can omit any code or documentation related to chunking.  I meant everything in the number 1 and 2 scenarios to be exclusive.  I am imagining a single `pairwise` function that wraps the `da.blockwise` call to individual numba functions, each of which will take whole vector inputs.

eric-czech:
> @eric-czech Thanks for clarification, I have pushed the changes. I'll update the docs in a bit. Can you confirm if the change seems appropriate for what you have suggested.

Thanks @aktech, that's more or less what I was picturing.  I'll hold off on any specific questions/suggestions though until the PR comes out of draft status.

aktech:
Thanks, I'll bring this out of draft in an hour or so, after I update the documentation.

sgkit/distance/api.py
```diff
@@ -0,0 +1,62 @@
...
+    """Calculates the pairwise distance between all pairs of vectors in the
+    given two dimensional array x. The API is similar to:
+    ``scipy.spatial.distance.pdist``.
+
+
+    Parameters
+    ----------
+    x
+        [array-like, shape: (M, N)]
+        An array like two dimensional matrix
```
eric-czech, line 21: It's worth mentioning which dimension the vectors are in that will be compared (i.e. the rows).

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+import numpy as np
+
+from sgkit.typing import ArrayLike
+
+
+def pairwise(
+    x: ArrayLike,
+    metric_func: np.ufunc,
+) -> ArrayLike:
+    """A generic pairwise function for any separable distance metric.
```
eric-czech, line 11: "Separable" isn't necessary anymore

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+    This calculates the pairwise distance between a set of vectors in the
+    given two-dimensional array, using the ``metric_func`` ufunc. To illustrate
+    the algorithm consider the following (4, 5) two dimensional array:
+
+    [e.00, e.01, e.02, e.03, e.04]
+    [e.10, e.11, e.12, e.13, e.14]
+    [e.20, e.21, e.22, e.23, e.24]
+    [e.30, e.31, e.32, e.33, e.34]
+
+    The rows of the above matrix are the set of vectors. Now lets label all
```
eric-czech, line 21:
```suggestion
    The rows of the above matrix are the set of vectors. Now let's label all
```

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+    the vectors as v0, v1, v2, v3
```
eric-czech, line 22:
```suggestion
    the vectors as v0, v1, v2, v3.
```

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+
+    The result will be a two dimensional symmetric matrix which will contain
+    the distance between all pairs. Since there are 4 vectors, calculating the
+    distance of each vector with every other vector, will result in 16
```
eric-czech, line 26:
```suggestion
    distance of each vector with every other vector will result in 16
```

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+
+    [v0.v0, v0.v1, v0.v2, v0.v3]
+    [v1.v0, v1.v1, v1.v2, v1.v3]
+    [v2.v0, v2.v1, v2.v2, v2.v3]
+    [v3.v0, v3.v1, v3.v2, v3.v3]
+
+    The (i, j) position in the resulting array (matrix) denotes the distance
+    between vi and vj vectors.
+
+    The given array can be either of type numpy or dask array. For better
```
eric-czech, line 37: I think you can remove this paragraph.  The input types should be clear from the parameter docs, it's not clear what "suitable chunking" is, and I'm not sure that parallelizing with dask would necessarily increase performance.  I would assume that's not true for the majority of smaller inputs.

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+    performance, it is recommended to use dask array with suitable
+    chunking, which makes it highly parallelizable.
+
+    Parameters
+    ----------
+    x
+        [array-like, shape: (M, N)]
+        An array like two dimensional matrix.
+    metric_func
+        Map function for the distance metric.
```
eric-czech, line 47: "Map" = just metric function?  No need to refer to a "map" function now.

sgkit/distance/core.py
```diff
@@ -0,0 +1,87 @@
...
+    array([[1.        , 0.73704347, 0.99717646],
+           [0.73704347, 1.        , 0.78571429],
+           [0.99717646, 0.78571429, 1.        ]])
+    >>> pairwise(x, euclidean).compute()
+    array([[0.        , 2.44948974, 4.69041576],
+           [2.44948974, 0.        , 5.47722558],
+           [4.69041576, 5.47722558, 0.        ]])
+    """
+    x = da.asarray(x)
+    return da.blockwise(
```
eric-czech, line 77: What happened to ```return da.triu(x) + da.triu(x).T```?  Do you know if that works as expected?

sgkit/distance/metrics.py
```diff
@@ -0,0 +1,77 @@
+"""This module implements various distance metrics."""
+
+import numpy as np
+from numba import guvectorize
+
+from sgkit.typing import ArrayLike
+
+
+@guvectorize(["void(float64[:], float64[:], float64[:])"], "(n),(n)->()", fastmath=True)  # type: ignore
+def correlation(x: ArrayLike, y: ArrayLike, out: ArrayLike) -> None:
```
eric-czech, line 10: These should have some more overloads, or at least `int8` since that would be the common input to them.

sgkit/distance/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+    >>> x = da.array([4, 3, 2, 3], dtype='i1')
+    >>> y = da.array([5, 6, 7, 0], dtype='i1')
+    >>> euclidean(x, y).compute()
+    6.6332495807108
+
+    >>> euclidean(x, x).compute()
+    0.0
+
+    """
+    out[0] = np.sqrt(((x - y) ** 2).sum())
```
eric-czech, line 77: How do missing values get handled here?

sgkit/tests/test_distance.py
```diff
@@ -0,0 +1,45 @@
...
+    return x
+
+
+def test_distance_correlation() -> None:
+    x = get_vectors()
+    distance_matrix = pdist(x, metric="correlation")
+    np.testing.assert_almost_equal(distance_matrix, np.corrcoef(x).compute())
+
+
+def test_distance_euclidean() -> None:
```
eric-czech, line 26: I think missing values and chunking are two things that are worth adding to these tests (probably as parameters).

sgkit/distance/core.py
aktech: I got rid of it, as we are already returning a symmetric matrix. That was used to create a symmetric matrix from lower (or upper) triangular matrix, when we were skipping repeated calculations. Yes, it works, as the tests passed and in [tests](https://github.com/pystatgen/sgkit/pull/306/files#diff-bfebf015a7d0a911d66743edcf9806a15b4661814667c0683fef0af8a6222b35R30) we compare it with the `squareform` matrix (from `scipy`), which is a symmetric distance matrix as well.

eric-czech:
> Here is a brief comparison of speed, memory and CPU utilisation between this implementation, the previous implementation and Alistair's prototype on matrix of size 100 million: notebook.

Thanks @aktech, can you rerun that when they're computing the same outputs (i.e. only the upper triangle)?  It's difficult to draw a conclusion from that test.  Also, can you run it on a GPU or add an issue for doing so at another time?  I think pairwise calculations are a great context for testing if guvectorize can really replace custom kernels and the sooner we can get information on that the better.

sgkit/distance/core.py
eric-czech: Ok cool, can you bring that back to avoid redundant calculations?  I'm not aware of any non-symmetric genetic distance functions we've talked about so we can probably ignore that possibility for now.

sgkit/distance/metrics.py
aktech: If a row has a missing value in it, it will return `nan` for all the distances, that includes that row. This behaviour is consistent with `scipy`'s `pdist`:

```
In [4]: x = np.arange(16, dtype=float).reshape((4, 4))

In [5]: x[0][1] = np.nan

In [6]: x
Out[6]:
array([[ 0., nan,  2.,  3.],
       [ 4.,  5.,  6.,  7.],
       [ 8.,  9., 10., 11.],
       [12., 13., 14., 15.]])


In [7]: from scipy.spatial.distance import pdist, squareform

In [8]: squareform(pdist(x))
Out[8]:
array([[ 0., nan, nan, nan],
       [nan,  0.,  8., 16.],
       [nan,  8.,  0.,  8.],
       [nan, 16.,  8.,  0.]])

In [9]: from sgkit.distance.api import pdist

In [10]: pdist(x)
Out[10]:
array([[nan, nan, nan, nan],
       [nan,  0.,  8., 16.],
       [nan,  8.,  0.,  8.],
       [nan, 16.,  8.,  0.]])

```

aktech:
> Thanks @aktech, can you rerun that when they're computing the same outputs (i.e. only the upper triangle)?

Also to answer this question: https://github.com/pystatgen/sgkit/pull/306#discussion_r505518510

I was thinking the same, but the problem is, earlier we were running two custom loops, where we had the control to skip blocks from upper (or lower) triangular matrix, but now since, we're using `blockwise`, I haven't found a way to do that with `blockwise` yet.

>  Also, can you run it on a GPU or add an issue for doing so at another time?

I have been trying to do that using `dask_cuda` and `cupy` this morning, I haven't got it working yet. I was getting:
`TypeError: operand type(s) all returned NotImplemented from __array_ufunc__` with `cupy`. I haven't debuged this yet.

eric-czech:
> I was thinking the same, but the problem is, earlier we were running two custom loops, where we had the control to skip blocks from upper (or lower) triangular matrix, but now since, we're using blockwise, I haven't found a way to do that with blockwise yet.

I think that was actually redundant.  The code John was calling out it in https://github.com/pystatgen/sgkit/issues/241#issuecomment-702265516 looks like it replaces the definition of the chunks in the dask graph corresponding to the unwanted chunks with 0's (or calls np.triu if it splits multiple chunks).  It should still work.

sgkit/tests/test_distance.py
eric-czech: Oh also that float32 inputs result in float32 output arrays.

aktech: I think its probably worth having separate tests for each instead of parameters, so that when it fails we know, what exactly has failed?

aktech:
I updated the notebook after removing symmetric (repeated) calculations:
https://nbviewer.jupyter.org/gist/aktech/cc6518f66ac2bb02047ddbbbafdf9019

The results are much better now, even faster than prototype, with appropriate chunking.

aktech:
@eric-czech @alimanfoo I was thinking about handling of missing values this morning. For float input its quite straightforward, the missing values could be `nan`, but for `int` input its not that straightforward. There are two ways to handle it as per my understanding:

1. Convert everything to float and replace, let's say all `-1`'s with `nan`'s, but with this you lose the usability and speed of `int` overloaded `guvectorize`.

2. Take a user argument to the `pdist` function, such as let's say `missing_value=True`, this basically represents that the integer array has missing values which means the array should be converted to float and `-1` should be replaced with `nan`. The benefit of this is you don't convert to float everytime, even when its not required.

I think "2" is more suitable of the above two options. Also, I think since the logic of handling missing values is same for any distance metric, i.e. -1 will be missing value irrespective of distance metric used, it should be handled in the `pairwise` function itself and not in the individual distance metric functions.

eric-czech:
> this basically represents that the integer array has missing values which means the array should be converted to float and -1 should be replaced with nan. The benefit of this is you don't convert to float everytime, even when its not required.

Hm I'm not following on this -- how do you represent missing values in the int arrays without a float conversion then?

> Take a user argument to the pdist function, such as let's say missing_value=True, this basically represents that the integer array has missing values which means the array should be converted to float and -1 should be replaced with nan. The benefit of this is you don't convert to float everytime, even when its not required.

What happens when a float array is provided that happens to have negative values?  Is this trying to account for the space of all possible inputs or just non-negative ones?

--

As a general note operating on int arrays is a crucial feature IMO, otherwise for correlation and euclidean distance we might as well use standard matrix multiplication (both can be expressed that way) or da.corrcoef for the former.

eric-czech:
> The results are much better now, even faster than prototype, with appropriate chunking.

Very nice!

sgkit/distance/api.py
```diff
@@ -0,0 +1,63 @@
+import numpy as np
+
+from sgkit.distance import metrics
+from sgkit.distance.core import pairwise
+from sgkit.typing import ArrayLike
+
+
+def pdist(
```
eric-czech, line 8: Do we need this function?  Seems like this could be collapsed easily into `pairwise`.

Thinking about the name some more, appealing to `pdist` makes sense but it won't return squareform matrices, it will definitely have different dtype semantics, will possible handle missing values differently, will have different distance metrics, and I would guess that a lot of users won't be familiar with that in scipy anyhow.  

I'm leery of libraries that say they implement an API like that but with enough differences that it does more harm than good.  I think it would be safer to start with something generic like `pairwise` and rename to `pdist` later if we find a good reason to want greater parity with the scipy version.

aktech:
> Hm I'm not following on this -- how do you represent missing values in the int arrays without a float conversion then?

We need to define a way for the user to represent missing values, so for example either we can define that -1 in integer array and (-1 & nan) in float arrays represent missing values or alternatively the user can specify the missing values is represented by say -9932 or anything. The gist of what I am saying is whenever missing values are involved, we convert the arrays to float and also the representation of missing (like say -1, -9999, etc) values to nan.


> What happens when a float array is provided that happens to have negative values? Is this trying to account for the space of all possible inputs or just non-negative ones?

We convert those missing values (i.e. -1) to nan. My understanding is, the input of this function should be such that, it knows there are missing values in it and those should be replaced with nan after conversion to float.

> As a general note operating on int arrays is a crucial feature IMO, otherwise for correlation and euclidean distance we might as well use standard matrix multiplication (both can be expressed that way) or da.corrcoef for the former.

Yes, I am suggesting conversion to float only in the case of missing values, the reason being, its not possible to propogate the missing values in the final solution, without the missing value being `nan`. Otherwise everything else will be seen valid by distance metric functions. This would have been easier, if we had a nullable integer type.

sgkit/distance/api.py
aktech: That's a very valid point, would you consider the following renaming:
`pdist` -> `pairwise` a thin wrapper around `_pairwise`
`pairwise` -> `_pairwise`

or just scrap the latter completely?

eric-czech:
> Yes, I am suggesting conversion to float only in the case of missing values, the reason being, its not possible to propogate the missing values in the final solution, without the missing value being nan

Well, there will essentially always be some missing values.  So I'm understanding you correctly then, you are proposing that there is no way to compute on integer arrays with missing values?  

And why can nan's not be in the float results even if the functions operate on ints?  I'm expecting that we wouldn't rely on nan propagation in numpy, and instead use our own loops like Alistair did in the prototype and cython routines in scikit-allel (e.g. [rogers_huff_r](https://scikit-allel.readthedocs.io/en/stable/stats/ld.html?highlight=r2#allel.rogers_huff_r)).

I do like the idea of forcing everything to float and then treating this like a normal distance library with standard nan semantics, but a general purpose distance library isn't very useful here.  Neither of us (@alimanfoo or me) were aware of any reason to use this other than for comparing alt allele count vectors so I think we should tailor to that as much as possible.  

The alternative is to show that the memory and compute cost of promoting types to floats is irrelevant.  You could go that route if you'd like, but I think a demonstration of that much would be necessary to steer us away from what @alimanfoo already proposed or did in the past w/ scikit-allel.  

We have in general been preferring higher performance methods that implement less functionality at decision points like these.  I think it would be reasonable to say our distance metrics only work on alt allele counts.  This is also how Hail and PLINK work.

sgkit/distance/metrics.py
eric-czech: Hm I see.  That definitely won't be the behavior most users would want so the two ways around this I see are:

1. Force them to use some kind of imputation function first
2. Implement nan-aware metrics

I had been assuming 2 would make the most sense where missing values in either vector are ignored rather than leading to nan propagation.  Option 1 certainly seems to be a more popular choice in other domains but the missingness in genetics is typically low enough that nan-aware semantics become appealing too.  I could go either way on it though committing to operating on integers makes the nan-aware approach more appealing.  I had envisioned our implementations looking like this until having a good reason to change course: https://github.com/cggh/scikit-allel/blob/961254bd583572eed7f9bd01060e53a8648e620c/allel/opt/stats.pyx#L58 (how scikit-allel ignores missing values in pairwise LD calcs).

alimanfoo:
> We have in general been preferring higher performance methods that implement less functionality at decision points like these. I think it would be reasonable to say our distance metrics only work on alt allele counts. This is also how Hail and PLINK work fwiw.

FWIW this seems very reasonable to me. 

I.e., pretty much the only thing I need to be able to do to compute pairwise distance for arrays of alt allele counts per call, where dtype is always int and -1 is always missing.

sgkit/distance/api.py
aktech: I have removed `pdist` and merged it into `pairwise`.

sgkit/distance/metrics.py
aktech: I have incorporated this in the latest commit.

aktech:
@eric-czech @alimanfoo Thanks for enlightening me on this, that makes sense to me now. I have updated the metrics to handle missing values. Basically, negative and nan are considered as missing values as per the discussion above. I have also added tests for the same and for all the functionality so far.

Also, I had to get rid of `fastmath=True` param from `guvectorize` due to [this issue](https://github.com/numba/numba/issues/2919) in Numba.

sgkit/tests/test_distance.py
eric-czech: Pytest makes it obvious which parameter a failure is associated with so I wouldn't worry about it much.  All of the other "dtype preservation" tests use parameters as well so my weak preference would be for parameters.

aktech: Never mind, I have already [parameterised most of the them](https://github.com/pystatgen/sgkit/pull/306/files#diff-bfebf015a7d0a911d66743edcf9806a15b4661814667c0683fef0af8a6222b35R111), let me know if that doesn't makes sense.

eric-czech:
Thanks @aktech, this looks great!

What ever happened with the GPU experiments?  Can you drop some details on that into an issue if you weren't able to get it working?

> Also, I had to get rid of fastmath=True param from guvectorize due to this issue in Numba.

Good to know, but does it matter if all the checks in the guvectorize functions are for values > 0?

eric-czech (approved):

sgkit/distance/api.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Parameters
+    ----------
+    x
+        [array-like, shape: (M, N)]
+        An array like two dimensional matrix. The rows are the
+        vectors used for comparison, i.e. for pairwise distance.
+    metric
+        The distance metric to use. The distance function can be
+        'euclidean' or 'correlation'
```
eric-czech, line 49:
```suggestion
        'euclidean' or 'correlation'.
```

sgkit/distance/api.py
```diff
@@ -0,0 +1,102 @@
...
+
+    >>> import numpy as np
+    >>> x = np.array([[6, 4, 1,], [4, 5, 2], [9, 7, 3]])
+    >>> pairwise_distance(x, metric='euclidean')
+    array([[0.        , 2.44948974, 4.69041576],
+           [2.44948974, 0.        , 5.47722558],
+           [4.69041576, 5.47722558, 0.        ]])
+
+    >>> x = np.array([[6, 4, 1,], [4, 5, 2], [9, 7, 3]])
+    >>> pairwise_distance(x, metric='correlation')
```
eric-czech, line 78: It is odd for us to call this a distance function and then not invert correlation in some way.  It is a similarity measure here rather than a distance measure, so we should probably be subtracting it from 1 (e.g. https://en.wikipedia.org/wiki/Pearson_correlation_coefficient#Pearson's_distance).

aktech: Makes sense, I noticed this too, while comparing it with scipy's `pdist(x, metric="correlation")`, then it completely got off my head.

aktech:
> What ever happened with the GPU experiments? Can you drop some details on that into an issue if you weren't able to get it working?

I am planning to working on it this week, I gave a quick try with GPU on a borrowed machine last week and it didn't work straightaway throwing some NotImplementedError from cupy, probably to due guvectorize functions, I didn't put much time on it last week to conclude anything. I don't have the access to that machine anymore, so now I am creating one properly in the cloud on GCP to do proper testing, will share the results after that, meanwhile, I'll create an issue for the same.

> Good to know, but does it matter if all the checks in the guvectorize functions are for values > 0?

I does actually for filtering `nan` values, because while comparing `> 0`, it should know that `np.nan > 0` should return `False`, which it doesn't in the case of `fastmath=True`

alimanfoo:
For GPU experiments, have you tried Google colab? I think they give you an
Nvidia GPU for free.

aktech:
> For GPU experiments, have you tried Google colab? I think they give you an
Nvidia GPU for free.

I gave a quick try yesterday and setting up the environment was a bit non-trivial, like installing conda and adding stuff to path and installing system requirements. I didn't thought it will be easily usable, so didn't put much effort into it, now that you have suggested, I'll give it an another go.

eric-czech:
@alimanfoo did you want to take another look at this?  Otherwise, I'll set it to merge today.

sgkit/distance/api.py
```diff
@@ -0,0 +1,102 @@
...
+    array([[1.11022302e-16, 2.62956526e-01, 2.82353505e-03],
+           [2.62956526e-01, 0.00000000e+00, 2.14285714e-01],
+           [2.82353505e-03, 2.14285714e-01, 0.00000000e+00]])
+    """
+
+    try:
+        metric_ufunc = getattr(metrics, f"{metric}")
```
alimanfoo, line 85:
```suggestion
        metric_ufunc = getattr(metrics, metric)
```

sgkit/distance/api.py
```diff
@@ -0,0 +1,102 @@
...
+
+from sgkit.distance import metrics
+from sgkit.typing import ArrayLike
+
+
+def pairwise_distance(
+    x: ArrayLike,
+    metric: str = "euclidean",
+) -> np.ndarray:
+    """Calculates the pairwise distance between all pairs of vectors in the
```
alimanfoo, line 12:
```suggestion
    """Calculates the pairwise distance between all pairs of row vectors in the
```

Just to help make obvious this is computing distance between **rows**. (And thus when computing distance between samples, user will have to transpose input.)

sgkit/distance/api.py
```diff
@@ -0,0 +1,102 @@
...
+        The distance metric to use. The distance function can be
+        'euclidean' or 'correlation'.
+
+    Returns
+    -------
+
+    [array-like, shape: (M, N)]
+    A two dimensional distance matrix, which will be symmetric. The dimension
+    will be (M, N). The (i, j) position in the resulting array
+    (matrix) denotes the distance between ith and jth vectors.
```
alimanfoo, line 57:
```suggestion
    [array-like, shape: (M, M)]
    A two dimensional distance matrix, which will be symmetric. The dimension
    will be (M, M). The (i, j) position in the resulting array
    (matrix) denotes the distance between ith and jth row vectors in the input array.
```

alimanfoo (commented):
Some very minor suggestions, otherwise looks good. There really is some magic here! At some point I'd love to understand how this has eliminated the need for separate map and reduce steps.

alimanfoo:
Oops, didn't notice this was merged already. Hope review comments are helpful anyway, perhaps a small follow-up PR to make the small suggested clarifications would be good.

eric-czech:
> There really is some magic here! At some point I'd love to understand how this has eliminated the need for separate map and reduce steps.

FYI @alimanfoo this implementation, like matrix multiplication in dask, assumes whole rows of blocks fit in memory.  It isn't particularly scalable this way but is in greater alignment with how dask does similar things (e.g. `da.corrcoef`).  IMO this is ok since you can't really scale quadratic operations up to gigantic datasets anyhow.  Let me know if you disagree.

aktech:
@alimanfoo I'll fix these in a follow up PR.

Issue #282, opened 2020-09-25
Add `count_call_alternate_alleles` function


eric-czech:
I'm not so sure about that name but it seems like the most obvious choice given how our other counting functions are named.

This should do what https://scikit-allel.readthedocs.io/en/stable/model/ndarray.html#allel.Genotypes.to_n_alt does.

Likely tasks:

- [ ] Ensure that `count_call_alleles` has been run
- [ ] Use the above to sum non-reference allele counts (e.g. `ds.call_allele_count[:, 1:].sum(dim='alleles')`)
  - For handling missing data, it may be easier to add another numba gufunc or have an option that tells `call_allele_count` to count missing alleles too.  That's probably better than checking if the number of called alleles is less than some fixed ploidy value.
- [ ] Document the fact that we always assume the first allele is the reference allele in this function
- [ ] Decide what to do with missing data, my preference is:
  - Default to the behavior of `allel.Genotypes.to_n_alt(fill=-1)`, meaning that partially or completely missed calls result in a -1 count
  - Not even have an option to fill missing values with 0 -- I'm not sure why that's the default in scikit-allel.  @alimanfoo is there an important application for that?

timothymillar:
@eric-czech I think this could be calculated directly from genotype-calls similar to scikit-allel ([source](https://github.com/cggh/scikit-allel/blob/master/allel/model/ndarray.py#L872)) this should also work with #243.
There is a bit of discussion about identifying partial/missing genotypes in this issue #223.

Following the scikit-allel code an implementation could look something like:

```python
alternate_counts = (ds['genotype_calls'] > 0).sum(axis=-1)
partial = patial_genotype_calls(ds['genotype_calls'])
alternate_counts = xr.where(partial, fill, alternate_counts)
```

The function `patial_genotype_calls` (or a better name) should be aware of mixed-ploidy genotypes and only return `True` if a genotype contained a unknown allele `-1`.

eric-czech:
So I'm following, you're saying `patial_genotype_calls = lambda ds: (ds.call_genotype == -1).any(dim='ploidy')` right (ignoring non-alleles)?

I would have no need for it, but would you ever use a switch that allows for partial calls to be included in the alternate counts?  In other words, it would only return the fill value if all calls were missing rather than any one of them.

timothymillar:
[...]

Issue #243, opened 2020-09-10
Mixed ploidy genotype_call_dataset


timothymillar:
#### Overview
 
This is a proposal to support mixed-ploidy data within a single `genotype_call_dataset`. 
It aims to have minimal impact on existing code and to avoid causing issues when implementing diploid only functionality in future.
Criticism is very welcome!

Within the VCF specification, mixed-ploidy data can be defined by using a mixture of genotype string (GT) lengths.
For example the following encodes one diploid and two tetraploid individuals:
```
#CHROM    POS    ID    REF    ALT   QUAL    FILTER    INFO         FORMAT    SAMPLE1    SAMPLE2    SAMPLE3
1         1000   .     C      G     60      PASS      NS=3;DP=30   GT        0/0/0/0    0/0        0/0/0/1
1         2000   .     T      A     60      PASS      NS=3;DP=30   GT        0/0/0/0    0/1        0/1/./.
```

#### Implementation

I think that the simplest approach to supporting mixed ploidy in a `genotype_call_dataset` is to introduce a second sentinel value e.g. `-2` which indicates a *non-allele*.
The above VCF would translate to the array:
```python
call_genotype = da.array([
    [[0, 0, 0, 0], [0, 0, -2, -2], [0, 0, 0, 1]], 
    [[0, 0, 0, 0], [0, 1, -2, -2], [0, 1, -1, -1]]
])
```

A convenience mask could also be created, similar to the one indicating missing values:
```python
call_genotype_mask = call_genotype == -1
call_genotype_non_allele = call_genotype < -1
```

we can also include the ploidy of each genotype call:
```python
call_genotype_ploidy = (~call_genotype_non_allele).sum(axis=-1)  # shape (variants, samples)
```

and ploidy across each sample or variant using `-1` to indicate mixed/inconsistent ploidy:
```python
sample_ploidy_fixed = (call_genotype_ploidy[0,:] == call_genotype_ploidy).all(axis=0)
sample_ploidy = xr.where(sample_ploidy_fixed, call_genotype_ploidy[0,:], -1)  # shape (samples,)

variant_ploidy_fixed = (call_genotype_ploidy[:,0] == call_genotype_ploidy.T).all(axis=0)
variant_ploidy = xr.where(variant_ploidy_fixed, call_genotype_ploidy[:,0], -1)  # shape (variants,)
```

#### Checking if a dataset has consistent ploidy

[...]

Issue #223, opened 2020-09-01
Filtering partial genotype calls


timothymillar:
It's often necessary to filter out partial genotype calls in order to avoid bias in derived statistics such as allele frequencies.

For example (at a single locus) the set of 3 tetraploid genotype calls `0/0/0/1    0/1/./.    0/0/1/1`  would be filtered to `0/0/0/1    ./././.    0/0/1/1` and be treated as 2 observations for the purpose of calculating summary statistics.

In scikit-allel this can be achieved by setting the [`mask`](https://github.com/cggh/scikit-allel/blob/master/allel/model/ndarray.py#L228) attribute which indicates genotype calls to be excluded from further calculations (It's worth noting that in scikit-allel the `mask` is of shape `(variants, samples)` unlike the sgkit `mask` which is of shape `(variants, samples, ploidy)` and is used in a different context).

A consideration is that the current API design is to accumulate calculated statistics into a single DataSet object (issue #103). Therefore any statistics calculated prior to removing partial genotypes are likely to be inaccurate after the removal of partial genotypes.

One approach would be to create a "new" dataset that does not copy additional values from the original.

```python
def filter_partial_calls(ds: DataSet) -> DataSet:
    G = ds.call_genotype
    F = (G < 0).any(axis=-1)
    G1 = xr.where(F, -1, G)

    return sgkit.create_genotype_call_dataset(
        variant_contig_names=ds.contigs,
        variant_contig=ds.variant_contig,
        variant_position=ds.variant_position,
        variant_alleles=ds.variant_allele,
        sample_id=ds.sample_id,
        call_genotype=G1,
        call_genotype_phased=ds.call_genotype_phased,
        variant_id=ds.variant_id,
    )
```

This could also be broken into 2 more general functions

```python
def partial_calls(ds: Dataset, merge: bool = True) -> Dataset:
    # Returns ds with array of booleans called `call_is_partial` with shape (variant, sample)
    ...

def filter_calls(ds: DataSet, bools: str) -> DataSet:
    # Filters calls based on array `bools` with shape (variant, sample)
    ...
```

jeromekelleher:
Sounds good @timothymillar. I wonder how these specific considerations for filtering out different types of calls feed into the more general problem of grouping things. We started a discussion in [this thread](https://github.com/pystatgen/sgkit/pull/100#discussion_r467917194) about what such an xarray-idiomatic grouping API might look like. Is this related, or a side-issue do you think?

[...]

Issue #288, opened 2020-09-30
Improve warnings usage in docs examples


eric-czech:
See https://github.com/pystatgen/sgkit/pull/278#discussion_r497043147.

This may go away with https://github.com/pystatgen/sgkit/issues/287 if it will render warnings.

PR #308, opened 2020-10-06
Filter partial calls

timothymillar:
Issue #223 

Some notes:

- Adding a new top-level sub-module might be OTT but this doesn't really fit in `stats` or `model`.
- Currently this will always throw a merge warning which is a safe default but may be inconvenient. Is an argument to optionally silence merge warnings a good idea?
- This implementation will require some minor changes if/when #243 is resolved.
- Some discussion in issue #282 on handling partial genotypes in other functions.

### DIFF
--- a/sgkit/stats/preprocessing.py
+++ b/sgkit/stats/preprocessing.py
@@ -1,10 +1,15 @@
-from typing import Optional
+from typing import Hashable, Optional
 
 import dask.array as da
 import numpy as np
+import xarray as xr
 from sklearn.base import BaseEstimator, TransformerMixin
+from xarray import Dataset
+
+from sgkit import variables
 
 from ..typing import ArrayLike
+from ..utils import conditional_merge_datasets
 
 
 class PattersonScaler(BaseEstimator, TransformerMixin):  # type: ignore[misc]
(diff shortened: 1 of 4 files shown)


timothymillar:
Handling of warnings related to #288

sgkit/curate.py
```diff
@@ -0,0 +1,75 @@
...
+    Dataset containing updated variables `call_genotype` and
+    `call_genotype_mask` in which partial genotype calls are replaced
+    with compleately missing genotype calls.
+
+    Examples
+    --------
+
+    >>> import warnings
+    >>> import sgkit as sg
+    >>> from sgkit.testing import simulate_genotype_call_dataset
```
tomwhite, line 43: `simulate_genotype_call_dataset` is in the top-level namespace now, so you can just reference as `sg. simulate_genotype_call_dataset`

sgkit/curate.py
```diff
@@ -0,0 +1,75 @@
+from typing import Hashable
```
tomwhite, line 1: Note sure about the name of this file. Another possibility might be `preprocessing.py`?

tomwhite (commented):
> Currently this will always throw a merge warning which is a safe default but may be inconvenient. Is an argument to optionally silence merge warnings a good idea?

It does seem odd that this will always produce a merge warning, even though it's the expected behaviour. I think that will be confusing for users.

It's worth pointing out that the dataset that is passed in will not be updated, so the unfiltered calls are still available if needed. So this makes me think that the default behaviour here should probably *not* be to throw the warning, but I'm not sure how to make this consistent with other methods.

Thoughts @eric-czech, @jeromekelleher, @ravwojdyla?

eric-czech:
Hey @timothymillar, we were talking about this on a call today and we discussed a few options:

1. Making new variables for all functions that perform some kind of mutation (e.g. `call_genotype_complete[_mask]` here maybe)
2. Make an exception for mutation functions and have them only return new variables instead of merging
3. Disable the warning 
4. Rename the original variable as well as the new variable 

The consensus we landed on was option 1.  The advantages of this are:

- Multiple mutation functions relying on the same original variable can be run in sequence without the order mattering
- This can be more consistent with mutation functions that alter the interpretation of the variable in such a way that we would likely want to create a new sgkit.variable for the result anyhow, e.g. mean imputation.
- It doesn't alter the input variables as all other functions do

The biggest disadvantage is that downstream functions expecting to work on `call_genotype` need to be pointed at something else, which won't be obvious to the user.  For example:

```python
ds.pipe(sg.filter_partial_calls)\
    .pipe(sg.count_call_alleles, call_genotype=sg.variables.call_genotype_complete)
```

We aren't thrilled about that but it seems like the lesser of several evils.  How does that sound to you?

sgkit/curate.py
timothymillar: Agreed, `preprocessing` is better

timothymillar:
@eric-czech @tomwhite thanks for the feedback. 
I'm happy with option 1, it is the most consistent approach. Though it would require some clear documentation to ensure that users don't create `call_genotype_complete` and then use `call_genotype` without realizing.

timothymillar:
Sorry for the delay on this (busy week/new dev machine).

For the example code, it is also possible to drop and rename via xarray methods:
```python
ds2 = (
    sg.filter_partial_calls(ds)
    .drop(['call_genotype', 'call_genotype_mask'])
    .rename(call_genotype_complete='call_genotype', call_genotype_complete_mask='call_genotype_mask')
)
```
However my thought here is that simple overwriting syntax is probably more familiar to most users.

sgkit/variables.py
```diff
@@ -176,6 +176,20 @@ def _check_field(
 A flag for each call indicating if it is phased or not. If omitted
 all calls are unphased.
 """
+call_genotype_complete, call_genotype_complete_spec = SgkitVariables.register_variable(
+    ArrayLikeSpec("call_genotype_complete", kind="i", ndim=3)
+)
+"""
+Call genotypes in which partial genotype calls are replaced with
+compleately missing genotype calls.
```
eric-czech, line 184:
```suggestion
completely missing genotype calls.
```

sgkit/stats/preprocessing.py
```diff
@@ -102,3 +107,77 @@ def inverse_transform(self, X: ArrayLike, copy: Optional[bool] = None) -> ArrayL
...
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
+
+    Returns
+    -------
+    Dataset containing `call_genotype_complete` and
+    `call_genotype_complete_mask` in which partial genotype calls are
+    replaced with compleately missing genotype calls.
```
eric-czech, line 138:
```suggestion
    replaced with completely missing genotype calls.
```

eric-czech (approved):

eric-czech:
Would you like to take another look at this one @tomwhite?  I was going to set it to merge if not.

tomwhite (approved):

eric-czech:
Thanks @tomwhite, I'll set it to merge once those two tiny misspelling suggestions are committed @timothymillar.

Issue #56, opened 2020-07-21
VCF Zarr improvements


tomwhite:
It would be good to make VCF Zarr reading more flexible:

- [x] Support grouping by contig (https://github.com/pystatgen/sgkit/pull/40#issuecomment-658276008)
- [ ] Support case where `samples` is a group (https://github.com/pystatgen/sgkit/pull/40#issuecomment-658280113)
- [ ] Support reading from a `Mapping` object (not just a string or `Path`)
- [x] Support consolidated metadata

jerowe:
Hi there. I started on this today. 

@alimanfoo gave us some access to vfczarr data. I'm using that as the basis to read in data.

It looks like it the data is for all contigs you have:

```
2L
 ├── calldata
 │   └── GT (11524923, 1142, 2) int8
 ├── samples (1142,) object
 └── variants
     ├── ABHet (11524923,) float32
...

3L
 ├── calldata
 │   └── GT (11524923, 1142, 2) int8
 ├── samples (1142,) object
 └── variants
     ├── ABHet (11524923,) float32
```

Where `2L` and `3L` are the contig names, along with an additional `samples` key that is an array (that I think could optionally be a group).

It can also be the case that you have 1 zarr file per contig, in which case it looks like this:

```
 ├── calldata
 │   └── GT (11524923, 1142, 2) int8
 ├── samples (1142,) object
 └── variants
     ├── ABHet (11524923,) float32
```

So I broke this functionality down into 2 parts. The first is defining an Xarray dataset that we can use as a very minimal interface.

```
## Define a GenotypeCallSetDocument
  
from typing import Any, Dict, Hashable, List, Optional
import xarray as xr

def create_genotype_call_dataset_fetcher(
    *,
    allele_pos_key: str = 'variants/POS',
    allele_ref_key: str = 'variants/REF',
    allele_alt_key: str = 'variants/ALT',
    variant_id_key: str = 'variants/ID', 
    # potential alternate is 'calldata/genotype'
    genotype_call_key : str = 'calldata/GT',
    sample_key: str = 'samples',
    grouped_by_contig: bool = True,
    contigs: Optional[List[str]] = None
) -> xr.Dataset:
    """
    Define some keys to read in the zarrVCF
    
    If contigs is None we'll try to guess the contigs from the top level keys
    """
    
    attrs: Dict[Hashable, Any] = {
        "variant_id_key": variant_id_key,
        "allele_pos_key": allele_pos_key, 
[...]

Issue #298, opened 2020-10-05
Add save/load dataset methods to API


eric-czech:
In order to use sgkit in a pipeline now, we need methods that at least do these things:

```python
def save_dataset(ds, path, **kwargs):
    for v in ds:
        ds[v].encoding.pop('chunks', None)
    ds.chunk(chunks='auto').to_zarr(str(path), **kwargs)
        
def load_dataset(path):
    return xr.open_zarr(str(path), concat_characters=False)
```

This would address https://github.com/pystatgen/sgkit/issues/296, https://github.com/pystatgen/sgkit/issues/297, and https://github.com/pydata/xarray/issues/4452.

PR #324, opened 2020-10-13
Vcfzarr concatenate and rechunk

tomwhite:
This addresses parts of #56, and was motivated by the need to convert MalariaGEN data in VCF Zarr format to sgkit's representation.

The difficult case is where each contig is stored in a separate Zarr group, since the arrays need to be concatenated (along the variants dimension), and then uniformly rechunked so they can be stored in Zarr. Rechunker doesn't handle the case where Zarr arrays are concatenated (via Dask), since the input chunk sizes are not uniform (each Zarr array will have a final chunk that is less than the chunk size, so stacking them introduces non-uniformity).

Furthermore, using Dask to rechunk using the fix in https://github.com/dask/dask/issues/5105#issuecomment-625282112, doesn't work either since Dask caches input chunks that are split across output chunks (this was not the problem addressed in that issue).

This PR introduces a new `concatenate_and_rechunk` function that implements the solution outlined in the same issue: https://github.com/dask/dask/issues/5105#issuecomment-528430620. The idea is to create an empty output array and fill it in by slicing from the Zarr stores directly. This means that there is very low memory usage, and the whole thing is embarrassingly parallel. (The downside is that it operates at the Zarr level, not Xarray.)

The `concatenate_and_rechunk` function is used in a new `vcfzarr_to_zarr` function that converts (scikit-allel) VCF Zarr format to a sgkit's Zarr file.

This code will be useful for the native VCF support where we need to concatenate and rechunk multiple Zarr files, but I haven't integrated it there yet.


eric-czech:
fwiw I'd love to talk about this more on our call tomorrow:

> Furthermore, using Dask to rechunk using the fix in dask/dask#5105 (comment), doesn't work either since Dask caches input chunks that are split across output chunks (this was not the problem addressed in that issue).

I haven't found that to be a problem in concatenating datasets and rechunking them w/ Dask yet (it has worked at least), though it certainly worries me that it will be constantly necessary as a part of https://github.com/pystatgen/sgkit/issues/298.  I imagined doing this kind of rechunking with Dask would be very common in a QC workflow.

Did you start off using Dask to rechunk and have trouble or go right for a more optimal solution from the start?

tomwhite:
> Did you start off using Dask to rechunk and have trouble or go right for a more optimal solution from the start?

I started off using Dask to rechunk, but hit memory problems. I would rather have avoided writing this code, since ideally Dask rechunk would handle this case better, but it's probably worth keeping around until there's a more reliable way of doing it.

tomwhite:
I found the following thread which mentions the `block_size_limit` keyword to Dask `rechunk`, which might be helpful for controlling memory usage, although it's not obvious (to me) what to set it to in general:

https://stackoverflow.com/questions/55324848/for-a-bigger-than-memory-dask-array-of-size-m-n-how-to-re-chunk-from-chunks/55413100

tomwhite:
I opened an [issue](https://github.com/dask/dask/issues/6745) on the Dask project that dives into this problem in more detail.

tomwhite:
Discussion upstream (in https://github.com/dask/dask/issues/6745 and linked issues) has made it clear that there are no easy fixes for this problem. It's possible that Dask fixes and/or tuning will produce a solution, but in the meantime I think the pragmatic course of action is to merge the functionality in this PR.

I have refactored it slightly to make it possible for the caller to choose which concatenate and rechunk algorithm to use (the optimized one introduced in this PR is the default, the other one is the Xarray/Dask internal one) - this also makes it easier to remove the optimized one in the future if there is a Dask solution.

I have tested this code again on the MalariaGEN data. The notebook is [here](https://nbviewer.jupyter.org/github/tomwhite/shiny-train/blob/sgkit/notebooks/gwss/sgkit_import.ipynb).

jeromekelleher (approved):
LGTM. 

Do we have follow-up issues for all the unresolved points raised?

tomwhite:
Thanks @jeromekelleher. I've opened #365 and #366 to track outstanding work.

jeromekelleher:
OK, thanks. I'd say add the merge label if no one objects by tomorrow morning.

sgkit/io/vcfzarr_reader.py
```diff
@@ -42,35 +48,213 @@ def read_vcfzarr(path: PathType) -> xr.Dataset:
...
+
+def vcfzarr_to_zarr(
+    input: PathType,
+    output: PathType,
+    *,
+    contigs: Optional[List[str]] = None,
+    grouped_by_contig: bool = False,
+    consolidated: bool = False,
+    tempdir: Optional[PathType] = None,
+    concat_algorithm: Optional[str] = None,
```
ravwojdyla, line 63: you could probably use `Literal` type here to make it more user friendly and safe.

sgkit/io/vcfzarr_reader.py
```diff
@@ -42,35 +48,213 @@ def read_vcfzarr(path: PathType) -> xr.Dataset:
...
+                # Compute fixed-length string dtype for array
+                if kind == "O" or var in ("variant_id", "variant_allele"):
+                    kind = "S"
+                max_len = max_str_len(arr).values
+                dt = f"{kind}{max_len}"
+                ds[var] = arr.astype(dt)  # type: ignore[no-untyped-call]
+
+                if var == "variant_id":
+                    ds.attrs["max_variant_id_length"] = max_len
+                if var == "variant_allele":
```
ravwojdyla, line 204: nit: should be `elif`, or:

```
if var in {"variant_id", "variant_allele"}:
  ds.attrs[f"max_{var}_length"] = max_len
```

ravwojdyla (approved):

PR #329, opened 2020-10-15
Remove create_*_dataset methods from public API.

### LINKED ISSUE #252: Remove create_*_dataset methods from public API
As pointed out in https://github.com/pystatgen/sgkit/issues/243#issuecomment-691553120 these are not really public - they are used by the PLINK/BGEN/VCF reading functions, which is what users should use.

Also might be worth moving `simulate_genotype_call_dataset` to the top-level at the same time as it being in `testing` is an anomaly.

tomwhite:
Also add simulate_genotype_call_dataset to public API.

Fixes #252

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -34,17 +34,6 @@ VCF
 
    read_vcfzarr
 
-.. currentmodule:: sgkit
-
-Creating a dataset
-==================
-
-.. autosummary::
-   :toctree: generated/
-
-   create_genotype_call_dataset
-   create_genotype_dosage_dataset
-
 .. _api_methods:
 
 Methods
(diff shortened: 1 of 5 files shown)


eric-czech:
Should this be manually merged @tomwhite?  FWIW I had several PR's go nowhere today after adding the merge label so I've been merging them myself (cc @jeromekelleher).

tomwhite:
I think it needs to be approved first.

jeromekelleher (approved):

eric-czech:
Bah, right.  The ones I mentioned were approved though I only waited ~10 minutes or so.  I see that's about how long this one took so I'll try being more patient.

Issue #139, opened 2020-08-24
Simulated datasets don't serialize


eric-czech:
You can't currently save a dataset from `sgkit.testing.simulate_genotype_call_dataset` because the attributes contain numpy scalars which don't serialize as JSON when calling `ds.to_zarr`.  The simplest fix would be to have the function convert contig names to Python ints/string in https://github.com/pystatgen/sgkit/blob/master/sgkit/testing.py#L56.  A better fix would be to have `create_genotype_call_dataset` guard against this.

hammer:
Should this be closed or should it remain open until the "better fix" is implemented?

eric-czech:
I would say that my "better fix" isn't that important now that https://github.com/pystatgen/sgkit/issues/252 is on the way.  That, the to_zarr/from_zarr test in https://github.com/pystatgen/sgkit/pull/331, and the tests necessary for https://github.com/pystatgen/sgkit/issues/298 are all enough IMO to say that it should become very unlikely for a user to run into the issue I had in the future.

PR #331, opened 2020-10-16
Change simulated contig names to be python ints

eric-czech:
Closes https://github.com/pystatgen/sgkit/issues/139

### DIFF
--- a/sgkit/testing.py
+++ b/sgkit/testing.py
@@ -64,13 +64,13 @@ def simulate_genotype_call_dataset(
 
     contig_size = split_array_chunks(n_variant, n_contig)
     contig = np.repeat(np.arange(n_contig), contig_size)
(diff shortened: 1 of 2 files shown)


tomwhite (approved):
It took me a while to see what this changes - I didn't know that previously the names were of type `numpy.int64`.

Perhaps add an assert in a test to show they are python ints now?

tomwhite:
Test looks good. +1

eric-czech:
> Perhaps add an assert in a test to show they are python ints now?
>> Test looks good. +1

Thanks @tomwhite, I thought it made sense to test the save/load instead since it didn't seem otherwise very important for the contigs to be python ints.  Maybe it will catch other json-related serialization issues that sneak into our attributes too.

Issue #221, opened 2020-08-31
Add optional ploidy flag to HWE test function


eric-czech:
If you don't have an array with a ploidy dimension, the current function will throw an error.  It shouldn't require an array just to check that the ploidy in the genotype data is diploid and adding a dummy array as a workaround is not a good UX.

Issue #115, opened 2020-08-17
Have HWE function use genotype counting function


eric-czech:
See https://github.com/pystatgen/sgkit/pull/76#discussion_r468408000

PR #334, opened 2020-10-19
HWE function updates

eric-czech:
Closes https://github.com/pystatgen/sgkit/issues/221.
Closes https://github.com/pystatgen/sgkit/issues/115.

The most unusual part of this function is that it is looking for some notion of ploidy and allele count without explicitly needing data related to either for the calculation it runs (it only requires genotype counts).  This works by adding parameters to the function that only need to be set if the same values cannot be found in the dataset dimensions.  That seemed like a reasonable approach to maximize safety -- let me know if you have other suggestions @ravwojdyla / @tomwhite.

### DIFF
--- a/sgkit/stats/hwe.py
+++ b/sgkit/stats/hwe.py
@@ -8,6 +8,7 @@ from numpy import ndarray
 from xarray import Dataset
 
 from sgkit import variables
+from sgkit.stats.aggregation import genotype_count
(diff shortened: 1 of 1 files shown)


ravwojdyla (approved):

eric-czech:
Hm 2/3 times today that I've tried to use mergify nothing has happened after ~10 minutes.  Merging manually.

PR #349, opened 2020-10-26
Change count_variant_alleles to not reconstruct DataArray.

### LINKED ISSUE #305: Change count_variant_alleles to not reconstruct DataArray
I noticed that `count_variant_alleles` evaluated immediately when running it on a client to a dask cluster.  That may be because numba isn't installed on the workers.  If that was it, we should document this in a "troubleshooting" section of the docs.

tomwhite:
Test that all allele counting functions are lazy.

Note that `test_count_variant_alleles__chunked` fails without the change in `count_variant_alleles`. 

Fixes #305.

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -406,7 +406,7 @@ def Tajimas_D(
     S = ((ac > 0).sum(axis=1) > 1).sum()
 
     # assume number of chromosomes sampled is constant for all variants
-    n = ac.sum(axis=1).max()
+    n = ac.sum(axis=1).max().compute()
 
     # (n-1)th harmonic number
     a1 = (1 / da.arange(1, n)).sum()
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -208,12 +208,9 @@ def count_variant_alleles(
     """
     new_ds = Dataset(
         {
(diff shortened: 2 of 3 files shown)


sgkit/stats/popgen.py
```diff
@@ -406,7 +406,7 @@ def Tajimas_D(
     S = ((ac > 0).sum(axis=1) > 1).sum()
 
     # assume number of chromosomes sampled is constant for all variants
-    n = ac.sum(axis=1).max()
+    n = ac.sum(axis=1).max().compute()
```
tomwhite, line 409: I'm not sure if there's a way to make `Tajimas_D`, so this `compute` is needed for the time being at least. This was effectively being hidden by `variant_allele_count` not being lazy.

jeromekelleher (approved):
LGTM

jeromekelleher:
Doesn't look like anything controversial, feel free to add the merge label if you're happy with merging.

sgkit/stats/popgen.py
ravwojdyla: @tomwhite we should probably open an issue to remove this compute, and make it lazy again?

ravwojdyla: FYI https://github.com/pystatgen/sgkit/issues/354

Issue #347, opened 2020-10-22
Add documentation on how to persist computations in a workflow


eric-czech:
Because our library may define many precursor variables for any one calculation, it will become crucial for users to be able to persist/cache some of those variables.  This means that instead of letting the variable be defined automatically as a dask array, they would have to define it first themselves, call `.compute()` or `.persist()`, and then call the original function with that  precursor variable already defined.  There should be a section in the docs on this.

tomwhite:
One pattern that may be useful is to compute new variables in a new dataset (via `merge=False`), save only the new variables, then merge the datasets. This avoids having to re-persist the whole dataset to disk - only the new arrays are written.

```python
ds = ... # original dataset
cac = sg.count_cohort_alleles(ds, merge=False) # new variable
cac.to_zarr(...) # compute and persist to disk

# later
ds = xr.open_zarr(...) # reload from disk
cac = xr.open_zarr(...) # reload from disk
ds = xr.merge([ds, cac]) # merge into one dataset
```

I think this could be used in conjunction with #298.

tomwhite:
I've been trying out a checkpoint function that I think is a better way of doing what I was trying to achieve in my previous [comment](https://github.com/pystatgen/sgkit/issues/347#issuecomment-717400639).

```python
ds = ... # original dataset
ds = sg.count_cohort_alleles(ds) # new variable
with ProgressBar():
    # compute and persist to disk
    ds = sg.checkpoint_dataset(ds, path, [sg.variables.cohort_allele_count])
```

The way this works uses Xarray's ability to append a subset of variables to an existing Zarr store. This avoids having to re-write existing variables, and keeps the whole dataset together with no merging required. It's also possible to use the same function to overwrite existing variables. (The `load_dataset` and `save_dataset` functions are from #298/#392.)

```python
def checkpoint_dataset(
    ds: Dataset,
    path: PathType,
    data_vars: Optional[Sequence[Hashable]] = None
) -> Dataset:
    if data_vars is not None:
        ds = ds.drop_vars(set(ds.data_vars) - set(data_vars))
    save_dataset(ds, path, mode="a")
    return load_dataset(path)
```

tomwhite:
I used a conditional form of this pattern in a [notebook](https://nbviewer.jupyter.org/github/tomwhite/shiny-train/blob/sgkit/notebooks/gwss/sgkit_pbs.ipynb), so that it can be rerun without having to recompute variables:

```python
ds = ... # original dataset
[...]

PR #352, opened 2020-10-27
Standardize intermediate variables

tomwhite:
This addresses #286.

This is what I've implemented:

1. Methods declare only the variables that they use directly.
2. If the variable of the given name exists in the dataset it will be used for the computation.
3. If the variable doesn't exist then it will be computed if the variable name is the default, and added to the dataset.
4. An error is raised if the variable doesn't exist and it has a non-default name or it can't be computed (e.g. `call_genotype`).

These rules mean you can always call a method and it will compute (and retain) all intermediate variables:

```python
ds = pbs(ds)
# ds now has variables stat_divergence, stat_Fst, stat_pbs
```

To use non-default parameters for an intermediate step, you can do something like this:

```python
ds = Fst(ds, estimator="Nei")
# ds now has variables stat_divergence, stat_Fst

ds = pbs(ds) # uses existing stat_Fst
# ds now has variables stat_divergence, stat_Fst, stat_pbs
```

This approach should also play well with being able to persist intermediate steps that are computationally expensive (for #347):

```python
ds = count_cohort_alleles(ds)
ds.to_zarr(...) # compute and persist to disk

ds = xr.open_zarr(...) # reload from disk

ds = pbs(ds) # doesn't need to run any count alleles functions
```

(Note that `pbs` isn't done yet, so these examples won't work.)

This is based on #349, which should be merged first.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -43,6 +43,7 @@ Methods
    :toctree: generated/
 
    count_call_alleles
+   count_cohort_alleles
    count_variant_alleles
    divergence
    diversity
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -8,7 +8,12 @@ from .model import (  # noqa: F401
     create_genotype_call_dataset,
     create_genotype_dosage_dataset,
 )
-from .stats.aggregation import count_call_alleles, count_variant_alleles, variant_stats
+from .stats.aggregation import (
+    count_call_alleles,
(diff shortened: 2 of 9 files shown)


jeromekelleher (approved):
Nice - very elegant.

tomwhite:
Thank you @jeromekelleher! Would you mind taking a look at #349 (which this one depends on) when you have time.

Issue #354, opened 2020-10-27
Make Tajimas_D lazy again


ravwojdyla:
#349 and specifically caddf7c makes part of the computation eager, we should try to make it lazy again.

PR #355, opened 2020-10-27
Make Tajimas_D lazy again

ravwojdyla:
Re: #354


jeromekelleher (approved):
LGTM. 

Side note: Is there any way we can test if these computations are lazy/eager? Seems like we could easily slip in a ``.compute()`` somewhere in the pipeline which could lead to really tricky to diagnose perf problems.

tomwhite (approved):
@ravwojdyla thanks for fixing this (and I apologise for not opening an issue about it).

tomwhite:
> Is there any way we can test if these computations are lazy/eager?

Yes, we can test if dask array inputs produce dask array outputs, as in #349 for the count allele functions. I've opened #356 to track this.

Issue #357, opened 2020-10-28
Make sgkit pep-561 compliant


ravwojdyla:
Signal to mypy that sgkit is typed (essentially making it PEP-0561 compliant) by including `py.typed` in the top level package.

Without this sgkit users who use mypy will get this error:

```
test.py:1: error: Skipping analyzing 'sgkit': found module but no type hints or library stubs  [import]
test.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

PR #358, opened 2020-10-28
Follow pep-561

ravwojdyla:
Re: #357 
Doc: https://www.python.org/dev/peps/pep-0561/#packaging-type-information

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -43,6 +43,7 @@ Methods
    :toctree: generated/
 
    count_call_alleles
+   count_cohort_alleles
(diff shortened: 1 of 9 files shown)


jeromekelleher (approved):

tomwhite:
Do we need to add something to setup.py too (as mentioned in the linked doc)?

ravwojdyla:
@tomwhite our current setup already collects that extra file, so afaiu no need to add that extra config.

tomwhite (approved):

PR #364, opened 2020-11-02
Enable caching of compiled guvectorize functions

### LINKED ISSUE #363: Importing sgkit is slow
It takes ~5 seconds on my machine - most of this is numba compilation. I notice this a lot when running tests.

tomwhite:
Fixes #363

```
$ time python -c "import sgkit"
real	0m5.481s
user	0m5.960s
sys	0m0.527s
$ time python -c "import sgkit"
real	0m1.982s
user	0m2.520s
sys	0m0.471s
```

BTW @aktech I noticed that the `guvectorize` metric functions don't have `nopython=True` - do you think we should add it? (I could do it in this PR.)

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -121,9 +121,7 @@ def diversity(
 
 # c = cohorts, k = alleles
 @guvectorize(  # type: ignore
-    ["void(int64[:, :], float64[:,:])"],
-    "(c, k)->(c,c)",
-    nopython=True,
+    ["void(int64[:, :], float64[:,:])"], "(c, k)->(c,c)", nopython=True, cache=True
 )
 def _divergence(ac: ArrayLike, out: ArrayLike) -> None:
     """Generalized U-function for computing divergence.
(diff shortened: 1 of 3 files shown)


sgkit/stats/popgen.py
```diff
@@ -121,9 +121,7 @@ def diversity(
 
 # c = cohorts, k = alleles
 @guvectorize(  # type: ignore
```
jeromekelleher, line 123: Is there any good reason not to have ``cache=True`` for our use-case? Perhaps it's worth us having our own local ``guvectorize`` decorator that does this by default?

This is the sort of thing that's hard to test for and easy to forget, so maybe it's worth a little indirection.

jeromekelleher (approved):
LGTM, minor query

aktech:
> BTW @aktech I noticed that the guvectorize metric functions don't have nopython=True - do you think we should add it? (I could do it in this PR.)

That will be a good addition, I think.

aktech:
This is a really nice improvement, by the way. 👍

aktech (approved):

sgkit/stats/popgen.py
tomwhite: Maybe? It's not hard to add and now we have the convention started it is likely to be propagated in our code.

tomwhite: Opened #371

PR #373, opened 2020-11-05
[WIP] Mixed ploidy VCF IO #243

timothymillar:
Initial support for reading mixed ploidy data from VCF #243.

This approach requires a user to specify if their VCF should be treated as mixed ploidy or not using the `mixed_ploidy` argument.

- If `mixed_ploidy=False` then the value `-1` is used to fill in non-alleles (i.e. coercing lower ploidy calls to partial calls).
- If `mixed_ploidy=True` then the value `-2` is used to fill in non-alleles allowing downstream functionality to differentiate between calls of different ploidy.
- The `mixed_ploidy` argument is passed through to a `"mixed_ploidy"` attribute which is added to the `call_genotype` array to allow for fast distinction (without checking for `-2` values).
- If `mixed_ploidy=True` then an additional `call_genotype_non_allele` boolean array is added to the dataset indicating the position of `-2` values.
- A `ploidy` argument (default=2) is added to the vcf io functions to specify the ploidy dimension size (maximum ploidy).
- By default, an exception is raised if a genotype call is encountered that exceeds the size of `ploidy`.
- A `truncate_calls` argument is also added which, if true, will truncate genotype calls that exceed the size of `ploidy`.

Some questions/thoughts:

- Currently I have left `call_genotype_mask` to include anything < 0 (i.e. both -1 and -2 rather than just -1) though I'm not certain about the implications of this?
- I don't like the name `call_genotype_non_allele` but can think of a better one which has a clear meaning.
- In #243 I proposed automatically adding the following arrays to a new dataset but I now think that they should implemented as seperate methods in another PR:
    - `call_genotype_ploidy` with shape (variants, samples)
    - `sample_ploidy` with shape (samples,)
    - `variant_ploidy` with shape (variants,)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -105,8 +107,14 @@ def vcf_to_zarr_sequential(
                     max_variant_allele_length, max(len(x) for x in alleles)
                 )
 
-                gt = variant.genotype.array()
-                call_genotype[i] = gt[..., 0:-1]
+                fill = -2 if mixed_ploidy else -1
+                gt = variant.genotype.array(fill=fill)
+                gt_length = gt.shape[-1] - 1
```
tomwhite, line 112: I wasn't sure why 1 was being subtracted here at first. Maybe add a comment saying the last entry indicates if the genotype is phased. (I should have done so in the original code!)

sgkit/io/vcf/vcf_reader.py
```diff
@@ -225,6 +241,17 @@ def vcf_to_zarrs(
         Width (number of samples) to use when storing chunks in output, by default 1,000.
     output_storage_options
         Any additional parameters for the storage backend, for the output (see ``fsspec.open``).
+    ploidy : int
```
tomwhite, line 244: No need to add types here any more, since they are automatically generated. (Several of these below too.)

sgkit/io/vcf/vcf_reader.py
```diff
@@ -225,6 +241,17 @@ def vcf_to_zarrs(
...
+        The (maximum) ploidy of genotypes in the VCF file.
+    mixed_ploidy : bool
+        If true, genotype calls with fewer alleles than the specified ploidy will be padded
```
tomwhite, line 247: Nit: we've been using True to refer to the python value (not in monospaced font though). Not sure if there's a best practice to follow for that.

sgkit/model.py
```diff
@@ -52,6 +53,9 @@ def create_genotype_call_dataset(
     variant_id
         [array_like, element type: str or object, optional]
         The unique identifier of the variant.
+    mixed_ploidy
+        Specify if the dataset contatins genotype calls with a mixture of ploidy levels
```
tomwhite, line 57:
```suggestion
        Specify if the dataset contains genotype calls with a mixture of ploidy levels
```

sgkit/tests/io/vcf/test_vcf_reader.py
```diff
@@ -64,7 +64,7 @@ def test_vcf_to_zarr__small_vcf(shared_datadir, is_path, tmp_path):
             [[0, 0], [0, 0], [0, 0]],
             [[0, 1], [0, 2], [-1, -1]],
             [[0, 0], [0, 0], [-1, -1]],
-            [[0, -2], [0, 1], [0, 2]],  # -2 signifies non-allele
+            [[0, -1], [0, 1], [0, 2]],
```
tomwhite, line 67: Presumably because this is tested more extensively below now?

sgkit/variables.py
```diff
@@ -162,13 +162,20 @@ def _check_field(
...
     ArrayLikeSpec("call_genotype_mask", kind="b", ndim=3)
 )
 """TODO"""
+(
+    call_genotype_non_allele,
+    call_genotype_non_allele_spec,
+) = SgkitVariables.register_variable(
+    ArrayLikeSpec("call_genotype_non_allele", kind="b", ndim=3)
+)
+"""TODO"""
```
tomwhite, line 178: This should be filled in.

tomwhite (changes requested):
Thanks for submitting this @timothymillar - it looks great! I have made a few minor suggestions.

I think defining `call_genotype_mask` for anything less than zero is what we want generally. It would be worth checking that we don't explicitly check for -1 in other places (e.g. vectorized functions).

 > I don't like the name `call_genotype_non_allele`

Nothing jumps out at me I'm afraid. Anyone else?

+1 to adding new variables in separate PR(s).

sgkit/model.py
timothymillar: My spellcheck needs a spellcheck...

sgkit/tests/io/vcf/test_vcf_reader.py
timothymillar: Yes, I wanted to add a more extensive test for the mixed ploidy and polyploid (tetraploid) cases so I left this case as fixed ploidy. This resulted in the reversion of -2 to -1 due to this [change](https://github.com/pystatgen/sgkit/pull/373/files#diff-0eaa71fa8d1bd0b47476d65927c49ac5e40a1856df5d4c56719fcbced1a2766dR110) defaulting to -1 for the fill value.

timothymillar:
Thanks for the review @tomwhite. I'll make a new commit for now and can squash later.

tomwhite (approved):

Issue #231, opened 2020-09-02
Garud H Haplotype diversity statistic


jeromekelleher:
Add support for computing the Garud H haplotype diversity statistics based on the implementation in [scikit-allel](https://github.com/cggh/scikit-allel/blob/5f2f73ff926e105e5a65cacfc28f2928aba70f69/allel/stats/selection.py#L797). This can be tested using data generated by msprime and comparing the results to scikit-allel.

PR #377, opened 2020-11-10
Add to_haplotype_calls function

tomwhite:
This is needed for #231, which works on haplotypes.

I'm open to suggestions for a different name (`to_haplotype_calls`), or where it lives (`model.py` at the moment).

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -7,6 +7,7 @@ from .model import (
     DIM_VARIANT,
     create_genotype_call_dataset,
     create_genotype_dosage_dataset,
+    to_haplotype_calls,
 )
 from .stats.aggregation import (
     count_call_alleles,
@@ -48,6 +49,7 @@ __all__ = [
     "pbs",
     "pc_relate",
     "simulate_genotype_call_dataset",
+    "to_haplotype_calls",
     "variables",
     "pca",
     "window",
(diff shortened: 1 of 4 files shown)


jeromekelleher:
I'm not sure what we mean by "haplotypes" here. If we had a variant matrix V for haploids, where the rows are the genotype calls, then I'd call the columns of this matrix the "haplotypes". That is, each sample has a haplotype, which is its array of genotypes.

If we have diploid phased data, then by extension I'd say that we'd have two haplotype arrays for each sample. If the data is *unphased* then it's not clear what we mean by haplotypes, unless we arbitrarily phase them first.

tomwhite:
I think this PR does reflect my slightly fuzzy understanding of the term "haplotype", so thanks for your clarification @jeromekelleher, that makes a lot of sense.

I don't think we need this PR now, given the changes I've made in #378, so I'm closing it. We can always revisit if it is needed for another purpose.

hammer:
> my slightly fuzzy understanding of the term "haplotype"

Please don't feel bad about having a fuzzy understanding of this term @tomwhite! I'm 8 years in and remain confused about the many uses of "haplotype" in genetics.

jeromekelleher:
> Please don't feel bad about having a fuzzy understanding of this term @tomwhite! I'm 8 years in and remain confused about the many uses of "haplotype" in genetics.

Totally, I have no idea what people mean by "haplotype" half the time!

Issue #232, opened 2020-09-02
Genome-wide selection scans


jeromekelleher:
This is an umbrella issue covering a significant area of functionality, similar to #226 and #67.

- [x]  Windowing along the genome #229
- [x]  The PBS statistic #230
- [x]  The Garud H statistic #231

The goal here is to be able to run these selection scans at interactive speed on large datasets, for multiple population samples, and where each population sample has 100-1000 individuals. Note that this means we also need methods for grouping samples (#224).

hammer:
Some additional details on our discourse: https://discourse.pystatgen.org/t/genome-wide-selection-scans/90

alimanfoo:
Many thanks @jeromekelleher for raising this. Just to mention that I'm currently working on genome-wide selection scans in our mosquito data ([Ag1000G phase 2](https://github.com/pystatgen/sgkit/blob/0cb78b2f2df40ed44464139fbe849d572d6dbe9c/docs/public-datasets/ag1000g-phase2/README.md)), and thought it might be interesting to give a flavour of the data. I'm currently running scans using four methods, PBS, H12 (Garud H), iHS and XPEHH, on each of 14 populations. Because of >20 years of intense mosquito control efforts, there are some very strong signals, some of which are driven by genes we already know about, others we don't. Here's what the scans look like for one of these populations:

![gwss_bf_gam_gw_ug_gam_gq_gam](https://user-images.githubusercontent.com/703554/92145855-bd3cfe80-ee10-11ea-825a-b7f2a1e2341a.png)

Here's some plots zoomed in on a known insecticide resistance gene (Cyp6p3), which is under selection in multiple mosquito populations:

![locus_cyp6p3_h12](https://user-images.githubusercontent.com/703554/92145964-e9587f80-ee10-11ea-8662-dae1c2fd0031.png)

tomwhite:
With #368 we can now compute PBS for windows defined along the genome. I have started to port @alimanfoo's [PBS notebook](https://nbviewer.jupyter.org/github/alimanfoo/shiny-train/blob/master/notebooks/gwss/pbs_scans.ipynb) (using scikit-allel) to sgkit.

[...]

PR #378, opened 2020-11-10
Garud H statistics

### LINKED ISSUE #231: Garud H Haplotype diversity statistic

tomwhite:
This fixes #231.

* This PR depends on #377 for getting a haplotype representation of calls, so that should be merged first. The function there takes genotype calls of shape (variants, samples, ploidy) to (variants, haplotypes), where haplotypes=samples * ploidy.
* If the dataset is not windowed, the variants are assumed to be in a single window. This won't work for large numbers of variants (see discussion about hashing below), so we could choose not to support this and say that the input must be windowed. Would be good to hear thoughts about this @alimanfoo, @jeromekelleher.
* All of the H statistics work by computing statistics on the frequency of occurrences of each haplotype in a cohort. In scikit-allel, the haplotypes (columns of calls) are [hashed](https://github.com/cggh/scikit-allel/blob/5f2f73ff926e105e5a65cacfc28f2928aba70f69/allel/model/ndarray.py#L2507) by calling `hash(x.tobytes())` on the numpy array column `x`. When I tried this on MalariaGEN-scale data using Dask the computation ground to a halt, which I think was due to issues with the GIL. To avoid this, I have written a `hash_columns` function that is Numba-jit compiled, and outperforms the Python hash method by a factor of 5 in a single thread.
* I've used this code in a [notebook](https://nbviewer.jupyter.org/github/tomwhite/shiny-train/blob/sgkit/notebooks/gwss/sgkit_h12.ipynb) to compare with scikit-allel on MalariaGEN. The results for a single cohort and statistic (H12) are concordant (for the first 1000 windows at least), which is promising.
* The H stats notebook that uses scikit-allel has [different window sizes for different cohorts](https://github.com/alimanfoo/shiny-train/blob/master/notebooks/gwss/pop_defs.yml), which is not possible in sgkit easily at the moment (see https://github.com/pystatgen/sgkit/issues/232#issuecomment-722377847 for the same problem with PBS). The most pragmatic way to fix this will probably be to specify the subset of cohorts to calculate the statistic for.

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -1,3 +1,4 @@
+import collections
 from typing import Hashable, Optional
 
 import dask.array as da
@@ -5,9 +6,14 @@ import numpy as np
 from numba import guvectorize
 from xarray import Dataset
 
+from sgkit import to_haplotype_calls
 from sgkit.stats.utils import assert_array_shape
 from sgkit.typing import ArrayLike
-from sgkit.utils import conditional_merge_datasets, define_variable_if_absent
+from sgkit.utils import (
+    conditional_merge_datasets,
+    define_variable_if_absent,
+    hash_columns,
+)
 from sgkit.window import has_windows, window_statistic
 
 from .. import variables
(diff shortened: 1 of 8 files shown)


hammer:
> which I think was due to issues with the GIL.

cc @ravwojdyla who has debugged GIL contention issues previously.

ravwojdyla:
Just dropping some tools I find useful when I suspect GIL issues:
 * [py-spy](https://github.com/benfred/py-spy) which is good at ad-hoc, approximate profiling
 * [gil_load](https://github.com/chrisjbillington/gil_load) which is more accurate but also requires setup

alimanfoo:
> To avoid this, I have written a `hash_columns` function that is Numba-jit compiled, and outperforms the Python hash method by a factor of 5 in a single thread.

Amazing!!

alimanfoo:
> * If the dataset is not windowed, the variants are assumed to be in a single window. This won't work for large numbers of variants (see discussion about hashing below), so we could choose not to support this and say that the input must be windowed.

Yes, this only makes sense for windowed data, I think it would be fine to require windows.

alimanfoo:
Maybe there is a potential problem with hash collisions using dbx33a? https://gist.github.com/91f00a3d327ac07fb23be7cb2e332b4b

tomwhite:
> Maybe there is a potential problem with hash collisions using dbx33a?

It looks like the generator is creating duplicates, because it overflows a single-byte unsigned int. It works if you change `u1` to `u4`.

alimanfoo:
> > Maybe there is a potential problem with hash collisions using dbx33a?
> 
> It looks like the generator is creating duplicates, because it overflows a single-byte unsigned int. It works if you change `u1` to `u4`.

Beautiful, thanks, sorry for the noise.

sgkit/stats/popgen.py
```diff
@@ -682,3 +688,190 @@ def pbs(
         {variables.stat_pbs: (["windows", "cohorts_0", "cohorts_1", "cohorts_2"], p)}
     )
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+N_GARUD_H_STATS = 4  # H1, H12, H123, H2/H1
+
+
+def _Garud_h(k: ArrayLike) -> ArrayLike:
```
jeromekelleher, line 696: I'm having trouble understanding what ``k`` is here - any chance of a more descriptive name?

Unless k is quite small, isn't ``sorted(collections.Counter(k.tolist()).values(), reverse=True)`` going to be bottleneck?

jeromekelleher (commented):
Looks great @tomwhite.

Given that this is a function that could be applied to very large datasets, I wonder if the reshaping of the genotype data into haplotypes is necessary/worth it. Would we always want to double the memory footprint just to compute the haplotype hashes (which it looks to me is what we're doing)?

We don't have to answer this now, but I thought it was worthwhile asking this question before we add the to-haplotyes function over in #377

tomwhite:
Thanks for the review @jeromekelleher. You are right about `to_haplotype_calls` - I don't think we need it. Following your suggestion I have changed the `Garud_h` code to work directly on genotype calls. I did this by changing the hashing function to use `guvectorize` so it can cope with arbitrarily shaped arrays. The [updated notebook](https://nbviewer.jupyter.org/github/tomwhite/shiny-train/blob/sgkit/notebooks/gwss/sgkit_h12.ipynb) is a lot simpler now too (while producing the same output), since there's no need to create a haplotype representation.

I've also removed the non-windowed path following @alimanfoo's suggestion.

sgkit/stats/popgen.py
tomwhite: Changed to `haplotypes`.

I thought that the sorted call would be a bottleneck too, but I haven't seen that during my MalariaGEN testing.

jeromekelleher (approved):
Awesome! This is super cool, thanks @tomwhite !

PR #381, opened 2020-11-12
Fix coverage

### LINKED ISSUE #380: Coverage is less than 100%
The check that coverage must be 100% stopped working at some point. I'm not sure if it's related to the numba jit coverage issue (see #77), but I think we should mark numba functions as `no cover` (since there doesn't seem to be a way of getting coverage for these functions, see https://github.com/nedbat/coveragepy/issues/849).

We also need to reinstate the mechanism that fails the build if coverage is under 100%, and fix the places not being covered. I'll submit a patch to do all this.

tomwhite:
This is to fix #380, first by causing the build to fail (by adding the `--cov-fail-under` pytest option) - I will then add a commit to fix the remaining coverage issues.


eric-czech (approved):
LGTM

tomwhite:
As expected this fails with

```
FAIL Required test coverage of 100% not reached. Total coverage: 98.79%
```

tomwhite:
Thanks @eric-czech, do the [fixes](https://github.com/pystatgen/sgkit/pull/381/commits/6b38be390843836a8e5d0ada6c93fc13ce33a8f9) look OK too?

sgkit/io/vcfzarr_reader.py
```diff
@@ -169,7 +169,7 @@ def _vcfzarr_to_dataset(
     if "variants/ID" in vcfzarr:
         variants_id = da.from_zarr(vcfzarr["variants/ID"]).astype(str)
     else:
-        variants_id = None
```
eric-czech, line 172: Not worth testing this one?

eric-czech:
> Thanks @eric-czech, do the fixes look OK too?

Yep, the new tests look good to me.

jeromekelleher (approved):
LGTM

sgkit/io/vcfzarr_reader.py
tomwhite: Refactored to cover this case (see https://github.com/pystatgen/sgkit/pull/381/commits/be7f5beff41973ef69dd60d90bd2ef3130a891bd)

jeromekelleher:
Two approvals here @tomwhite, so merge as you see fit!

Issue #367, opened 2020-11-02
Release 0.1.0a1


tomwhite:
This is an umbrella issue for our first release.

This would be an alpha [pre-release](https://www.python.org/dev/peps/pep-0440/#pre-releases) (suggested version number: 0.1.0a1), which means that users would have to opt in to use it by specifying the `--pre` flag: `pip install --pre sgkit`.

Here are some of the tasks we need to do - please suggest more (or say if they are not needed). 

- [ ] cyvcf2 wheel (should we wait for this?)
- [ ] Build release artifacts locally
- [ ] Add release number to documentation URL
- [ ] Use TestPyPI to test release process
- [ ] Update installation instructions for users
- [ ] Document release process

Note that these are all about what we need to do to make a release; since it's a pre-release, the exact set of features doesn't matter so much.

hammer:
@jeromekelleher notes that `tskit` makes use of GH Actions to automate testing of installation of releases. @jeromekelleher can you point us to that?

jeromekelleher:
Yep, [here](https://github.com/tskit-dev/tskit/blob/main/.github/workflows/wheels.yml) it is. This action builds the release artefacts on each pull request, does some basic tests on them to make sure they work, and uploads them to PyPI when a release is tagged.

jeromekelleher:
Should be quite a bit simpler here since we're not actually compiling anything.

horta:
I'm taking care of the cyvcf2 wheels. It is on test-pypi already. I will open a PR for Brent today. Meanwhile you can test it as follows.


Installing cyvcf2
-----------------

```
# coloredlogs is not on test-pypi
pip install coloredlogs
```

```
# numpy wheels are on pypi
pip install numpy
```

```
# Our testing distribution
pip install -i https://test.pypi.org/simple/ cyvcf2
```

The following should then work:

`cyvcf2 --help`

Platforms: MacOS, Linux
Python: 3.5, 3.6, 3.7., 3.8, 3.9

Related issues
--------------

https://github.com/pystatgen/sgkit/issues/327
https://github.com/pystatgen/sgkit/issues/367

horta:
I've just opened a PR: https://github.com/brentp/cyvcf2/pull/175

hammer:
[...]

PR #387, opened 2020-11-17
Post 0.1.0a1 release doc fixes

tomwhite:
This moves the required Python versions to a more prominent position following https://github.com/pystatgen/sgkit/issues/367#issuecomment-728056891.

I have also removed the `--pre` option from the `pip install` commands. I misunderstood the meaning of this flag - I assumed that you had to include it to install *any* pre-release version, but that's not quite correct - if there are *only* pre-release versions available then pip will install the latest one of those, but if a stable version is available pip will favour that, even if it is older. This is discussed a bit more here: https://github.com/pypa/pip/issues/5503

So for example, at the moment users will get version 0.1.0a1 (whether they use `--pre` or not), but once we have released 0.1.0 and 0.2.0a1 (say), then users will get 0.1.0, unless they use `--pre` in which case they will get 0.2.0a1.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -121,12 +121,6 @@ Contributing to code
 *sgkit* maintains development standards that are similar to most PyData projects.  These standards include
 language support, testing, documentation, and style.
 
-Python versions
-~~~~~~~~~~~~~~~
-
-*sgkit* supports Python versions 3.7 and 3.8.
-
-
 Continuous Integration
 ~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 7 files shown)


jeromekelleher:
Nice clarification on ``--pre``, I hadn't realised that either.

One thing I was wondering about - is there a notation for install *all* the extras - would be good to document if there is.

jeromekelleher (approved):

tomwhite:
> is there a notation for install _all_ the extras

I think the convention is to have an `all` extra that just depends on the others. We could add that in a separate issue.

PR #389, opened 2020-11-17
Make window step default to window size

tomwhite:
A simplification that addresses this comment: https://github.com/pystatgen/sgkit/pull/303#discussion_r508392377


jeromekelleher (approved):
LGTM, and it does simplify code quite nicely.

We should have an explicit test that the default does this, though?

PR #392, opened 2020-11-19
Add load_dataset and save_dataset functions

### LINKED ISSUE #298: Add save/load dataset methods to API

tomwhite:
This fixes #298, and also incorporates some of the implementation from https://github.com/related-sciences/ukb-gwas-pipeline-nealelab/blob/4f862e31b8093d25fdaa8da7f841b9be8583cda4/scripts/gwas.py#L53-L71

Unlike #298 it does *not* rechunk the dataset before saving. I worry about having implicit rechunk methods that the user can't control since we have seen them perform poorly in some cases. For the moment I think it's preferable for the user to explicitly rechunk before saving. This is what I have been doing in the MalariaGEN notebooks, and Eric has too judging by [this example](https://github.com/related-sciences/ukb-gwas-pipeline-nealelab/blob/4f862e31b8093d25fdaa8da7f841b9be8583cda4/scripts/gwas.py#L228-L230).

I also haven't added `fsspec` support even though the GWAS pipeline uses it. This is because I was getting a warning that files were not being closed when running on local files. That could be addressed separately.

### DIFF
--- a/sgkit/io/dataset.py
+++ b/sgkit/io/dataset.py
@@ -1,0 +1,53 @@
+from typing import Any
+
+import xarray as xr
+from xarray import Dataset
+
+from sgkit.typing import PathType
+
+
+def save_dataset(ds: Dataset, path: PathType, **kwargs: Any) -> None:
+    """Save a dataset to Zarr storage.
+
+    This function is a thin wrapper around :meth:`xarray.Dataset.to_zarr`
+    that uses sensible defaults and makes it easier to use in a pipeline.
+
+    Parameters
+    ----------
+    ds
+        Dataset to save.
+    path
+        Path to directory in file system to save to.
+    kwargs
+        Additional arguments to pass to :meth:`xarray.Dataset.to_zarr`.
+    """
+    store = str(path)
+    for v in ds:
+        # Workaround for https://github.com/pydata/xarray/issues/4380
+        ds[v].encoding.pop("chunks", None)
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):

sgkit/io/dataset.py
```diff
@@ -0,0 +1,52 @@
...
+    ds
+        Dataset to save.
+    path
+        Path to directory in file system to save to.
+    kwargs
+        Additional arguments to pass to :meth:`xarray.Dataset.to_zarr`.
+    """
+    store = str(path)
+    for v in ds:
+        ds[v].encoding.pop("chunks", None)
```
eric-czech, line 26: Worth mentioning https://github.com/pydata/xarray/issues/4380 in a comment?

eric-czech (approved):

PR #401, opened 2020-11-23
Ignore samples that are not in any cohort

### LINKED ISSUE #400: Ignore samples that are not in any cohort
Sometimes there are samples that we don't want to define a cohort for (this happens in the MalariaGEN analyses for example) - they should be skipped.

tomwhite:
Fixes #400


jeromekelleher (approved):
LGTM

PR #403, opened 2020-11-23
Divergence should handle case when all calls for a cohort are missing

### LINKED ISSUE #402: Divergence should handle case when all calls for a cohort are missing
This case currently fails with

```
E           RuntimeWarning: invalid value encountered in _divergence
```

tomwhite:
Fixes #402


jeromekelleher (approved):
LGTM

Issue #341, opened 2020-10-21
Flexible specification of windows


tomwhite:
As discussed in https://github.com/pystatgen/sgkit/pull/303#discussion_r508489478

jeromekelleher:
See also #229

tomwhite:
I'm wondering about having separate windowing functions for windowing by index, position in genome, etc. So rather than have a single `window` function that handles all options, we could have multiple functions that only take the arguments that apply to that function.

For example, to window by index (number of variants - the current `window` function), the function might be:

```python
def window_by_index(
    ds: Dataset,
    *,
    size: int,
    step: Optional[int] = None,
    variant_contig: Hashable = variables.variant_contig,
    merge: bool = True,
) -> Dataset:
  pass
```

(Note that we use the `variant_contig` variable to ensure that windows don't cross contigs.)

For genomic position, we don't need `step`, but we do need the `variant_position` variable. Also, `size` now measures number of base pairs (rather than variant index).

```python
def window_by_position(
    ds: Dataset,
    *,
    size: int,
    variant_contig: Hashable = variables.variant_contig,
    variant_position: Hashable = variables.variant_position,
    merge: bool = True,
) -> Dataset:
  pass
```

In the future we might add `window_by_bed_file`, which would take a path; we might also want `window_by_accessible_bases` like scikit-allel's [`equally_accessible_windows`](https://scikit-allel.readthedocs.io/en/stable/stats/window.html#allel.equally_accessible_windows) (although that one could conceivably be a part of `window_by_index` by adding an optional variable to indicate if a base is accessible).

Any thoughts?

eric-czech:
👍 

Makes the interpretation of the parameters much more clear and extends more naturally.

hammer:
I also like the separate functions

jeromekelleher:
Good idea, I like it. Makes the semantics much clearer.

PR #404, opened 2020-11-24
Add docs for windowing and cohorts

tomwhite:
This covers the work done for #229 and #224 and related PRs.

We could certainly improve the user API for cohorts (and there are also improvements for windowing in #341), but this documents what we currently have at least.


jeromekelleher (approved):
Looks great, really clear!

I still haven't gotten my head around how xarray assign_coords works, but I think that's just me needing to do some work with it.

PR #410, opened 2020-11-30
Enable strict checking in Sphinx build #86

### LINKED ISSUE #86: Enable strict checking in Sphinx build
This is "nit-picky mode" (`-n`). It's possible to ignore false positives by setting [nitpick_ignore](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpick_ignore).

tomwhite:
Enables nit picky mode when building Sphinx docs, which picks up some errors that were getting through before.

Fixes #86

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Check for Sphinx doc warnings
       run: |
         cd docs
-        make html SPHINXOPTS="-W --keep-going"
+        make html SPHINXOPTS="-W --keep-going -n"
     - name: Test with pytest and coverage
       run: |
         pytest -v --cov=sgkit --cov-report=term-missing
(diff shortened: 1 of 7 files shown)


sgkit/display.py
```diff
@@ -129,7 +129,7 @@ def truncate(ds: xr.Dataset, max_sizes: Mapping[Hashable, int]) -> xr.Dataset:
 
 def display_genotypes(
     ds: xr.Dataset, max_variants: int = 60, max_samples: int = 10
-) -> GenotypeDisplay:
+) -> object:
```
tomwhite, line 132: I made this change because `GenotypeDisplay` isn't public (and probably doesn't need to be), and Sphinx was complaining about that. There may be a better way of dealing with this though.

jeromekelleher (approved):

sgkit/display.py
tomwhite: Nevermind, I changed this back as I found a way of getting Sphinx to ignore this (while also keeping mypy happy).

PR #412, opened 2020-12-01
Add target_part_size to vcf_to_zarr

### LINKED ISSUE #313: Support automatic partitioning in vcf_to_zarr
Currently you have to call `partition_into_regions` then pass the resulting regions to `vcf_to_zarr`. It would be nice for `vcf_to_zarr` to handle partitioning automatically so that users don't have to use `partition_into_regions`.

tomwhite:
Fixes #313

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -11,6 +11,7 @@ import xarray as xr
 from cyvcf2 import VCF, Variant
 
 from sgkit.io.utils import zarrs_to_dataset
+from sgkit.io.vcf import partition_into_regions
 from sgkit.io.vcf.utils import build_url, chunks, temporary_directory, url_filename
 from sgkit.model import DIM_VARIANT, create_genotype_call_dataset
 from sgkit.typing import PathType
@@ -311,6 +312,7 @@ def vcf_to_zarr(
     input: Union[PathType, Sequence[PathType]],
     output: Union[PathType, MutableMapping[str, bytes]],
     *,
+    target_part_size: Optional[int] = None,
     regions: Union[None, Sequence[str], Sequence[Optional[Sequence[str]]]] = None,
     chunk_length: int = 10_000,
     chunk_width: int = 1_000,
@@ -323,12 +325,13 @@ def vcf_to_zarr(
 ) -> None:
     """Convert VCF files to a single Zarr on-disk store.
 
-    For a single input and a single region, the conversion is carried out sequentially.
+    For a single input and a single region, the conversion is carried out sequentially,
+    unless `target_part_size` is specified.
 
-    For multiple outputs or regions, the conversion is carried out in parallel, by writing
-    the output for each region to a separate, intermediate Zarr store in ``tempdir``. Then,
-    in a second step the intermediate outputs are concatenated and rechunked into the final
-    output Zarr store in ``output``.
+    For multiple outputs or regions, or if `target_part_size` is specified, the conversion
+    is carried out in parallel, by writing the output for each part to a separate,
+    intermediate Zarr store in ``tempdir``. Then, in a second step the intermediate outputs
+    are concatenated and rechunked into the final output Zarr store in ``output``.
 
     For more control over these two steps, consider using :func:`vcf_to_zarrs` followed by
     :func:`zarrs_to_dataset`, then saving the dataset using Xarray's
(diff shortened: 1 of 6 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -322,12 +324,13 @@ def vcf_to_zarr(
 ) -> None:
     """Convert VCF files to a single Zarr on-disk store.
 
-    For a single input and a single region, the conversion is carried out sequentially.
+    For a single input and a single region, the conversion is carried out sequentially,
```
eric-czech, line 327: This seems a little confusing since regions and target_part_size can't be specified together.  Perhaps: 

> Conversion is carried out sequentially unless multiple inputs, a target partition size, or multiple regions within at least one input are specified.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -322,12 +324,13 @@ def vcf_to_zarr(
...
+    unless `target_part_size` is specified.
 
-    For multiple outputs or regions, the conversion is carried out in parallel, by writing
-    the output for each region to a separate, intermediate Zarr store in ``tempdir``. Then,
-    in a second step the intermediate outputs are concatenated and rechunked into the final
-    output Zarr store in ``output``.
+    For multiple outputs or regions, or if `target_part_size` is specified, the conversion
```
eric-czech, line 330: How are the results sorted?  Are there any guarantees for that or ways to ensure that the contig + genomic position  sort order is preserved?

sgkit/io/vcf/vcf_reader.py
```diff
@@ -341,6 +344,11 @@ def vcf_to_zarr(
         have a ``.csi`` index file.
     output
         Zarr store or path to directory in file system.
+    target_part_size
```
eric-czech, line 347: Should this have a non-None default value?  It reminds me a bit of the per-worker memory limit used in rechunker that I didn't make into a required parameter.  Anyways, I think we could be opinionated on it if you have a feeling for a what a good setting is and since other target chunk settings have defaults.  Is there any other reason a user might prefer synchronous processing?

eric-czech (approved):
Looks good, had a few minor questions/comments.

sgkit/io/vcf/vcf_reader.py
tomwhite: This is a good question and something worth documenting. Within each VCF file, order is preserved, since regions are numbered sequentially, and concatenated in sequence order. For multiple files, the order is the order that the files are specified in (by the user), and usually corresponds to the order that the contigs are arranged in.

eric-czech: I see, nice.  Worth adding a test for it?

sgkit/io/vcf/vcf_reader.py
tomwhite: Parallel processing by default would probably be a good idea. But I don't have a great feeling for what a good setting is yet. I've used 1GB for processing the Ag1000G data and each part took around 5 minutes to process, which seems OK for larger (~100GB) VCF files. OTOH for more modest-sized files it would still be nice to have some parallelism, so 100MB might be a better default.

sgkit/io/vcf/vcf_reader.py
tomwhite: I was just thinking that too 😄

tomwhite: I've opened #425 to address this

sgkit/io/vcf/vcf_reader.py
tomwhite: I've changed the default to 100MB, and updated the documentation to reflect the change.

jeromekelleher: I'd argue it's better to not encode the default in the signature though - we can easily imagine a future where we try to find a good value, or the best value depends on the inputs. Using a local sentinel via ``DEFAULT_TARGET_PART_SIZE=object()`` would allow us to detect when the user doesn't specify a value as well as supporting the None semantics here.

tomwhite: +1 Perhaps we could have a string value like `"auto"` - then we could also allow human-readable byte sizes like `"1GB"` etc.

tomwhite: Latest commit supports human-readable strings and `"auto"` is the default, see https://github.com/pystatgen/sgkit/pull/412/commits/502511bef08fed037a7a8c681e54ad5a9cec8024

PR #415, opened 2020-12-02
Mixed-ploidy filter_partial_calls #414

### LINKED ISSUE #414: Filter mixed-ploidy partial calls
Currently `filter_partial_calls` treats any value less than 0 as a missing call resulting in the removal of complete calls of lower ploidy (e.g. `[0,1,-2,-2]`). In the case of mixed-ploidy call_genotypes, only -1 should be treated as a missing allele. In the case of a partial lower ploidy call (e.g. `[0,-1,-2,-2]`) the non-allele values should be retained in the filtered call (e.g. `[-1,-1,-2,-2]`).

timothymillar:
Fixes #414

### DIFF
--- a/sgkit/stats/preprocessing.py
+++ b/sgkit/stats/preprocessing.py
@@ -172,7 +172,11 @@ def filter_partial_calls(
     """
     variables.validate(ds, {call_genotype: variables.call_genotype_spec})
     G = ds[call_genotype]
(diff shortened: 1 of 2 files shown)


tomwhite (approved):
LGTM

eric-czech (approved):
LGTM too

PR #417, opened 2020-12-03
Sample summary stats #29

### LINKED ISSUE #29: Add variant/sample summary statistic methods

tomwhite:
Fixes #29

### DIFF
--- a/sgkit/variables.py
+++ b/sgkit/variables.py
@@ -263,10 +263,34 @@ pc_relate_phi, pc_relate_phi_spec = SgkitVariables.register_variable(
     ArrayLikeSpec("pc_relate_phi", ndim=2, kind="f")
 )
 """PC Relate kinship coefficient matrix."""
+sample_call_rate, sample_call_rate_spec = SgkitVariables.register_variable(
+    ArrayLikeSpec("sample_call_rate", ndim=1, kind="f")
+)
+"""The fraction of variants with called genotypes."""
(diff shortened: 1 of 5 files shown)


sgkit/variables.py
```diff
@@ -263,10 +263,34 @@ def _check_field(
     ArrayLikeSpec("pc_relate_phi", ndim=2, kind="f")
 )
 """PC Relate kinship coefficient matrix."""
+sample_call_rate, sample_call_rate_spec = SgkitVariables.register_variable(
+    ArrayLikeSpec("sample_call_rate", ndim=1, kind="f")
+)
+"""The number of variants with heterozygous calls."""
```
eric-czech, line 269: This comment is for a different variable.

eric-czech (approved):
LGTM

PR #420, opened 2020-12-14
Add docs for variables to xarray attributes #393

### LINKED ISSUE #393: Add docs for variables to xarray attributes
It would be nice if the doc strings for sgkit variables appeared in the xarray variable metadata. Users would be able to inspect this information when looking at datasets in notebooks. As an example, have a look at this [Pangeo dataset](https://catalog.pangeo.io/browse/master/ocean/sea_surface_height/) - try clicking on the document icon for a data variable (e.g. `adt`).

pangeo

Unfortunately, I can't see an easy way of doing this, since the `__doc__` attribute for sgkit variables is for the `ArrayLikeSpec` class, not for the variable instance itself. This is regular Python semantics, and the reason that the docs have the correct descriptions is because Sphinx is [reading them itself](https://www.sphinx-doc.org/en/1.4.8/ext/autodoc.html#directive-autoattribute).

@ravwojdyla any thoughts on how to do this?

tomwhite:
This is a draft to show how it might work. I haven't converted all the docstrings yet, but will do if this looks OK.

Here is an example rendered dataset - note the `comment` attributes (they are hidden by default and can be viewed by clicking on the file icon):

ds

I also checked that the Sphinx generated documentation is the same (note use of `dataclasses.field(repr=False)` so that the docstring does not show up as an attribute in the string representation of `ArrayLikeSpec`).

Fixes #393

### DIFF
--- a/sgkit/utils.py
+++ b/sgkit/utils.py
@@ -1,10 +1,11 @@
 import warnings
-from typing import Any, Callable, Hashable, List, Optional, Set, Tuple, Union
+from typing import Any, Callable, Hashable, List, Mapping, Optional, Set, Tuple, Union
 
 import numpy as np
 from numba import guvectorize
 from xarray import Dataset
 
+from . import variables
 from .typing import ArrayLike, DType
 
 
@@ -192,6 +193,36 @@ def define_variable_if_absent(
             f"Variable '{variable_name}' with non-default name is missing and will not be automatically defined."
         )
     return func(ds)
+
+
+def create_dataset(
+    data_vars: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+    coords: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+    attrs: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+) -> Dataset:
+    """Create an Xarray dataset and validate its variables.
+
+    This is a wrapper around `xarray.Dataset`, with the additional
+    convenience of validating variables against the ones defined by sgkit,
+    and annotating these variables with a `comment` attribute containing
+    their doc comments.
+
+    Parameters
+    ----------
+    data_vars
+        A mapping defining data variables.
+    coords
+        A mapping defining coordinates.
+    attrs
+        Global attributes.
+
+    Returns
+    -------
+    A new dataset.
+    """
+    ds = Dataset(data_vars, coords, attrs)
+    ds = variables.annotate(ds)
+    return ds
 
 
 def split_array_chunks(n: int, blocks: int) -> Tuple[int, ...]:
--- a/sgkit/variables.py
+++ b/sgkit/variables.py
@@ -1,6 +1,7 @@
+import dataclasses
 import logging
 from dataclasses import dataclass
-from typing import Dict, Hashable, Mapping, Set, Tuple, Union, overload
+from typing import Dict, Hashable, Mapping, Optional, Set, Tuple, Union, overload
 
 import xarray as xr
 
@@ -12,6 +13,7 @@ class Spec:
     """Root type Spec"""
 
     default_name: str
+    __doc__: Optional[str] = dataclasses.field(default=None, repr=False)
 
     # Note: we want to prevent dev/users from mistakenly
     #       using Spec as a hashable obj in dict, xr.Dataset
@@ -110,6 +112,8 @@ class SgkitVariables:
             check_array_like(
                 xr_dataset[field], kind=field_spec.kind, ndim=field_spec.ndim
             )
+            if field_spec.__doc__ is not None:
+                xr_dataset[field].attrs["comment"] = field_spec.__doc__.strip()
         except (TypeError, ValueError) as e:
             raise ValueError(
                 f"{field} does not match the spec, see the error above for more detail"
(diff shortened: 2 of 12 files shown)


tomwhite:
Test failures are due to #421

jeromekelleher:
Fancy! LGTM

ravwojdyla (approved):
@tomwhite great, I'm happy this has worked without too much resistance. The only real comment I have is the naming right now. There are two methods:
 * `_check_field` - which used to just check field, now also optionally adds doc
 * `validate` - which used to just "validate" dataset, now might also add doc

I wonder if we should reflect that in the names of the methods?

tomwhite:
@ravwojdyla I agree that the names aren't right. I think we might need two functions, since methods typically validate their inputs (which shouldn't modify the input dataset), and validate new variables in their output (which would be a good place to update attributes as well). So we could have `validate` and `validate_and_update_attributes`?

Another option would be to have a utility function that creates (known) variables in a dataset and sets the attributes at the time of creation.

eric-czech:
Very cool!

ravwojdyla:
> I think we might need two functions, since methods typically validate their inputs (which shouldn't modify the input dataset), and validate new variables in their output (which would be a good place to update attributes as well). So we could have validate and validate_and_update_attributes?

Naming ... 🤔  How about `validate` for to "validate inputs", and `annotate` to "annotate newly created dataset". We can say that `annotate` goes over the variables and validates and annotates them. wdyt @tomwhite?

> Another option would be to have a utility function that creates (known) variables in a dataset and sets the attributes at the time of creation.

afair I've played around with that kind of API, but it ended up being more user code, but I'm curious how would you envision that @tomwhite?

tomwhite:
> afair I've played around with that kind of API, but it ended up being more user code, but I'm curious how would you envision that

I was thinking of a wrapper around the `xarray.Dataset` [constructor](http://xarray.pydata.org/en/stable/generated/xarray.Dataset.html) that delegates to Xarray, then sets comment attributes for all registered variables.

ravwojdyla:
@tomwhite I'm +1, as long as that is going to play well with `conditional_merge_datasets`/`merge_datasets`, but I guess we can have both custom xarray ctor and public "annotate" method used by the ctor.

tomwhite:
The latest commit (https://github.com/pystatgen/sgkit/pull/420/commits/af41aef7687f71d5a4dbb647cfef9f19f9ad3153) shows a `create_dataset` function that calls `annotate`.

tomwhite:
@ravwojdyla what do you think of the approach in https://github.com/pystatgen/sgkit/pull/420/commits/af41aef7687f71d5a4dbb647cfef9f19f9ad3153? Shall I apply it across the codebase?

sgkit/utils.py
```diff
@@ -194,6 +195,16 @@ def define_variable_if_absent(
     return func(ds)
 
 
+def create_dataset(
+    data_vars: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+    coords: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+    attrs: Mapping[Hashable, Any] = None,  # type: ignore[assignment]
+) -> Dataset:
```
ravwojdyla, line 202: \+ doc from xr.Dataset CTOR?

sgkit/variables.py
```diff
@@ -115,10 +117,59 @@ def _check_field(
                 f"{field} does not match the spec, see the error above for more detail"
             ) from e
 
+    @classmethod
+    def _annotate(
```
ravwojdyla, line 121: These should probably reuse code from `_validate`/`_check_field`?

ravwojdyla (approved):

ravwojdyla:
@tomwhite I'm +1. I do want to highlight one thing tho, this adds a "sgkit blessed" xarray dataset contractor, which is somewhat close to the idea of a lightweight wrapper around xarray, but arguable a bit harder to discover. Also does the implementation in af41aef work well with `conditional_merge_datasets`/`merge_datasets`?

tomwhite:
> I do want to highlight one thing tho, this adds a "sgkit blessed" xarray dataset contractor, which is somewhat close to the idea of a lightweight wrapper around xarray, but arguable a bit harder to discover. 

That's a very good point. All of the variable and validation logic is "opt-in", so it should always be possible for users writing statgen methods to just use regular xarray APIs when using sgkit.

Regarding being harder to discover, the function should be documented better and added to the public API when we are happy with it.

> Also does the implementation in [af41aef](https://github.com/pystatgen/sgkit/commit/af41aef7687f71d5a4dbb647cfef9f19f9ad3153) work well with `conditional_merge_datasets`/`merge_datasets`?

I think there are further possible improvements here. I was thinking of making `create_dataset` also call `variables.validate()`. It could also (optionally) do the dataset merging, or we could have another function `create_dataset_from` that takes an input dataset.

ravwojdyla:
> I think there are further possible improvements here. I was thinking of making create_dataset also call variables.validate(). It could also (optionally) do the dataset merging, or we could have another function create_dataset_from that takes an input dataset.

Sounds good, I wonder what's the good middle ground between scope creeping this PR and just adding the descriptions in a good enough way. But I leave that up to you. Regarding `create_dataset` vs `create_dataset_from`, we could also have a single `create_dataset` that optionally takes a variable number of Datasets to merge from. So it would be one stop CTOR for sgkit blessed (validated/documented) Datasets.

tomwhite:
Updated to remove code duplication between `validate` and `annotate`, and to convert calls in the main codebase from `xr.Dataset` to `create_dataset`. Also added doc for `create_dataset`.

I've left changes to merging for later.

ravwojdyla (approved):
🎊

PR #423, opened 2020-12-15
Use fsspec in load_dataset and save_dataset #395

### LINKED ISSUE #395: Use fsspec in load_dataset and save_dataset
Following on from #392.

We need to make sure that resources are properly closed, as simply calling `fsspec.get_mapper()` resulted in warnings when running on local files.

tomwhite:
Fixes #395.

The problem with warnings about files not being closed was fixed in https://github.com/intake/filesystem_spec/pull/463. There hasn't been an `fsspec` release with that fix in yet, but I think we can merge this PR anyway, since I haven't seen it cause any problems (and @eric-czech has been using something equivalent in https://github.com/related-sciences/ukb-gwas-pipeline-nealelab/blob/4f862e31b8093d25fdaa8da7f841b9be8583cda4/scripts/gwas.py#L53-L71).

### DIFF
--- a/sgkit/io/dataset.py
+++ b/sgkit/io/dataset.py
@@ -1,12 +1,18 @@
-from typing import Any
+from typing import Any, Dict, Optional
 
+import fsspec
 import xarray as xr
 from xarray import Dataset
 
 from sgkit.typing import PathType
 
 
-def save_dataset(ds: Dataset, path: PathType, **kwargs: Any) -> None:
+def save_dataset(
+    ds: Dataset,
+    path: PathType,
+    storage_options: Optional[Dict[str, str]] = None,
+    **kwargs: Any
+) -> None:
     """Save a dataset to Zarr storage.
 
     This function is a thin wrapper around :meth:`xarray.Dataset.to_zarr`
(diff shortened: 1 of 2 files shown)


jeromekelleher:
LGTM, but it does feel a bit icky having a ``storage_options`` dictionary for passing arguments to ``fsspec`` but then using ``kwargs`` to pass arguments to zarr. Ideally, I guess, from a user perspective you'd like to just pass them all in as kwargs and let the library figure out which option goes where. Obviously this makes our code a bit more complicated and it is a maintenance burden. How many parameters are we envisaging this using? How stable are the relevant APIs?

sgkit/io/dataset.py
```diff
@@ -18,17 +24,25 @@ def save_dataset(ds: Dataset, path: PathType, **kwargs: Any) -> None:
         Dataset to save.
     path
         Path to directory in file system to save to.
+    storage_options:
+        Any additional parameters for the storage backend (see ``fsspec.open``).
     kwargs
         Additional arguments to pass to :meth:`xarray.Dataset.to_zarr`.
     """
-    store = str(path)
+    if isinstance(path, str):
```
eric-czech, line 32: My recollection was that the mapper should be created here to do any context management (I thought something couldn't be left open without throwing a warning?).  I don't see that anywhere in this PR though so is it still true that `path` can't be `store: Union[MutableMapping, Path, str]` like in `to_zarr`?

eric-czech (changes requested):

sgkit/io/dataset.py
tomwhite: > My recollection was that the mapper should be created here to do any context management (I thought something couldn't be left open without throwing a warning?).

The mapper doesn't have a close method, so it's not used within a context manager itself. Instead, each get and set item call on the mapper uses a context manager internally to ensure resources are closed. There is a bug in the latest fsspec release (fixed in intake/filesystem_spec#463) where a context manager is *not* used for the get item call, which is why there are lots of warnings when running tests.

Re `MutableMapping` - we should probably allow that too.

tomwhite:
> Ideally, I guess, from a user perspective you'd like to just pass them all in as kwargs and let the library figure out which option goes where.

The `storage_options` dictionary is a pattern used elsewhere, e.g. in Dask it's used here to pass options to fsspec: https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.read_csv, while `kwargs` are passed to Pandas. We've also used `storage_options` in sgkit like this in the VCF code.

> How many parameters are we envisaging this using? How stable are the relevant APIs?

Many of the storage options are filesystem dependent (https://docs.dask.org/en/latest/remote-data-services.html has a good rundown), so I think it would be difficult to maintain as a single `kwargs`. Also, it looks like for S3 the storage options can include their own `kwargs` too!

sgkit/io/dataset.py
eric-czech: > Re MutableMapping - we should probably allow that too.

👍

jeromekelleher:
Fair enough, thanks for explaining @tomwhite.

sgkit/io/dataset.py
tomwhite: Added in https://github.com/pystatgen/sgkit/pull/423/commits/55fdd4e54e5f3d6d4eb31462126c254d77928b29

eric-czech (approved):

Issue #374, opened 2020-11-05
PC Relate validation suite is failing


tomwhite:
E.g. https://github.com/pystatgen/sgkit/runs/1355826443?check_suite_focus=true

ravwojdyla:
Looks like something in the docker image got updated and is causing different results. Will look into this.

ravwojdyla:
Some context on the issue:
* when I run the validation on an old image, everything works fine, thus there has to be an issue in the docker itself (likely R packages)
* BioManager has a special instructions to install specific versions of packages, documented [here](https://cran.r-project.org/web/packages/BiocManager/vignettes/BiocManager.html).

When I compare the new image with an old one, there are some updated packages, noticeably `GENESIS`: `2.18.0` -> `2.20.0` (which is where PR Relate comes from).

Kinship coefficient distributions:
![newplot (20)](https://user-images.githubusercontent.com/1419010/99241884-bb26f100-27fe-11eb-99a5-de0ff0826153.png)

ravwojdyla:
#382 was merged, I've manually triggered the validation and it was [successful](https://github.com/pystatgen/sgkit/actions/runs/365818796).

ravwojdyla:
We could technically close this issue, since the validation suite isn't failing anymore. But at the same time there seem to be a difference in the results, which could be interesting to investigate further. In the past we have agreed to have a point in time validation. In #382 I also add a documentation to the `pc_relate` that the user can expect results compatible with GENESIS 2.18.0. This situation might appear also for other methods like `regenie`. How do we want to handle this?

hammer:
My vote would be to close this issue, since the validation suite is no longer failing, and file another issue to follow up on the failure caused by the GENESIS: 2.18.0 -> 2.20.0 update.

As discussed in the past, I don't think we should commit to validate our implementation against all future releases--that's just too much work. We could document this commitment someplace, I suppose.

ravwojdyla:
Validation suite is failing again - for a different reason tho, we can reuse this issue I guess. Problem:

```
Traceback:
usr/local/lib/python3.8/dist-packages/dask/distributed.py:3: in <module>
    from distributed import *
usr/local/lib/python3.8/dist-packages/distributed/__init__.py:3: in <module>
    from .actor import Actor, ActorFuture
usr/local/lib/python3.8/dist-packages/distributed/actor.py:7: in <module>
    from .client import Future, default_client
[...]

PR #443, opened 2021-01-26
Use the same distributed as dask

ravwojdyla:
Re: #374 
There seem to be some dependencies between those two (like `dask.utils`), so if we pin `dask` (like in #426), we should also pin `distributed`, otherwise users (or in this case validation suite) will experience hard to debug failures. In case of the validation suite it was a cryptic:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'stringify' from 'dask.utils' (/Users/rav/miniconda3/envs/_sgkit_test/lib/python3.8/site-packages/dask/utils.py)
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 numpy
 xarray
 dask[array] == 2.30.0
+distributed == 2.30.0
 dask-ml
 scipy
 typing-extensions
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ install_requires =
     numpy
     xarray
     dask[array] == 2.30.0
+    distributed == 2.30.0
     dask-ml
     scipy
     zarr


tomwhite (approved):

hammer:
Good catch! Is there any way to test requirements to ensure these kinds of inconsistencies are caught by code in the future?

ravwojdyla:
@hammer I'm not sure if this issue failing validation suite counts as being caught by code? Here it might be worth noting that in the PC Relate validation we only install sgkit + pytest, which more closely represents a potential user (than for example the dev environment with all the extra dev dependencies). For future it might be interesting to add the Upstream CI with main branches of the core dependencies (#444) and run some validation use cases in environment closely resembling a potential user.

hammer:
> I'm not sure if this issue failing validation suite counts as being caught by code?

Fair point. I had imagined something a bit closer to `requirements.txt`/`setup.cfg` where we might be able to make assertions over the dependency graph produced by `pip` or whichever tool is reading and resolving those files. I don't think it's possible or worth it though, just a thought I had.

hammer:
(Also will this PR fix the validation suite? That would be lovely. The red box in our README makes me a little bit sad.)

ravwojdyla:
@hammer yes, it should fix the validation suite. And yea, I get anxiety each time I see that red badge (which I guess means it's "working").

Issue #444, opened 2021-01-26
Add upstream CI build


ravwojdyla:
Dask has an "upstream" CI build, which is an interesting take on this problem, it essentially runs dask CI with main/master versions of some of the core dependencies. They run this on schedule as well as on master changes. This could be valuable for us to detect breaking changes before releases. See the upstream CI definition [here](https://github.com/dask/dask/blob/master/.github/workflows/ci-upstream.yml).

Edit: but I'm not sure if it leads to "actions", see the state of that [build](https://github.com/dask/dask/actions?query=workflow%3AUpstream).

_Originally posted by @ravwojdyla in https://github.com/pystatgen/sgkit/issues/422#issuecomment-757365235_

ravwojdyla:
Our current dependencies:

```
numpy
xarray
dask[array] == 2.30.0
dask-ml
scipy
typing-extensions
numba
zarr
fsspec
```

out of those, we should probably test upstream of:

```
xarray
dask[array] 
distributed
zarr
```

as our "core", wdyt?

tomwhite:
I agree those core dependencies are the ones to track.

ravwojdyla:
Closed via #445

PR #430, opened 2020-12-22
Fix PCA tests running on Dask 2020.12.0

tomwhite:
This will fail and should not be merged until there is a new release of dask-ml with https://github.com/dask/dask-ml/pull/767 in it.

PR #445, opened 2021-01-27
Add upstream CI

ravwojdyla:
Re #444

This is right now draft, wanna get some feedback on the approach before polishing.

### DIFF
--- a/.github/scripts/upstream_install.py
+++ b/.github/scripts/upstream_install.py
@@ -1,0 +1,49 @@
+import subprocess
+import sys
+from pathlib import Path
+
+
+def install_deps() -> None:
+    # NOTE: need to use legacy-resolver due to https://github.com/dask/community/issues/124
+    install_cmd = (
+        sys.executable,
+        "-m",
+        "pip",
+        "install",
+        "--use-deprecated=legacy-resolver",
+        "--upgrade",
+    )
+    upstream_deps = (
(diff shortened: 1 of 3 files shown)


ravwojdyla:
@tomwhite I was expecting the upstream tests to pass given the comments in https://github.com/pystatgen/sgkit/pull/430 ~> https://github.com/dask/dask/pull/7004. But it seems like there are still failures: https://github.com/ravwojdyla/sgkit/runs/1777383393?check_suite_focus=true, is this expected?

tomwhite:
Looks like we need dask-ml too, due to https://github.com/dask/dask-ml/commit/7d04c9519aed58c66c4e0577240e20a5ab1a3a96?

Approach looks good to me.

ravwojdyla:
@tomwhite this is ready to review. As you can see the Upstream build is failing ([link](https://github.com/pystatgen/sgkit/pull/445/checks?check_run_id=1793482369), BUT not due to the outdated dask-ml, there are 2 separate issues that the upstreams "introduce", I suggest we fix/investigate those in separate PRs? Also should we add badge to readme? I presume upstream build will be red pretty frequently. wdyt?

tomwhite (approved):
Looks great - thanks for doing this.

+1 to fixing the issues causing problems in separate PRs. I think we should probably add a badge if only to surface the extent of the problem.

ravwojdyla:
Added the badge, will merge when green.

PR #447, opened 2021-01-28
Rewrite pairwise to remove concatenation from blockwise

### LINKED ISSUE #375: Pairwise distance scalability
Raising this issue to revisit the scalability of our pairwise distance calculation and whether it's worth returning to a map-reduce style implementation that would allow chunking along both dimensions.

In the work that @aktech is doing on early scalability demonstrations (#345) there are some memory usage difficulties emerging. @aktech is I believe trying to run a pairwise distance computation over data from Ag1000G phase 2, using all SNPs and samples from a single chromosome arm. This is of the order ~20 million variants and ~1000 samples. With the current implementation, it is hard to get this to run on systems with average memory/CPU ratios, below 12G/CPU. My understanding is that, essentially, this is because the pairwise distance implementation currently does not support chunking along the variants dimension, and so to reduce memory footprint you need short chunks along the samples dimension. Depending on how the input data have been chunked natively, this may be suboptimal, i.e., you may need to run the computation with sample chunks that are (much) shorter than the native storage.

If this is correct, then it raises two questions for discussion.

First, should we revisit the map-reduce implementation of pairwise distance? This would allow chunking on both samples and variants dimensions, and so could naturally make use of whatever the underlying chunking of the data in storage, without large memory footprint. 

Secondly, do we ever really need to run pairwise distance on arrays that are large in the variants dimension? I.e., do we care about scaling this up to large numbers of variants? xref https://github.com/pystatgen/sgkit/pull/306#issuecomment-714654217

aktech:
This rewrites pairwise distance to fix the concatenation in blockwise. The implementation is motivated from dask's [matmul PR](https://github.com/dask/dask/pull/7000/files).

Fixes #375 

I ran this on MalariaGEN data, it took nearly the same time as blocks implementation on 50 Workers, 16 GB each. The time comparions can be seen in the notebook [here](https://nbviewer.jupyter.org/github/aktech/sgkit/blob/pairwise-via-blockwise/reports-and-notebooks/sgkit_with_coiled_euclidean.ipynb)

The dask reports can be seen here:

### Blocks implementation:
- [dask report](https://rawcdn.githack.com/aktech/sgkit/b37b3895a84352cfd272bbd2b6fb38c06b32068a/reports-and-notebooks/dask-report-pairwise-distance-blocks.html): **20 min 28 sec**
- [task stream](https://rawcdn.githack.com/aktech/sgkit/b37b3895a84352cfd272bbd2b6fb38c06b32068a/reports-and-notebooks/task-stream-pairwise-distance-blocks.html)

### Blockwise implementation:
- [dask report](https://rawcdn.githack.com/aktech/sgkit/b37b3895a84352cfd272bbd2b6fb38c06b32068a/reports-and-notebooks/dask-report-pairwise-distance-blockwise.html): 
- [task stream](https://rawcdn.githack.com/aktech/sgkit/b37b3895a84352cfd272bbd2b6fb38c06b32068a/reports-and-notebooks/task-stream-pairwise-distance-blockwise.html): **21 min 31 sec**

Important note: These times are only rough estimates for comparison, not  for quoting. When I ran the blocks implementation on the same configuration of coiled cloud couple of months ago it took ~ 10 min, now its taking more, since workers are dying due to: `OSError: Timed out during handshake while connecting to tls:...`. I am not sure why is that happening since no code change has bee made in the blocks implementation. I am communicating with coiled team to find the root cause at the moment.

cc @ravwojdyla 

### Update

A quick performance for correlation on only **10%** of malariaGEN data:

- [Notebook](https://nbviewer.jupyter.org/github/aktech/sgkit/blob/pairwise-via-blockwise/reports-and-notebooks/sgkit_with_coiled_correlation.ipynb)
- [Blocks dask report](https://rawcdn.githack.com/aktech/sgkit/c9efe081a20c1d42cbfacef565593e2fc93f9b50/reports-and-notebooks/dask-report-pairwise-distance-blocks-correlation.html):  **56.25 sec**
- [Blockwise dask report](https://rawcdn.githack.com/aktech/sgkit/f47025ed33b8900cb94b0fdbde517e67335ce35b/reports-and-notebooks/dask-report-pairwise-distance-blockwise-correlation.html): **58.94 sec**

TODO:
- [x] Remove notebooks and performance reports before merging.


sgkit/distance/api.py
```diff
@@ -77,27 +77,53 @@ def pairwise_distance(
...
-           [2.62956526e-01, 0.00000000e+00, 2.14285714e-01],
-           [2.82353505e-03, 2.14285714e-01, 0.00000000e+00]])
+    array([[-4.44089210e-16,  2.62956526e-01,  2.82353505e-03],
+           [ 2.62956526e-01,  0.00000000e+00,  2.14285714e-01],
+           [ 2.82353505e-03,  2.14285714e-01,  0.00000000e+00]])
     """
-
     try:
-        metric_ufunc = getattr(metrics, metric)
+        metric_map_func = getattr(metrics, f"{metric}_map")
```
ravwojdyla, line 85: nit, since `metric` only has 2 valid values, maybe it should be a Literal type instead of a str for better user API?

sgkit/distance/api.py
```diff
@@ -77,27 +77,53 @@ def pairwise_distance(
...
+        concatenate=False,
     )
-    x_distance = da.triu(x_distance, 1) + da.triu(x_distance).T
-    return x_distance.compute()
+
+    def _aggregate(x_chunk, axis, keepdims):
+        # This is to handle the case when an empty array is passed in the first iteration
+        # This check is not required in dask >= 2021.1.1
+        if x_chunk.size != 0:
+            x_chunk = x_chunk.reshape(x_chunk.shape[:-2] + (-1, n_map_param))
```
ravwojdyla, line 114: So this happens because dask needs to compute metadata for the reduction? Is this necessary? Is it possible/Should we specify `meta` for reduction to avoid this extra complexity? What's the common practice to deal with this?

sgkit/distance/metrics.py
```diff
@@ -1,52 +1,110 @@
...
     Parameters
     ----------
     x
-        [array-like, shape: (M,)]
-        A vector
+        An array chunk, a partial vector
     y
-        [array-like, shape: (M,)]
-        Another vector
+        Another array chunk,  a partial vector
```
ravwojdyla, line 38: extra space

sgkit/distance/metrics.py
```diff
@@ -1,52 +1,110 @@
...
+        The euclidean array on which map step of euclidean distance has been
+        applied.
+
+    Returns
+    -------
+    An ndarray, which contains square root of the sum of the squared sums obtained from
+    the map step of `euclidean_map`.
+    """
+    # This check is not required in dask >= 2021.1.1
+    if v.size != 0:
```
ravwojdyla, line 73: Is this related to the meta comment?

ravwojdyla (commented):
Thanks for the update @aktech! Scanned this PR, looks good, some small comments below, but nothing critical. As discussed in the chat, it would be nice to see the `correlation` metric performance report between blocks and blockwise.

sgkit/distance/api.py
aktech: Yes, makes sense.

ravwojdyla: @aktech feel free to do it in a follow up PR if you like.

sgkit/distance/api.py
aktech: > So this happens because dask needs to compute metadata for the reduction?

I believe so.

> Is it possible/Should we specify meta for reduction to avoid this extra complexity? What's the common practice to deal with this?

Good suggestion. I think it should be possible to use meta, let me try that.

sgkit/distance/metrics.py
aktech: Yes, it is.

sgkit/distance/api.py
aktech: I am happy to do this in this PR as well. I have been a bit swamped recently, hoping to wrap this up within couple of days.

aktech: It seems Literal Types were introduced in Python: 3.8 and we [support 3.7](https://github.com/pystatgen/sgkit/blob/c210fae783e927a45d5b852105cf42d9b47a32e4/.github/workflows/build.yml#L16) as well. So, I don't see a way to add them in here, that's why the precommit tests have been failing.

aktech:
@ravwojdyla I think I have addressed all the comments. I have added a quick performance test for `correlation` metric as well (See PR description above). I was initially seeing huge difference in performance between blockwise and blocks implementation, as in the blockwise was incredibly slower.

After the analysis of the dask performance reports, I found out the compute time was same and the issue as due to serialisation/deserialisation as I had included the `np.empty(..)` repeatedly and lambdas in the implementation which increased the serialisation/deserialisation significantly. Now I have fixed those issues and blockwise is as performant as blocks implementation.

sgkit/distance/api.py
tomwhite: We already depend on `typing-extensions` so you should be able to use that for Literal types.

aktech: Oh, great! I'll add that then in a bit.

aktech:
pre-commit seems to [fail](https://github.com/pystatgen/sgkit/pull/447/checks?check_run_id=1821415854) for unrelated changes:

```
sgkit/io/utils.py:186: error: Incompatible types in assignment (expression has type "number[Any]", variable has type "int")  [assignment]
sgkit/window.py:118: error: Item "ndarray" of "Union[ndarray, Any]" has no attribute "chunks"  [union-attr]
sgkit/stats/regenie.py:657: error: Incompatible types in assignment (expression has type "ndarray", variable has type "Optional[Sequence[float]]")  [assignment]
sgkit/stats/regenie.py:699: error: Argument "alphas" to "_stage_1" has incompatible type "Optional[Sequence[float]]"; expected "Optional[ndarray]"  [arg-type]
sgkit/stats/regenie.py:704: error: Argument "alphas" to "_stage_2" has incompatible type "Optional[Sequence[float]]"; expected "Optional[ndarray]"  [arg-type]
sgkit/stats/hwe.py:51: error: unused 'type: ignore' comment
sgkit/stats/hwe.py:94: error: unused 'type: ignore' comment
sgkit/stats/pca.py:133: error: Argument 1 to "__call__" of "ufunc" has incompatible type "DataArray"; expected "Union[Union[int, float, complex, str, bytes, generic], Sequence[Union[int, float, complex, str, bytes, generic]], Sequence[Sequence[Any]], _SupportsArray]"  [arg-type]
sgkit/io/vcfzarr_reader.py:199: error: Item "ndarray" of "Union[ndarray, Any]" has no attribute "values"  [union-attr]
sgkit/tests/test_pca.py:32: error: Returning Any from function declared to return "ndarray"  [no-any-return]
sgkit/io/bgen/bgen_reader.py:148: error: Returning Any from function declared to return "ndarray"  [no-any-return]
sgkit/io/bgen/bgen_reader.py:365: error: Non-overlapping container check (element type: "dtype[Any]", container item type: "Type[unsignedinteger[Any]]")  [comparison-overlap]
sgkit/tests/test_preprocessing.py:88: error: List item 0 has incompatible type "float"; expected "int"  [list-item]
sgkit/tests/test_preprocessing.py:124: error: Module has no attribute "bool"; maybe "bool_" or "bool8"?  [attr-defined]
sgkit/tests/test_preprocessing.py:162: error: Module has no attribute "bool"; maybe "bool_" or "bool8"?  [attr-defined]
sgkit/io/plink/plink_reader.py:87: error: Returning Any from function declared to return "ndarray"  [no-any-return]
sgkit/io/plink/plink_reader.py:268: error: Item "ndarray" of "Union[ndarray, Any]" has no attribute "compute"  [union-attr]
Found 17 errors in 10 files (checked 87 source files)
```

tomwhite:
@aktech I filed #451 for these (unrelated) failures

sgkit/distance/api.py
```diff
@@ -77,27 +82,61 @@ def pairwise_distance(
...
+        aggregate=_aggregate,
+        axis=-1,
+        dtype=x.dtype,
+        meta=np.ndarray((0, 0), dtype=x.dtype),
+        name="pairwise",
+    )
+
+    t = da.triu(r)
+    d = t + t.T
+    distance: np.ndarray = d.compute()
```
ravwojdyla, line 141: \- `compute()`

sgkit/distance/api.py
```diff
@@ -77,27 +82,61 @@ def pairwise_distance(
...
+
+    r = da.reduction(
+        out,
+        chunk=_chunk,
+        combine=_combine,
+        aggregate=_aggregate,
+        axis=-1,
+        dtype=x.dtype,
+        meta=np.ndarray((0, 0), dtype=x.dtype),
+        name="pairwise",
```
ravwojdyla, line 136: Question: would it make sense to expose `reduction `'s `split_every` to `pairwise_distance`, does it have impact depending on the chunking? wdyt? It certainly would not hurt to just expose it with default `None`, but I do wonder if there are cases where it might be actually useful in this computation. If it's not clear, feel free to open an issue to investigate this later.

sgkit/distance/api.py
```diff
@@ -77,27 +82,61 @@ def pairwise_distance(
...
+    if x.ndim != 2:
+        raise ValueError(f"2-dimensional array expected, got '{x.ndim}'")
+
+    metric_param = np.empty(n_map_param, dtype=x.dtype)
+
+    def _pairwise(f: ArrayLike, g: ArrayLike) -> ArrayLike:
+        result: ArrayLike = metric_map_func(f[:, None, :], g, metric_param)
+        return result[..., np.newaxis]
+
+    out = da.blockwise(
```
ravwojdyla, line 106: Would it make sense to comment here, why we try to avoid concatenate, maybe link to dask issue? wdyt?

sgkit/distance/api.py
```diff
@@ -77,27 +82,61 @@ def pairwise_distance(
...
-        # Lambda wraps reshape for broadcast
-        lambda _x, _y: metric_ufunc(_x[:, None, :], _y),
-        "jk",
+    if x.ndim != 2:
+        raise ValueError(f"2-dimensional array expected, got '{x.ndim}'")
+
+    metric_param = np.empty(n_map_param, dtype=x.dtype)
+
+    def _pairwise(f: ArrayLike, g: ArrayLike) -> ArrayLike:
+        result: ArrayLike = metric_map_func(f[:, None, :], g, metric_param)
```
ravwojdyla, line 103: Would it make sense to comment here why you need to add extra dimension?

sgkit/distance/api.py
```diff
@@ -77,27 +82,61 @@ def pairwise_distance(
...
 
     x = da.asarray(x)
-    x_distance = da.blockwise(
-        # Lambda wraps reshape for broadcast
-        lambda _x, _y: metric_ufunc(_x[:, None, :], _y),
-        "jk",
+    if x.ndim != 2:
+        raise ValueError(f"2-dimensional array expected, got '{x.ndim}'")
+
+    metric_param = np.empty(n_map_param, dtype=x.dtype)
```
ravwojdyla, line 100: Probably also worth commenting here why it's important to have this declared here? wdyt?

ravwojdyla (changes requested):

sgkit/distance/api.py
aktech: Now that I think about it, I think we don't really need this extra dimension. We can really simplify this by removing the extra dimension and making `combine` just return the chunk as the summation already happens in the metric's reduce function, this will also simplify the implementation and make is easy to grasp, let me change that and share in a bit. Let me know if you think otherwise.

sgkit/distance/api.py
aktech: Yeah, definitely.

sgkit/distance/api.py
aktech: If I make the `combine` (https://github.com/pystatgen/sgkit/pull/447#discussion_r570245094) to be trivial as just returning the chunk (instead of summing over an axis), then `split_every` might not have a lot of significance, (as per my understanding `split_every` depends on `combine` function). Also irrespective of that another way for people to use `split_every` is with setting the `split_every` key in `dask.config` (globally though). 

Otherwise I am happy to open an issue to investigate further, as at the moment I don't feel very comfortable about it.

sgkit/distance/api.py
aktech: Yes, indeed.

sgkit/distance/api.py
aktech: I had this in mind this time, but some how forgot. Thanks for pointing it out.

sgkit/distance/api.py
aktech: Ah, I see, combine makes sense for recursive aggregation, ignore the above comment in that case.

sgkit/distance/api.py
aktech: Ignore this comment, I think we can expose this ^

aktech:
The tests failures are unrelated, comments have been addressed. @ravwojdyla

ravwojdyla (approved):
Good work @aktech!

aktech:
@ravwojdyla Thanks for guiding me throughout, I have removed the notebooks.

PR #466, opened 2021-02-16
Setup air speed velocity benchmarks

### LINKED ISSUE #458: Setup air speed velocity for micro benchmarks
As discussed in the developer call it would be good to have some kind of benchmarking. Air speed velocity is a good tool for micro benchmarks: https://asv.readthedocs.io/en/stable/ as suggested.

Setting this up and throwing the artefacts somewhere will enable us to see the benchmarks over time like this:
https://pv.github.io/numpy-bench/

cc @jeromekelleher

aktech:
This setups the air speed velocity for micro benchmarking. Fixes #458 

### Benchmarks
I have added couple of functions from stats module to get us started with benchmarking, I am planning to add more, after some feedback on the approach and also after people have thoughts on which ones to benchmark (maybe in another follow up PR?)

### Benchmark Results
Benchmark results will be stored in a separate repo named something like: `sgkit-benchmarks` and that repo will also serve the static files to visualize benchmark results over time/commits.

### Frequency
Currently I am thinking to run benchmarks on every merge to master, so that people can see results on the static site, I am not sure if there is a value in running them on every commit/PR as they are not really a kind of tests. Open to suggestions on that though. The Github Action which runs the benchmarks on merge to master also pushes the results to the benchmarks repository with updated static files for visualization.

### Preview can be seen here:
http://iamit.in/sgkit-benchmarks/. (After repo creation in pystatgen, we'll see the results in http://pystatgen.github.io/sgkit-benchmarks/)

### Machine:
Ideally, one should use a fix/constant machine for benchmarking, but Github Actions machines don't change too much, so this gives us a free and quick way to setup up benchmarks, see [this](https://github.com/aktech/sgkit/blob/setup-asv-benchmarks/benchmarks/README.md#benchmark-machine) for more details: 

I have removed `prune sgkit/tests` from the `MANIFEST.in`, to be able to import `get_dataset` from `sgkit.tests.test_aggregation`.

### Instructions
All the necessary instructions have been added to `benchmarks/README.md`, let me know if anything is not clear or more information is required on anything.

Action items:
- [ ] Replace `aktech` with `pystatgen`
- [ ] Create a repo named `sgkit-benchmarks` in https://github.com/pystatgen
- [ ]  Add Github Personal Access Token in the repository secrets, this is used in the `.github/workflows/benchmark.yml` to be able to commit the results to the benchmarks repository.

cc @jeromekelleher @ravwojdyla @tomwhite


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Copy benchmarks to benchmarks repo directory
+        run: |
+          git clone https://$GITHUB_TOKEN@github.com/$BENCHMARKS_REPO.git ~/$BENCHMARKS_REPO
+          RESULTS_DIR=~/$BENCHMARKS_REPO/results
+          if [ -d "$RESULTS_DIR" ]
+          then
+            cp -r $RESULTS_DIR/$MACHINE_NAME/* benchmarks/results/$MACHINE_NAME/
+          else
+            echo "results/ directory does not exists in the benchmarks repository"
```
tomwhite, line 46:
```suggestion
            echo "results/ directory does not exist in the benchmarks repository"
```

benchmarks/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Usage
+
+Airspeed Velocity manages building the environment via conda itself. The recipe for the same in
+defined in the `asv.conf.json` configuration file.
+
+* To run the benchmark suite:
+
+```bash
+asv run --config benchmarks/asv.conf.json
```
tomwhite, line 13: Do I have to do anything extra to install asv?

benchmarks/README.md
```diff
@@ -0,0 +1,64 @@
+# sgkit Benchmarks
```
tomwhite, line 1: Should we move these instructions to https://pystatgen.github.io/sgkit/latest/contributing.html?

benchmarks/README.md
```diff
@@ -0,0 +1,64 @@
...
+```json
+{
+    "arch": "x86_64",
+    "cpu": "Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz",
+    "machine": "fv-az183-669",
+    "num_cpu": "2",
+    "os": "Linux 5.4.0-1039-azure",
+    "ram": "7121276",
+    "version": 1
+}
```
tomwhite, line 58: How did you discover this configuration? It might be useful to add a comment here explaining that.

It is possible to log the machine spec that each run took place on?

benchmarks/benchmarks_stats.py
```diff
@@ -0,0 +1,34 @@
...
+import numpy as np
+import xarray as xr
+
+from sgkit import count_call_alleles, count_cohort_alleles
+from sgkit.tests.test_aggregation import get_dataset
+
+
+def time_count_call_alleles() -> None:
+    count_call_alleles(
+        get_dataset(
```
tomwhite, line 12: I don't think the dataset creation should be a part of the benchmark, and should be moved to `setup` function, see https://asv.readthedocs.io/en/stable/writing_benchmarks.html#timing

benchmarks/benchmarks_stats.py
```diff
@@ -0,0 +1,34 @@
+"""Benchmarks suite for stats module."""
+
+import numpy as np
+import xarray as xr
+
+from sgkit import count_call_alleles, count_cohort_alleles
+from sgkit.tests.test_aggregation import get_dataset
```
tomwhite, line 7: I think you could use `simulate_genotype_call_dataset` directly (which is public), rather than `get_dataset`.

benchmarks/benchmarks_stats.py
```diff
@@ -0,0 +1,34 @@
...
+
+def time_count_cohort_alleles() -> None:
+    ds = get_dataset(
+        [
+            [[0, 0], [0, 0], [0, 0], [0, 0]],
+            [[0, 0], [0, 0], [0, 1], [0, 1]],
+            [[1, 1], [0, 1], [1, 0], [1, 0]],
+            [[1, 1], [1, 1], [1, 1], [1, 1]],
+        ]
+    )
```
tomwhite, line 31: These datasets seem rather small. Something 10 or 100 times larger would still run in under a second, and might be more representative of the per-row processing time (since one-off costs would be amortized).

tomwhite (commented):
This looks great @aktech!

I'm not sure what I think about using a GH actions machine - which may change in spec from run to run. Do you know if other projects are doing this? (And if not, what is their approach?)

I added a few comments and questions inline.

benchmarks/benchmarks_stats.py
jeromekelleher: Agreed, these are too small to be meaningful. We can use ``simulate_genotype_call_dataset`` to make some fixtures we can use. Probably we want something like samples = [10, 100, 1000] for a fixed 1000 (?) sites.

jeromekelleher (commented):
Yep, agree with @tomwhite, this looks great. I'm not sure about how well running this on Actions will work out, though, but maybe we should just try it.

benchmarks/README.md
aktech: Nothing extra apart from installing `requirements-dev.txt`

benchmarks/README.md
aktech: This is basically spitted out while running `asv machine` command. The problem I had while trying to log machine spec for each run is the machine name should be same for the graphs to be meaningful, otherwise it will be points for each machine. I can explore to log it anyways though, without affecting the graphs.

benchmarks/benchmarks_stats.py
aktech: Makes sense.

benchmarks/benchmarks_stats.py
aktech: Ah, I see thanks! didn't realise that.

aktech:
> I'm not sure what I think about using a GH actions machine - which may change in spec from run to run. Do you know if other projects are doing this? (And if not, what is their approach?)

Agree it's not the most ideal approach, most of the projects I looked at for motivation used a constant fixed machine. Github Action machine does change slightly in every run. I am happy to do this on a single machine, if we can get one?

ericdatakelly:
@aktech Aaron just did this for another project in case you'd like to confer with him.

benchmarks/README.md
aktech: I have added the logging of machine configuration of the run in the CI, did you meant saving that in the results repository?

tomwhite: Ideally yes, but this is a good start if that's not possible or easy to do.

sgkit/tests/test_aggregation.py
```diff
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, List, Union
```
tomwhite, line 1: The changes in this file are no longer needed, are they?

docs/contributing.rst
```diff
@@ -258,6 +258,61 @@ without warnings. You can do that locally with::
 
 .. _Sphinx: https://www.sphinx-doc.org/
 
+Contributing to Benchmarks
```
tomwhite, line 261: Maybe just call this section "Benchmarks"

tomwhite (approved):
These changes look good to me. Do the changes to the docs look OK when rendered?

Also, did anything come of conferring with your colleague about running ASV in practice?

sgkit/tests/test_aggregation.py
aktech: Actually, these were never required from test/mypy point of view since they are ignored for tests,  I did this to fix the type hint for `calls` param of `get_dataset` as we are inputting List of List as an argument. What do you think should we keep it?

aktech:
> Also, did anything come of conferring with your colleague about running ASV in practice?

He mentioned he is not benchmarking on the CI, he manually runs them on his local machine and uploads them to the results repository. He also mentioned there are a bunch of things we can do to make benchmarks more reliable like this:
https://pyperf.readthedocs.io/en/latest/system.html#system

I didn't tried this right away because, its probably a good idea to have these running for sometime to see how much deviation we get, before trying to do any optimisations.

Another thing he mentioned, might be useful is you can just have some fake benchmark that shouldn't actually change over time, then you can see the deviation due to different machine type, which I think is a good idea, I am happy to write a trivial  for loop or something like that, what do you think?

benchmarks/README.md
aktech: I think it is possible, as in we can somehow add a key in the results json here (assuming `asv` ignores it):
https://github.com/aktech/sgkit-benchmarks/blob/655317b0d50a52f85b4c43e81b00a4a72d76a0ed/results/github-actions/2196f390-conda-py3.8.json#L1

What do you think?

jeromekelleher (approved):
LGTM.

PR #470, opened 2021-02-23
Add alt_number to VCF reading

tomwhite:
This exposes `alt_number` in the VCF reading API, just like [scikit-allel](https://scikit-allel.readthedocs.io/en/stable/io.html#allel.vcf_to_zarr) does.

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -58,12 +58,10 @@ def vcf_to_zarr_sequential(
     ploidy: int = 2,
     mixed_ploidy: bool = False,
     truncate_calls: bool = False,
+    alt_number: int = DEFAULT_ALT_NUMBER,
 ) -> None:
 
     with open_vcf(input) as vcf:
-
-        alt_number = DEFAULT_ALT_NUMBER
-
         sample_id = np.array(vcf.samples, dtype=str)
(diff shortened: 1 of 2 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -261,6 +262,9 @@ def vcf_to_zarrs(
         If True, genotype calls with more alleles than the specified (maximum) ploidy value
         will be truncated to size ploidy. If false, calls with more alleles than the
         specified ploidy will raise an exception.
+    alt_number
```
jeromekelleher, line 265: Hmm, not sure ``alt_number`` captures this. What about ``max_alt_alleles``?

jeromekelleher (commented):
Looks good, but I'm not convinced ``alt_number`` is the right parameter name.

sgkit/io/vcf/vcf_reader.py
tomwhite: I was using the scikit-allel term, but happy to go with either.

jeromekelleher: We should probably get more opinions

hammer: I prefer `max_alt_alleles`.

Issue #464, opened 2021-02-14
Extract additional metadata from VCF files


hammer:
- [VCF 4.2 spec](https://samtools.github.io/hts-specs/VCFv4.2.pdf)
- Example VCF file: https://storage.googleapis.com/hail-tutorial/1kg.vcf.bgz
- [cyvcf2.pyx](https://github.com/brentp/cyvcf2/blob/master/cyvcf2/cyvcf2.pyx)
  - Header types: 'CONTIG', 'FILTER', 'FORMAT', 'GENERIC', 'INFO'
- [vcf_reader.py](https://github.com/pystatgen/sgkit/blob/master/sgkit/io/vcf/vcf_reader.py)

### ##INFO
- These fields are (usually?) per variant and can be stored as data variables with (variants) dimension prefixed with `variant_`. We can punt cases when `Number` is not 1 for now.
- We may want to consider specifically looking for some of the reserved keys enumerated in the spec.
- In particular, we may want to parse VEP annotations nicely.
- Filed https://github.com/brentp/cyvcf2/issues/192 to make it a bit easier to explore these fields with `cyvcf2`.

### ##FORMAT
- These fields are per genotype call and can be stored as data variables with (variants, samples) and sometimes (ploidy) dimensions. Again, we can punt cases when `Number` is not 1 for now.
- We may want to consider specifically looking for some of the reserved keys enumerated in the spec.
- `cyvcf2` makes it easy to see which FORMAT fields are available for each variant with `v.FORMAT`
- Some of the standard fields are also available as properties, with slight differences in representation, e.g. missing values in the VCF file are `.`, in`v.format('DP')` they are `-2147483648`, and in `v.gt_depths` they are `-1`.

```py
# v.format('DP')
v.gt_depths

# v.format('AD')
list(zip(v.gt_ref_depths, v.gt_alt_depths))

# v.format('GQ')
v.gt_quals

# v.format('PL')
list(zip(v.gt_phred_ll_homref, v.gt_phred_ll_het, v.gt_phred_ll_homalt))
```

### ##CONTIG
- It would be nice to get the assembly name and contig length too.
- `vcf.seqnames` has contig names
- `vcf.seqlens` has contig lengths
- `vcf['CONTIG']` should get `contig` header lines but it's giving me a `KeyError`; some kind of encoding issue, I guess.
[...]

PR #471, opened 2021-02-23
VCF info and format fields

tomwhite:
This changes the VCF reader to handle INFO and FORMAT fields of type Flag, Integer, or Float, and where number is 1, as suggested in #464.

A few things for discussion:
* Unlike scikit-allel which has a single `fields` list, I've created two, `info_fields` and `format_fields` to avoid ambiguities (like `DP`, which can apply to either), Alternatively, we could have a single list, and namespace them, e.g. `FORMAT/DP`.
* Naming. As it stands INFO fields are prefixed with `variant_`, and FORMAT fields with `call_`. So `call_DP` is an example. Do we want to add a renaming keyword (like scikit-allel has) so it could be renamed to e.g. `call_depth`?
* The `GT` FORMAT field is special since it is always included, even if `format_fields` doesn't include it. Should we make it optional?
* The `vcf_to_zarr_sequential` is getting a bit unwieldy and should probably be refactored into a class or set of classes to handle various field types.

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -103,6 +103,7 @@ def _vcf_type_to_numpy_type_and_fill_value(
         return "bool", False
     elif vcf_type == "Integer":
         return "i4", -1
+    # the VCF spec defines Float as 32 bit, and in BCF is stored as 32 bit
     elif vcf_type == "Float":
         return "f4", np.nan
     elif vcf_type == "String":
@@ -485,6 +486,8 @@ def vcf_to_zarrs(
         overrides the ``INFO/AC`` field to be Number ``A`` (useful if the VCF defines it as
         having variable length with ``.``), and names the final dimension of the ``HQ`` array
         (which is defined as Number 2 in the VCF header) as ``haplotypes``.
+        (Note that Number ``A`` is the number of alternate alleles, see section 1.4.2 of the
+        VCF spec https://samtools.github.io/hts-specs/VCFv4.3.pdf.)
 
     Returns
     -------
@@ -636,6 +639,8 @@ def vcf_to_zarr(
         overrides the ``INFO/AC`` field to be Number ``A`` (useful if the VCF defines it as
         having variable length with ``.``), and names the final dimension of the ``HQ`` array
         (which is defined as Number 2 in the VCF header) as ``haplotypes``.
+        (Note that Number ``A`` is the number of alternate alleles, see section 1.4.2 of the
+        VCF spec https://samtools.github.io/hts-specs/VCFv4.3.pdf.)
     """
 
     if temp_chunk_length is not None:
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -1,8 +1,19 @@
 import functools
 import itertools
 from contextlib import contextmanager
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, Hashable, Iterator, MutableMapping, Optional, Sequence, Union
+from typing import (
+    Any,
+    Dict,
+    Hashable,
+    Iterator,
+    MutableMapping,
+    Optional,
+    Sequence,
+    Tuple,
+    Union,
+)
 
 import dask
 import fsspec
@@ -13,8 +24,10 @@ from cyvcf2 import VCF, Variant
 from sgkit.io.utils import zarrs_to_dataset
 from sgkit.io.vcf import partition_into_regions
 from sgkit.io.vcf.utils import build_url, chunks, temporary_directory, url_filename
-from sgkit.model import DIM_VARIANT, create_genotype_call_dataset
-from sgkit.typing import PathType
+from sgkit.io.vcfzarr_reader import vcf_number_to_dimension_and_size
+from sgkit.model import DIM_SAMPLE, DIM_VARIANT, create_genotype_call_dataset
+from sgkit.typing import ArrayLike, DType, PathType
+from sgkit.utils import max_str_len
 
 DEFAULT_MAX_ALT_ALLELES = (
     3  # equivalent to DEFAULT_ALT_NUMBER in vcf_read.py in scikit_allel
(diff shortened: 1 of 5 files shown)


jeromekelleher:
> Naming. As it stands INFO fields are prefixed with variant_, and FORMAT fields with call_. So call_DP is an example. Do we want to add a renaming keyword (like scikit-allel has) so it could be renamed to e.g. call_depth?

One quick comment here: I'd tend to avoid renaming stuff from VCF. Ultimately we're just representing the VCF data so we should keep as close to the original as we can, IMO. By using ``call_DP`` it's clear we're not doing anything clever, just showing the user the value of the DP field in the VCF.

jeromekelleher:
This is great @tomwhite! I've had a quick skim through and it looks very nice.

I think it would be worthwhile building up a library of weird and wonderful VCF examples at this point, I suspect some of the assumptions being made won't hold up in practise (that was my repeated experience with the [VCF parser](https://github.com/wormtable/wormtable/blob/master/wormtable/vcf2wt.py) in wormtable. Using htslib/cyvcf2 is a huge help, but I suspect people play pretty fast and loose with INFO field types and so on, and we'll need to provide the user with ways of defining the numpy dtype, etc, themselves.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -126,13 +186,31 @@ def vcf_to_zarr_sequential(
                 call_genotype[i, ..., n:] = fill
                 call_genotype_phased[i] = gt[..., -1]
 
+                for info_field, arr in info_field_arrays.items():
+                    if info_field in [k for k, _ in variant.INFO]:
+                        if isinstance(variant.INFO[info_field], tuple):
+                            # for the case Number='.', just use the first value
```
timothymillar, line 192: It might be better to also throw an error for the '.' case as this is commonly used indicate that the field is variable in length, see this old [htslib issue](https://github.com/samtools/htslib/issues/664)

tomwhite: Agreed - let's not support this until it's done properly.

tomwhite:
> I think it would be worthwhile building up a library of weird and wonderful VCF examples at this point

I've used test data from GATK for this in the past. It also looks like scikit-allel has some test VCFs that have been chosen/altered to test issues such as these.

tomwhite:
Here's an update which better encapsulates the INFO and FORMAT field parsing, and removes support for the Number `'.'` case.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -49,6 +60,81 @@ def get_region_start(region: str) -> int:
...
+            n_sample = len(vcf.samples)
+            chunksize = (chunk_length, n_sample)
+            array = np.full(chunksize, -1, dtype=dtype)
+
+        return cls(category, key, variable_name, description, dims, chunksize, array)
+
+    def add_variant(self, i: int, variant: Any) -> None:
+        key = self.key
+        if self.category == "INFO":
+            if key in [k for k, _ in variant.INFO]:
```
timothymillar, line 124: I think you can use `variant.INFO.get(key)` here

jeromekelleher:
Here's a ready-made set of example VCFs @tomwhite: https://github.com/ga4gh/ga4gh-server/tree/master/tests/data/datasets/dataset1/variants

This is from the old GA4GH REST API reference implementation, where we did try to support all the INFO/GT fields in VCFs.

tomwhite:
I've overhauled this quite a bit with the following changes:
* Support all VCF Type fields (Integer, Float, String, Flag)
* Support the following VCF Number fields: any integer, A, R (but not `.` or G).
* Allow field definitions to be overridden using the `field_defs` argument. (This is useful to override a `.` Number field with a fixed number, for example.)
* Allow fields to be specified by a wildcard (e.g. `INFO/*`).
* Convert INFO and FORMAT fields from scikit-allel format, and add roundtrip tests to check equivalence. This is a useful check since it's effectively testing two VCF parsers against one another.

I've checked equivalence for all the test files in sgkit's test data, and it would be fairly straightforward to extend this to other sets of VCF files. That could be done separately though.

timothymillar:
This looks great @tomwhite. The size of "G" can actually be calculated from `n_alleles` and `ploidy` using n-choose-k with replacement 

```python
from math import comb
comb(n_alleles + ploidy - 1, ploidy)
```

hammer:
@timothymillar do you have any VCFs with data from variable ploidy samples that you might be able to contribute to our testing efforts?

timothymillar:
@hammer I'm looking into this, we can definitely share some data for (a chromosome of) a 4x potato population which is already public. I'm not aware of a true mixed ploidy example but at the very least I can call the 4x material with a range of ploidy values to make something similar.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -49,6 +64,163 @@ def get_region_start(region: str) -> int:
...
+
+def _vcf_type_to_numpy_type_and_fill_value(
+    vcf_type: str, category: str, key: str
+) -> Tuple[DType, Any]:
+    """Convert the VCF Type to a NumPy dtype and fill value."""
+    if vcf_type == "Flag":
+        return "bool", False
+    elif vcf_type == "Integer":
+        return "i4", -1
+    elif vcf_type == "Float":
```
jeromekelleher, line 106: Worth putting a note in here that BCF stores these values as single precision, so there's no point in using doubles.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -49,6 +64,163 @@ def get_region_start(region: str) -> int:
...
+            else:
+                self.array[i] = self.fill_value
+
+    def truncate_array(self, length: int) -> None:
+        self.array = self.array[:length]
+
+    def update_dataset(
+        self, ds: xr.Dataset, add_str_max_length_attrs: bool = False
+    ) -> None:
+        # cyvcf2 represents missing Integer values as the minimum int32 value
```
jeromekelleher, line 211: I'm slightly worried about this - could we be silently altering data? Suppose someone uses INT32_MIN in their data, won't we silently change this to fill_value then? I guess this is an upstream problem with cyvcf2.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -261,6 +467,24 @@ def vcf_to_zarrs(
         If True, genotype calls with more alleles than the specified (maximum) ploidy value
         will be truncated to size ploidy. If false, calls with more alleles than the
         specified ploidy will raise an exception.
+    max_alt_alleles
+        The (maximum) number of alternate alleles in the VCF file. Any records with more than
+        this number of alternate alleles will have the extra alleles dropped.
```
jeromekelleher, line 472: Perhaps not for this PR, but we should probably raise a warning if this happens. Users should know if their ALTs are being dropped, and ideally it should be discoverable later if they have been truncated. Could we add some metadata tracking when this has happened?

sgkit/io/vcf/vcf_reader.py
```diff
@@ -261,6 +467,24 @@ def vcf_to_zarrs(
...
+    fields
+        Extra fields to extract data for. A list of strings, with ``INFO`` or ``FORMAT`` prefixes.
+        Wildcards are permitted too, for example: ``["INFO/*", "FORMAT/DP"]``.
```
jeromekelleher, line 475: NICE!!!

sgkit/io/vcf/vcf_reader.py
```diff
@@ -261,6 +467,24 @@ def vcf_to_zarrs(
...
+    field_defs
+        Per-field information that overrides the field definitions in the VCF header, or
+        provides extra information needed in the dataset representation. Definitions
+        are a represented as a dictionary whose keys are the field names, and values are
+        dictionaries with any of the following keys: ``Number``, ``Type``, ``Description``,
+        ``dimension``. The first three correspond to VCF header values, and ``dimension`` is
+        the name of the final dimension in the array for the case where ``Number`` is a fixed
+        integer larger than 1. For example,
+        ``{"INFO/AC": {"Number": "A"}, "FORMAT/HQ": {"dimension": "haplotypes"}}``
```
jeromekelleher, line 484: Good to explain what "A" means here; maybe  we should say something like "see section 1.2.2 of the `VCF specification <url>`__ for further details on the available Number specifications."

jeromekelleher (approved):
This is fantastic @tomwhite, minor comments above! Probably some of these should be pushed out to follow-up PRs.

sgkit/io/vcf/vcf_reader.py
tomwhite: I opened https://github.com/brentp/cyvcf2/issues/195 for this.

timothymillar: I don't think they could use INT32_MIN for anything else as this value is explicitly interpreted as a 'missing' integer in the [VCF spec](https://samtools.github.io/hts-specs/VCFv4.3.pdf) (page 30)

Issue #416, opened 2020-12-03
Genotype ploidy summary functions


timothymillar:
It would be useful to be able to summarize ploidy along different dimensions when working with mixed ploidy data.
Currently ploidy can only be inferred from the call_genotype array by the number of allele values > -1 at each call.

Methods could be implemented to calculate the following arrays:
- `'call_genotype_ploidy'` of shape (variants, samples)
- `'variant_ploidy'` of shape (variants, )
- `'sample_ploidy'` of shape (samples, )

If ploidy is variable across a given variant or sample then -1 should be returned indicating mixed-ploidy along that dimension.

There was some discussion of this in #243 but it was agreed that this is better as a separate PR [#373 (comment)](https://github.com/pystatgen/sgkit/pull/373#pullrequestreview-524217735)

Example:
call_genotype array:
```
[[[0, 0, -1, -1], [0, 0,  1,  1], [0, 0, 1, 1]],
 [[0, 0,  0, -1], [0, 1, -2, -2], [0, 1, 1, 1]],
 [[0, 0,  0,  1], [0, 0, -2, -2], [0, 0, 0, 0]],]
```
call_genotype_ploidy array:
```
[[4, 4, 4],
 [4, 2, 4],
 [4, 2, 4]]
```
variant_ploidy array:
```
[ 4,
 -1,
 -1]
```
sample_ploidy array:
```
[4, -1, 4]
```

jeromekelleher:
This seems very sensible to me @timothymillar. I guess one might also reasonably like to know what the maximum ploidy is, so should we add an option to return this, rather than -1 in the mixed case? Or, perhaps we could return (-1 * maximum_ploidy) if it's mixed? Maybe this is trying to be too clever though, and would be confusing/lead to errors.

timothymillar:
It could be nice to return something more informative where the ploidy is variable, however I think I'd prefer to keep it simple for now. One use case for these methods/arrays is for when another statistic needs to check for consistent ploidy across a specific dimension. So keeping the nan/false value simple is probably easier.

The max/min ploidy can be attained with standard methods e.g `call_genotype_ploidy.max(axis=...)`. The proposed methods are really to fill a gap in the generic methods "if all values are equal return value, else return nan/-1".

jeromekelleher:
Thanks @timothymillar, your approach SGTM.

tomwhite:
Can this be closed now that #472 is merged?

timothymillar:
[...]

PR #472, opened 2021-02-25
Add ploidy summary functions #416

timothymillar:
See #416 

Ac ouple of notes:
- I don't know what I'm missing to populate the "comment" attribute in the HTML REPR?
- I think 'infer' is a better verb than 'impute' in this case.

### DIFF
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -546,3 +546,169 @@ def sample_stats(
         ]
     )
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+def infer_non_alleles(
+    ds: Dataset,
+    *,
+    call_genotype: Hashable = variables.call_genotype,
+    merge: bool = True,
+) -> Dataset:
+    variables.validate(ds, {call_genotype: variables.call_genotype_spec})
+    mixed_ploidy = ds[variables.call_genotype].attrs.get("mixed_ploidy", False)
+    if mixed_ploidy:
+        call_genotype_non_allele = ds[call_genotype] < -1
+    else:
+        call_genotype_non_allele = xr.full_like(ds[call_genotype], False, "b1")
+    new_ds = create_dataset(
+        {variables.call_genotype_non_allele: call_genotype_non_allele}
+    )
+    return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+def infer_call_ploidy(
+    ds: Dataset,
+    *,
+    call_genotype: Hashable = variables.call_genotype,
+    call_genotype_non_allele: Hashable = variables.call_genotype_non_allele,
+    merge: bool = True,
+) -> Dataset:
+    """Infer the ploidy of each call genotype based on the number of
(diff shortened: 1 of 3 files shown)


sgkit/stats/aggregation.py
```diff
@@ -546,3 +546,129 @@ def sample_stats(
...
+
+
+def infer_call_ploidy(
+    ds: Dataset,
+    *,
+    call_genotype_non_allele: Hashable = variables.call_genotype_non_allele,
+    merge: bool = True,
+) -> Dataset:
+    """Infer the ploidy of each genotype call.
+
```
jeromekelleher, line 558: It would be nice to give a little bit more detail here (one sentence) as to how it actually does this

sgkit/tests/test_aggregation.py
```diff
@@ -289,3 +292,64 @@ def test_sample_stats(precompute_variant_allele_count):
...
+    ds = get_dataset(
+        [
+            [[0, 0, 0, 0], [0, 1, -2, -2], [1, 0, 0, 0]],
+            [[-1, 0, -2, -2], [0, -1, -1, -1], [-1, -1, -1, -1]],
+        ],
+        n_ploidy=4,
+    )
+    # test as fixed ploidy
+    cp = infer_call_ploidy(ds).call_ploidy
+    np.testing.assert_equal(cp, np.array([[4, 4, 4], [4, 4, 4]]))
```
jeromekelleher, line 307: Just as a passing thing: I don't think you need to explicitly convert the second array to numpy, 
```python
  np.testing.assert_equal(cp, [[4, 4, 4], [4, 4, 4]])
```
also works.

jeromekelleher (approved):
LGTM, just needs a little bit more explanation on the docs side.

sgkit/tests/test_aggregation.py
timothymillar: It's actually throwing `ValueError: The truth value of an array ...` when used with `xr.DataArray` and `list`. `cp.values` would work but calling `np.array` is consistent with the other tests.

jeromekelleher: Ah, I didn't try with an xr array. My bad!

tomwhite:
> * I don't know what I'm missing to populate the "comment" attribute in the HTML REPR?

Try using `create_dataset` from `sgkit.utils` rather than calling the `Dataset` constructor directly. This was added in #420

timothymillar:
Thanks @jeromekelleher and @tomwhite 
I have added helper method `infer_non_alleles` to make things a little more consistent between mixed and fixed ploidy datasets.
This can result in nested use of `define_variable_if_absent` to define both `call_genotype_non_allele` and `call_ploidy` which might not be ideal but i think it's the most consistent approach.

jeromekelleher (approved):
LGTM

tomwhite:
@jeromekelleher is this ready to be merged?

jeromekelleher:
> @jeromekelleher is this ready to be merged?

Yep, sorry, this one fell off my stack. I've added the label

Issue #457, opened 2021-02-05
Is test_vcfzarr_to_zarr test flaky?


aktech:
There was a test failure here: test_vcfzarr_to_zarr: https://github.com/pystatgen/sgkit/runs/1835950760?check_suite_focus=true

It disappeared when I triggered the build again and I haven't been able to reproduce it locally.
cc @tomwhite 

Feel free to close it, if unable to reproduce.

tomwhite:
Thanks for reporting this @aktech. This is a duplicate of #388. I haven't seen it so often recently, but it appears to still be happening occasionally.

PR #476, opened 2021-03-04
Fix mypy errors with Xarray 0.17.0

### LINKED ISSUE #475: Fix mypy errors with Xarray 0.17.0
Xarray 0.17.0 declares types so some mypy ignore statements are no longer needed.

tomwhite:
Fixes #475

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,4 +26,4 @@ repos:
     hooks:
       - id: mypy
         args: ["--strict", "--show-error-codes"]
-        additional_dependencies: ["numpy", "xarray", "dask[array]", "scipy", "typing-extensions", "zarr", "numba", "dask-ml"]
+        additional_dependencies: ["numpy", "xarray>=0.17.0", "dask[array]", "scipy", "typing-extensions", "zarr", "numba", "dask-ml"]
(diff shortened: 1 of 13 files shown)


ravwojdyla:
@tomwhite re my comment during the meeting, I think in our case we ended up making a change in the `.pre-commit-config.yaml` to trigger cache invalidation. So in this case it would be adding sth like `xarray>=0.17.0`?

ravwojdyla:
@tomwhite this is the pre-commit cache key: https://github.com/pre-commit/action/blob/bd5a59564b23176961c230d5a8facdba087e2962/index.js#L53, so change in the hash of the `.pre-commit-config.yaml` will trigger invalidation. And in our case (sgkit) I think added `xarray>=0.17.0` to mypy's `additional_dependencies` is actually relatively sane (?)

ravwojdyla:
@tomwhite but if you don't want to add the `>=0.17.0`, as a "hack" you could also just change the order of the dependencies in the mypy `additional_dependencies`, and that would likely also invalidate the cache.

tomwhite:
Thanks @ravwojdyla! I've tried changing `additional_dependencies` to include `xarray>=0.17.0`.

ravwojdyla:
@tomwhite I think that helped, but coincidentally we have another case of #457 :)

tomwhite:
Merging this, since it fixes the underlying mypy issue. I'll look at the race condition separately.

Issue #482, opened 2021-03-08
Documentation build fails with latest Dask release (2021.3.0)


tomwhite:
```
>>>-------------------------------------------------------------------------
Exception in /home/runner/work/sgkit/sgkit/docs/getting_started.rst at block ending on line None
Specify :okexcept: as an option in the ipython:: block to suppress this message
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-55-0ac15a3c444c> in <module>
----> 1 sg.diversity(ds, merge=False)

~/work/sgkit/sgkit/sgkit/stats/popgen.py in diversity(ds, cohort_allele_count, merge)
     85         [1.        , 1.        ]])
     86     """
---> 87     ds = define_variable_if_absent(
     88         ds, variables.cohort_allele_count, cohort_allele_count, count_cohort_alleles
     89     )

~/work/sgkit/sgkit/sgkit/utils.py in define_variable_if_absent(ds, default_variable_name, variable_name, func)
    193             f"Variable '{variable_name}' with non-default name is missing and will not be automatically defined."
    194         )
--> 195     return func(ds)
    196 
    197 

~/work/sgkit/sgkit/sgkit/stats/aggregation.py in count_cohort_alleles(ds, call_allele_count, merge)
    303     AC, SC = da.asarray(ds.call_allele_count), da.asarray(ds.sample_cohort)
    304     n_cohorts = SC.max().compute() + 1  # 0-based indexing
--> 305     C = da.empty(n_cohorts, dtype=np.uint8)
    306 
    307     G = da.asarray(ds.call_genotype)

/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/dask/array/wrap.py in wrap_func_shape_as_first_arg(func, *args, **kwargs)
     63         )
     64 
---> 65     parsed = _parse_wrap_args(func, args, kwargs, shape)
     66     shape = parsed["shape"]
     67     dtype = parsed["dtype"]

/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/dask/array/wrap.py in _parse_wrap_args(func, args, kwargs, shape)
     33     dtype = np.dtype(dtype)
     34 
---> 35     chunks = normalize_chunks(chunks, shape, dtype=dtype)
     36 
     37     name = name or funcname(func) + "-" + tokenize(

[...]

PR #483, opened 2021-03-08
Pin to Dask 2021.02.0 while incompatibility with later versions is investigated

tomwhite:


tomwhite:
This will fix the main build and the documentation build (see #482).

ravwojdyla (approved):

jeromekelleher (commented):
Are we sure we want to pin in setup.cfg? This does more than just fix the CI builds, doesn't it?

tomwhite:
> Are we sure we want to pin in setup.cfg? This does more than just fix the CI builds, doesn't it?

You're right. I was hoping this pin would be short-lived, and certainly wouldn't get released.

I can just change the requirements file though for CI.

jeromekelleher:
> You're right. I was hoping this pin would be short-lived, and certainly wouldn't get released.

Pins have a habit of persisting, I'd rather keep it to a minimum, but it's up to you really.

jeromekelleher (approved):

jrbourbeau:
I'd be curious to know what changes in the Dask 2021.03.0 release impacted `sgkit` (apologies for any unexpected breakages)

tomwhite:
> I'd be curious to know what changes in the Dask 2021.03.0 release impacted `sgkit` (apologies for any unexpected breakages)

It was this Dask PR: https://github.com/dask/dask/pull/6738. I see the problem described in #482 with that commit, but not with the one before it.

More concretely, the following prints `array(3)` with that Dask PR, and just `3` with previous versions.

```python
import numpy as np
import xarray as xr
import dask.array as da
xr.DataArray.__module__ = "xarray" # simulate sphinx workaround, which has the effect of Dask not recognising it as xarray
data = np.arange(4)
da.asarray(xr.DataArray(data)).max().compute()
```

jrbourbeau:
Gotcha, thanks for the upstream PR in Dask!

Issue #388, opened 2020-11-17
Possible race condition in _concat_zarrs_optimized


tomwhite:
```
================================== FAILURES ===================================
_________________ test_vcfzarr_to_zarr[None-True-True-False] __________________

shared_datadir = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-runneradmin/pytest-0/test_vcfzarr_to_zarr_None_True1/data')
tmpdir = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_vcfzarr_to_zarr_None_True1')
grouped_by_contig = True, consolidated = True, has_variant_id = False
concat_algorithm = None

    @pytest.mark.parametrize(
        "grouped_by_contig, consolidated, has_variant_id",
        [
            (False, False, False),
            (False, False, True),
            (True, False, True),
            (True, True, False),
        ],
    )
    @pytest.mark.parametrize(
        "concat_algorithm",
        [None, "xarray_internal"],
    )
    def test_vcfzarr_to_zarr(
        shared_datadir,
        tmpdir,
        grouped_by_contig,
        consolidated,
        has_variant_id,
        concat_algorithm,
    ):
        if has_variant_id:
            fields = None
        else:
            fields = [
                "variants/CHROM",
                "variants/POS",
                "variants/REF",
                "variants/ALT",
                "calldata/GT",
                "samples",
            ]
    
        vcfzarr_path = create_vcfzarr(  # type: ignore[no-untyped-call]
            shared_datadir,
            tmpdir,
            fields=fields,
            grouped_by_contig=grouped_by_contig,
            consolidated=consolidated,
        )
    
        output = str(tmpdir / "vcf.zarr")
>       vcfzarr_to_zarr(
            vcfzarr_path,
            output,
            grouped_by_contig=grouped_by_contig,
            concat_algorithm=concat_algorithm,
            consolidated=consolidated,
        )

sgkit\tests\test_vcfzarr_reader.py:135: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sgkit\io\vcfzarr_reader.py:134: in vcfzarr_to_zarr
    _concat_zarrs_optimized(
sgkit\io\vcfzarr_reader.py:259: in _concat_zarrs_optimized
[...]

PR #486, opened 2021-03-08
Attempt to fix race condition setting Zarr array attributes

tomwhite:
See https://github.com/pystatgen/sgkit/issues/388#issuecomment-792962331

### DIFF
--- a/sgkit/io/vcfzarr_reader.py
+++ b/sgkit/io/vcfzarr_reader.py
@@ -2,6 +2,7 @@ import tempfile
 from pathlib import Path
 from typing import Any, Dict, Hashable, List, Optional, Tuple
 
+import dask
 import dask.array as da
 import xarray as xr
 import zarr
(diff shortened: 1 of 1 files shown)


sgkit/io/vcfzarr_reader.py
```diff
@@ -221,39 +222,107 @@ def _concat_zarrs_optimized(
 
     first_zarr_group = zarr_groups[0]
 
-    with zarr.open_group(str(output)) as output_zarr:
+    # create the top-level group
+    zarr.open_group(str(output), mode="w")
```
jeromekelleher, line 226: Is it intentional to open the zarr group twice, one with a context manager and one without?

jeromekelleher (approved):
LGTM, but I don't follow the dask-y details.

sgkit/io/vcfzarr_reader.py
tomwhite: Good question. The reason I did this is because the dask computation doesn't actually use the `output_zarr` variable. So the sequence of operations is: 1. create group, 2. store (rechunked) arrays within the group (including attributes for those arrays, which is the goal of this PR), 3. copy unchanged arrays and top-level metadata.

jeromekelleher: OK, sounds good. If this works I think we should try to get the logic down as some comments here. I can easily see someone coming along and "fixing" things. At least some comments saying "HERE BE DRAGONS!" or something.

tomwhite: Good idea. I have added more comments along those lines.

Issue #322, opened 2020-10-09
Use logo and colors in docs


hammer:
[sgkit-logo.zip](https://github.com/pystatgen/sgkit/files/5357313/sgkit-logo.zip)
[BrandGuide.pdf](https://github.com/pystatgen/sgkit/files/5357314/BrandGuide.pdf)

hammer:
While looking into how to customize Sphinx themes I noticed [Furo](https://github.com/pradyunsg/furo), a new theme for documentation that looks quite nice.

aktech:
I think we can start with putting this logo on the [GitHub organization](https://github.com/pystatgen).

hammer:
@aktech good call! Just put it up there. It's a little confusing given our org name is `pystatgen` but I'm into it.

jeromekelleher:
Looks great

hammer:
Another potential theme to consider https://github.com/bashtage/sphinx-material

tomwhite:
Fixed in #501

PR #488, opened 2021-03-09
Doc refresh

tomwhite:
I did a pass through the docs and fixed things that I noticed. I also added the logo from #322 (but not the background colour).

You can see the generated docs here: https://tomwhite.github.io/sgkit

One thing that would be good to get feedback on is the subdivision of methods on the [API page](https://tomwhite.github.io/sgkit/api.html#methods) into "Summary statistics", "Statistical genetics", and "Population genetics" - is this an improvement - or is there a better way of doing it? (Or perhaps there isn't a good way of categorising these functions?)

(I made lots of small commits, so I plan to squash before merging - after updating with any feedback.)


docs/contributing.rst
```diff
@@ -23,12 +23,15 @@ Discussion forums
 Conversation about *sgkit* happens in the following places:
 
 1.  `GitHub Issue Tracker`_: for discussions around new features or bugs
-2.  `Python for Statistical Genetics forum`_: for general discussion
+2.  `GitHub Discussions`_: for general discussion
```
jeromekelleher, line 26: I think it's nice to point people here for general support stuff, otherwise they'll use the issues. something like:

> General discussions and questions like "how do I do X" or "what's the best way to do Y"?

docs/user_guide.rst
```diff
@@ -7,20 +7,70 @@ User Guide
...
+    $ pip install 'sgkit[bgen]'
+
+To install sgkit with PLINK support::
+
+    $ pip install 'sgkit[plink]'
+
+To install sgkit with VCF support::
+
+    $ pip install 'sgkit[vcf]'
+
```
jeromekelleher, line 34: Question: do we have an 'sgkit[all]' that'll pull everything in? That'd be handy, and good to document.

jeromekelleher (approved):
Looks great, some minor comments.

docs/contributing.rst
tomwhite: Fixed

docs/user_guide.rst
tomwhite: We don't have an `sgkit[all]`.

tomwhite:
(I also removed the method subsections.)

PR #490, opened 2021-03-15
Upstream build failing with "cannot import name get_mapper from dask.…

### LINKED ISSUE #489: Upstream build failing with "cannot import name 'get_mapper' from 'dask.bytes.core'"
We should just import from fsspec like Dask does now.

tomwhite:
…bytes.core" #489

Fixes #489

### DIFF
--- a/sgkit/io/vcfzarr_reader.py
+++ b/sgkit/io/vcfzarr_reader.py
@@ -6,6 +6,7 @@ import dask
 import dask.array as da
 import xarray as xr
 import zarr
+from fsspec import get_mapper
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
LGTM

When did fsspec start supporting this? Should we pin a min version of fsspec accordingly, or is there any point?

tomwhite:
Thanks @jeromekelleher. I think fsspec has had it for a while, it's just that Dask had its own internal version, that we inherited from the work in #486. I don't think we need to pin a minimum version of fsspec.

Issue #478, opened 2021-03-04
VCF file zoo


hammer:
We should collect weird VCFs and ensure we parse them correctly.

It would be nice if this zoo had metadata about what's interesting about a file.

@jeromekelleher can you point to your GA4GH zoo with VEP VCFs?

jeromekelleher:
There's a handful of VCF files in here:  https://github.com/ga4gh/ga4gh-server/tree/master/tests/data/datasets/dataset1/variants

tomwhite:
Some previous work here: #432

PR #492, opened 2021-03-16
Add VCF exclude_fields parameter and fix test_all_fields #491

### LINKED ISSUE #491: VCF test_all_fields not testing all files
The test is not passing the `vcf_file` parameter, so the default file (sample.vcf.gz) is being used each time. Other test files are not being exercised in this test.

tomwhite:
Fixes #491.

This actually runs the tests against the other files. In doing so it exposed the fact that Number=G is not supported, so I added a way to exclude fields. There were a few other minor changes needed to get tests to pass.

I also removed two large files from the test, since they add ~120s to the test time. These would be better suited to testing in #478.


jeromekelleher (approved):
LGTM

PR #498, opened 2021-03-17
GPU Implementation of Pairwise calculation

### LINKED ISSUE #338: Add GPU support for pairwise distance function
We have (soon) the pairwise distance functionality via this: https://github.com/pystatgen/sgkit/pull/306, we need to make it use GPU capabilities for faster processing.

@eric-czech suggested the following API: https://github.com/pystatgen/sgkit/pull/306#issuecomment-705720794

aktech:
This is the ~first draft (WIP) of the~ GPU Implementation of the Pairwise distance calculation, its not perfect at the moment but the good news is it works. There are still some improvements required, I have tried to write them down below. 

Fixes #338

TODO:

- [x] Benchmark
- [x] Improve finding gridsize and blocksize numbers
- [x] Add documentation
- [x] Some tests for running on machine with GPU
- [x] ~Write and benchmark reduction Kernel~ (doesn't seems like its required, didn't made a huge difference)

Will share benchmarks as soon as I finish the implementation. To share a rough idea of the initial numbers I ran the `euclidean` distance on the MalariaGEN dataset and it took about 6 min 47 sec on 2 TITAN RTX GPU, I think this number can be further improved

cc @jeromekelleher


jeromekelleher (commented):
Looks like a great start @aktech!

I really like the GPU kernel code, super simple and clear.

hammer:
@quasiben @jacobtomlinson no action required from y'all but just an FYI that @aktech has gotten some good results for one of our workloads on the GPU!

quasiben:
Very cool!  @aktech, for comparison,  what is the the CPU time for this calculation ?  

Some other tools which may be of interest:
- [cupy.fuse](https://docs.cupy.dev/en/stable/reference/generated/cupy.fuse.html)
- [cupy.jit/vectorize](https://github.com/cupy/cupy/issues/4290) (still in development)

cc @pentschev in case you have thoughts as well

aktech:
@quasiben 
> Very cool! @aktech, for comparison, what is the the CPU time for this calculation ?

It was about 10-11 minutes on a 100 CPU cores (25 AWS machines with 16GB RAM and 4 Cores each)

> Some other tools which may be of interest:

Thanks for sharing, I'll take a look.

pentschev (commented):
I'm no Numba expert so I didn't immediately spot any large changes that could improve performance even further. However, I added a few comments to hopefully clarify a few things, but overall it's great to hear you've got good results with Numba!

sgkit/distance/metrics.py
```diff
@@ -172,3 +173,46 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+
+
+def euclidean_map_gpu(f, g):
+    # move input data to the device
+    d_a = cuda.to_device(f)
+    d_b = cuda.to_device(g)
+    # create output data on the device
+    out = np.zeros((f.shape[0], g.shape[0]))
+    #     print(f"out_shape: {out.shape}")
+    d_out = cuda.to_device(out)
```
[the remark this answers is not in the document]
aktech: I started with `cupy` initially, but there was something I don't recall that was not implemented in it, so it didn't work. But I think the code has changed a lot since then, so I guess that could work, thanks, I'll try that.

sgkit/distance/metrics.py
```diff
@@ -172,3 +173,46 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
     out[0] = value
+
+
+@cuda.jit(device=True)
+def _euclidean_distance(a, b):
+    square_sum = 0.0
+    for i in range(a.shape[0]):
+        if a[i] >= 0 and b[i] >= 0:
+            square_sum += (a[i] - b[i]) ** 2
+    return square_sum
```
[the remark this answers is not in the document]
aktech: So, we have implemented the distance metrics via map reduce to be able to run this on very large datasets and this is the map step of the algorithm, in the reduction step we do the square root of the squared sum.

sgkit/distance/metrics.py
```diff
@@ -172,3 +173,46 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+
+
+@cuda.jit
+def euclidean_map_kernel(x, y, out) -> None:
+    i1, i2 = cuda.grid(2)
+    if i1 >= x.shape[0] or i2 >= y.shape[0]:
```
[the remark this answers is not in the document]
aktech: Yep makes sense. 👍

sgkit/distance/metrics.py
```diff
@@ -172,3 +173,46 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    d_a = cuda.to_device(f)
+    d_b = cuda.to_device(g)
+    # create output data on the device
+    out = np.zeros((f.shape[0], g.shape[0]))
+    #     print(f"out_shape: {out.shape}")
+    d_out = cuda.to_device(out)
+    #     d_out = cuda.device_array_like(d_a)
+
+    blocks_per_grid = (32, 32)
+    threads_per_block = (32, 32)
```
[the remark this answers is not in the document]
aktech: Yep, indeed makes sense I have this todo for the PR. :)

leofang:
Hi, a CuPy developer here. The title "pairwise" kinda draw my attention. 

I worked with a few people who needed a very performant pair-counting kernel, which is for astrophysics I think and its detail is probably irrelevant to this PR. What we found with Numba kernels is that generally you need to be aggressively making all variables explicitly typed, or the performance would lose to a pure CUDA implementation by a consistent factor ~3x (which I also see in my own projects), see https://github.com/numba/numba/issues/4647 for a detailed report and discussion. Sometimes, even with that the performance could still be slightly suboptimal due to the Numba's compiler internal. AFAIK it's hard to fix.

There's a guidance written by the current `@numba.cuda.jit` maintainer for how to mitigate this, see https://numba.discourse.group/t/question-math-typing-in-cuda/215. Just FYI, in case you're interested in benchmarking against a pure CUDA implementation (which can be loaded via `cupy.RawKernel` or `cupy.RawModule` 😄).

aktech:
I have updated the PR with some improvements to implementation and adding a kernel for correlation metric and tests as well. I will share a notebook soon (maybe by tomorrow), here is the summary of the state of things:

- Currently the euclidean metric is quite fast on GPU compared to CPU, It takes about 6 minutes on 2 TITAN RTX GPU compared to 10-11 minutes on a 100 CPU cores (25 AWS machines with 16GB RAM and 4 Cores each) for full MalariaGEN data
- Correlation metric is relatively quite slow, but still faster than CPU, for instance it took about 4 min on 2 GPU compared to 9 min on 48 Core CPUs for about 250 MB array data.

- This PR requires the dev release of numba, due to the following numba issue:
https://github.com/numba/numba/issues/6824 This can be installed via numba conda channel:

```
conda install -c numba/label/dev numba
```

Also the current version of numba will not work on distributed Cluster due to the following issues:
https://github.com/dask/distributed/issues/4597
https://github.com/numba/numba/issues/6821

## Bottleneck:
- The main bottleneck as far as I can see is copying things things to the device (GPU). This bottleneck is more pronounced in the `correlation` metric. There are couple of options that could potentially reduce this bottleneck:

On the original array do the following:

```python
x.map_blocks(cuda.to_device)
```
This raises an issue of array not being contiguous, so I tried the following:

```python
x = x.map_blocks(np.ascontiguousarray)
x = x.map_blocks(cuda.to_device)
```

Using the above changes something in the dask graph in a weird way, the root cause of which I have not been able to traceback, the error is here: https://gist.github.com/aktech/f90d3632e9f60d5d119704a98f2de8f7

Another alternative is to use `cupy` as following:

```python
x = x.map_blocks(cupy.asarray)
```

This also gives the same error as mentioned above. Also for cupy to work we need latest version of dask, unless we get rid of `da.triu`, due to this issue: https://github.com/dask/dask/issues/7452

## Way forward

So, the way forward from here is me polishing this PR a bit more with the current implementation based on incoming reviews and adding a basic benchmark notebook, for people to test things. I suggest we should review and merge this and then focus on improving the performance in a separate PR also if we plan to use cupy in it we would need a conda `environment.yml` file for installing dependencies on local and potential CI system so its better to keep that in a separate PR, I think.

@jeromekelleher @ravwojdyla @tomwhite thoughts?

pentschev:
> Another alternative is to use `cupy` as following:
> 
> ```python
> x = x.map_blocks(cupy.asarray)
> ```

For CuPy arrays you can simply do `da.from_array(x)`, it's also possible to optionally specify the `chunks=` kwarg.

On using Numba with Dask, [Stencil Computations with Numba](https://examples.dask.org/applications/stencils-with-numba.html) may be helpful, in case you haven't seen it yet.

aktech:
> For CuPy arrays you can simply do da.from_array(x), it's also possible to optionally specify the chunks= kwarg.

I have a dask array to start with, i.e. `x` is already a dask array with a specified chunking.

> On using Numba with Dask, Stencil Computations with Numba may be helpful, in case you haven't seen it yet.

Yep, I saw that it was a really good starting point.

aktech:
By the way, I found an interesting thing with the correlation kernel, the following change made it about 3X more performant.
    
I think this probably because the kernel thread doesn't have to access the device memory in each iteration of the loop to increment `out[i]`

```diff
diff --git a/sgkit/distance/metrics.py b/sgkit/distance/metrics.py
index aff539a..d117889 100644
--- a/sgkit/distance/metrics.py
+++ b/sgkit/distance/metrics.py
@@ -211,14 +211,30 @@ def call_metric_kernel(
 @cuda.jit(device=True)  # type: ignore
 def _correlation(x: ArrayLike, y: ArrayLike, out: ArrayLike) -> None:
     m = x.shape[0]
+    # Note: assigning variable and only saving the final value in the
+    # array made this significantly faster.
+    v0 = 0.0
+    v1 = 0.0
+    v2 = 0.0
+    v3 = 0.0
+    v4 = 0.0
+    v5 = 0.0
+
     for i in range(m):
         if x[i] >= 0 and y[i] >= 0:
-            out[0] += x[i]
-            out[1] += y[i]
-            out[2] += x[i] * x[i]
-            out[3] += y[i] * y[i]
-            out[4] += x[i] * y[i]
-            out[5] += 1
+            v0 += x[i]
+            v1 += y[i]
+            v2 += x[i] * x[i]
+            v3 += y[i] * y[i]
+            v4 += x[i] * y[i]
+            v5 += 1
+
+    out[0] = v0
+    out[1] = v1
+    out[2] = v2
+    out[3] = v3
+    out[4] = v4
+    out[5] = v5
```

aktech:
I have added the  notebook, which can be seen [here](https://nbviewer.jupyter.org/github/pystatgen/sgkit/blob/d2f769e8c78449124b09cc860d921385d50e788c/sgkit-gpu-malariagen.ipynb).

Here are the benchmarks:

## Benchmark
Data: MalariaGEN (**28.28 GB**)
GPU: 2 TITAN RTX

* Pairwise Distance: Euclidean Metric
Time Taken: **5min 44s**
Dask Report: [Link](https://rawcdn.githack.com/aktech/sgkit/8ce8b2f46e4ca60375d8f0101ed726f0cda5d4cb/dask-report-euclidean-full.html)


* Pairwise Distance: Correlation Metric
Time Taken: **10min 52s**
Dask Report: [Link](https://rawcdn.githack.com/aktech/sgkit/8ce8b2f46e4ca60375d8f0101ed726f0cda5d4cb/dask-report-correlation-full.html)

pentschev:
> I have a dask array to start with, i.e. `x` is already a dask array with a specified chunking.

Ah yes, sorry, my mistake, it's the other direction that you want. 🙂

aktech:
The CI is complaining because the coverage is not 100%, which is expected as the GPU tests are skipped.

aktech:
CI is green now.

aktech:
@ravwojdyla @jeromekelleher I think this is ready for review.

eric-czech:
Very nice @aktech! Apologies for the delay.

Regarding https://github.com/pystatgen/sgkit/pull/498#issuecomment-810381656, did you use the same VM instance type for the GPU experiments (i.e. 16GB RAM, 4 cores)?  I'd like to see that difference in $.

aktech:
> Regarding #498 (comment), did you use the same VM instance type for the GPU experiments (i.e. 16GB RAM, 4 cores)? I'd like to see that difference in $.

@eric-czech The above GPU experiment was not done on cloud. The GPUs I used were Quansight's on Premise: 2 x TITAN RTX 24 GB each (and 48 cores on the machine).


## GPU on Cloud (AWS) with Coiled
To compare costs and performance on cloud, I did another **GPU** experiment today with the same configuration as CPU:

- 25 Workers
- 16 GiB each
- 4 cores each

.. and here are the config and results on MalariaGEN data:

**Note**: I ran these on Coiled so I don't know the exact instance name, the following are taken from [AWS portal](https://aws.amazon.com/ec2/pricing/on-demand/) which matched the closest configuration used on coiled:


Instance Type | vCPU  | Memory (GiB) | Workers | Euclidean | Correlation | Cost
-- | -- | --| -- | -- | -- | --
CPU - t3.xlarge | 4  | 16 GiB | 25 | 10-11 min | -* | $0.1664 per Hour
GPU - g4dn.xlarge | 4 | 16 GiB | 25 | 46.6 sec | 1 min 44 sec | $0.526 per Hour


*I don't have the exact time for correlation on CPU, it was usually 2-3 times of time taken by euclidean on smaller dataset IIRC.

GPU Notebook is [here](https://nbviewer.jupyter.org/gist/aktech/7d4d0f1d3b9212f2bba20e6b86d5af75).

ravwojdyla:
@aktech awesome work! Regarding your last comment, do you happen to have the perf reports (+ the CPU notebook/reports)? Also I'm curious if you have run any experiments with chunked sample dimension?

eric-czech:
Thanks @aktech!  That's perfect.

jeromekelleher:
I think we should try to get this merged now - would it be possible to get some reviews?

@aktech - I think we should pull out the extra content from this PR like the notebooks etc, so that we can get it ready for merge. I'm not sure what the appropriate venue for the notebooks is though.

leofang:
For a potential future work, please kindly allow me to point you to CuPy's new JIT compiler, to appear in CuPy v9.0.0 later this moth: https://docs.cupy.dev/en/latest/tutorial/kernel.html#jit-kernel-definition. I think @aktech's implementation can be converted to a `@cupyx.jit.rawkernel()` with almost zero effort, as the syntax follows closely with CUDA's (and thus Numba's). However, we expect to see a significant performance gain, as CuPy JIT eliminates entirely the need of generating Python bytecodes and LLVM IRs in two pass. In addition, the painful need of explicit typing in Numba for squeezing out performance should be gone. If @aktech you ever have the bandwidth to give it a shot, we'll be very excited to hear your finding 🙂

aktech:
@ravwojdyla 
> @aktech awesome work! Regarding your last comment, do you happen to have the perf reports (+ the CPU notebook/reports)? 

Yep the reports and notebooks for CPU are here: https://github.com/pystatgen/sgkit/pull/447#issue-563315452

> Also I'm curious if you have run any experiments with chunked sample dimension?

I did some experiments in the past with sample axis chunked, when doing CPU experiments, here is an example: https://github.com/pystatgen/sgkit/issues/375#issuecomment-726048286 this includes about 10 chunks in the sample axis, but not on GPU on large data.

Here is some of the dump of some of the CPU experiments: https://iamit.in/work/sgkit/ (Note: this may not be very helpful as it is not properly organized, though it contains the report of a bunch of experiments with different chunk sizes).

I know this is a lot of data to consume, I feel like I'll probably have to write a blog post sometime to summarise results and learnings in an easily consumable form.

aktech:
> @aktech - I think we should pull out the extra content from this PR like the notebooks etc, so that we can get it ready for merge. I'm not sure what the appropriate venue for the notebooks is though.

@jeromekelleher Sure, as soon as it's approved. 👍 

Hey @leofang that's very exciting! I would definitely love to try that, I was trying to make use to `cupy` in general and had some trouble with using it with dask's blockwise and didn't manage to get it working, but I would like to try `@cupyx.jit.rawkernel()` as soon as it's out. Thanks a lot for sharing!

sgkit-gpu-malariagen.ipynb
```diff
@@ -0,0 +1,284 @@
+{
```
eric-czech, line 1: I would suggest moving this to a personal repo; I currently use https://github.com/eric-czech/sgkit-dev.  That's where I've been putting benchmark notebooks like this and it's a better place for the `dask-report*` files too.

sgkit/distance/api.py
```diff
@@ -60,6 +62,8 @@ def pairwise_distance(
 
         Omit to let dask heuristically decide a good default. A default can
         also be set globally with the split_every key in dask.config.
+    target
```
eric-czech, line 65: I think `target` was a good name for this in the context of numba given what it does, but I think `device` is better at our API level.

aktech:
For anyone reviewing this PR, you would need the latest version of Numba, see this: https://github.com/pystatgen/sgkit/pull/498#issuecomment-810381656 cc @eric-czech @jeromekelleher

sgkit/distance/api.py
```diff
@@ -94,12 +98,21 @@ def pairwise_distance(
            [ 2.62956526e-01,  0.00000000e+00,  2.14285714e-01],
            [ 2.82353505e-03,  2.14285714e-01,  0.00000000e+00]])
     """
+    valid_targets = TargetTypes.__args__  # type: ignore[attr-defined]
```
eric-czech, line 101: I like this but it's unprecedented within the rest of the codebase AFAIK.  I'm weakly in favor of not doing checks like this and relying on mypy instead since that's what we do in say https://github.com/pystatgen/sgkit/blob/master/sgkit/stats/aggregation.py (see `Dimension`).  I think it is optimistic to expect that any of our users care about running mypy on their code, but we should pick a consistent idiom regardless (and add an issue to use this kind of check more often if we go this route).

Do you have any thoughts on this @ravwojdyla?

sgkit/distance/api.py
```diff
@@ -110,17 +123,31 @@ def pairwise_distance(
     # graph serialisation/deserialisation time significantly
     metric_param = np.empty(n_map_param, dtype=x.dtype)
 
-    def _pairwise(f: ArrayLike, g: ArrayLike) -> ArrayLike:
-        result: ArrayLike = metric_map_func(f[:, None, :], g, metric_param)
+    def _pairwise_cpu(f: ArrayLike, g: ArrayLike) -> ArrayLike:
+        # This getattr is not required when the following issues are fixed and released
```
eric-czech, line 127: Can you add this as an issue please so we can make sure to address it or explicitly decide to ignore it at some point?

sgkit/distance/api.py
```diff
@@ -135,7 +162,7 @@ def _aggregate(x_chunk: ArrayLike, **_: typing.Any) -> ArrayLike:
         producing the final output. It is always invoked, even when the reduced
         Array counts a single chunk along the reduced axes."""
         x_chunk = x_chunk.reshape(x_chunk.shape[:-2] + (-1, n_map_param))
```
eric-czech, line 164: Can you add some comments on what the `x_chunk` dimensions are coming into this function?

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,234 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    g = np.ascontiguousarray(g)
+
+    # move input data to the device
+    d_a = cuda.to_device(f)
+    d_b = cuda.to_device(g)
+    # create output data on the device
+    out = np.zeros((f.shape[0], g.shape[0], N_MAP_PARAM[metric]), dtype=f.dtype)
+    d_out = cuda.to_device(out)
+
+    # TODO: Consider using cupy for directly creating zeros on GPU
```
eric-czech, line 196: Can you add an issue for this?

sgkit/distance/api.py
eric-czech: FYI @jeromekelleher on dispatching, I think by accepting this we're effectively throwing our hands up at that problem and leaving it up to users with this solution (ideally they could define arrays on gpu backends with no further code changes).  I'm cool with that for now but I wanted to point it out to you as well.

eric-czech (changes requested):
I had a few minor suggestions so far and I'm submitting this review partially done now because something in the github UI crashed and it won't let me add more comments (and I don't want to lose pending ones).  I have a few more forthcoming.

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,234 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    # d_out = cp.zeros((f.shape[0], g.shape[0], N_MAP_PARAM[metric]), dtype=f.dtype)
+
+    threads_per_block = (32, 32)
```
eric-czech, line 199: How good do you feel about this as a default?  Did this come from somewhere else and do you think is it likely to matter much on different hardware?

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,234 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    )
+
+    metric_kernel[blocks_per_grid, threads_per_block](d_a, d_b, d_out)
+    # copy the output array back to the host system
+    d_out_host = d_out.copy_to_host()
+    return d_out_host
+
+
+@cuda.jit(device=True)  # type: ignore
+def _correlation(
```
eric-czech, line 212: This is a great reference function to have -- thanks for taking the time to optimize it.  How did you optimize it by the way?  Quick local benchmarks with small array inputs?

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,234 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    x: ArrayLike, y: ArrayLike, out: ArrayLike
+) -> None:  # pragma: no cover.
+    # Note: assigning variable and only saving the final value in the
+    # array made this significantly faster.
+    v0 = types.float32(0)
```
eric-czech, line 217: A somewhat unspoken rule within most sgkit functions is that 32-bit input -> 32-bit output and 64-bit input -> 64-bit output (with no other dtypes supported).  Do any of the pairwise functions support this? It looks like they won't, so we should probably make a note that all GPU arithmetic is going to be 32-bit in the docs for the functions.

sgkit/tests/test_distance.py
```diff
@@ -64,36 +77,44 @@ def create_distance_matrix(
 
 
 @pytest.mark.parametrize(
-    "size, chunk",
+    "size, chunk, target",
     [
-        ((100, 100), (20, 10)),
-        ((100, 100), (25, 10)),
-        ((100, 100), (50, 10)),
+        ((100, 100), (20, 10), "cpu"),
```
eric-czech, line 82: Given the specificity of the map/reduce funcs to chunkwise operations, I'd feel a little better about a test grid like this here:

```
((30, 30), (10, 10), "cpu"),
((30, 30), (10,  5), "cpu"),
((30, 30), ( 5, 10), "cpu")),
((15, 30), (10, 10), "cpu"),
((15, 30), (10,  5), "cpu"),
((15, 30), ( 5, 10), "cpu")),
((30, 15), (10, 10), "cpu"),
((30, 15), (10,  5), "cpu"),
((30, 15), ( 5, 10), "cpu")),
```

And the same for a gpu device.

I'm assuming that will still be quite fast.

sgkit/tests/test_distance.py
```diff
@@ -64,36 +77,44 @@ def create_distance_matrix(
...
+        ((100, 100), (25, 10), "cpu"),
+        ((100, 100), (50, 10), "cpu"),
+        pytest.param((100, 100), (25, 10), "gpu", marks=pytest.mark.gpu),
+        pytest.param((100, 100), (50, 10), "gpu", marks=pytest.mark.gpu),
+        pytest.param((100, 100), (50, 10), "gpu", marks=pytest.mark.gpu),
```
eric-czech, line 109: Duplicate params here.  I'd suggest the larger grid from above instead.

sgkit/distance/metrics.py
eric-czech: Nm I see `test_data_types` now!

sgkit/tests/test_distance.py
```diff
@@ -3,17 +3,30 @@
...
+    except cuda.CudaSupportError:
+        return False
+
+
+def skip_gpu_tests_if_no_gpu(target: TargetTypes) -> None:
+    if target == "gpu" and not detect_cuda_driver():
+        pytest.skip("Cuda driver not found")
+
+
 def get_vectors(
```
eric-czech, line 30: Can this not be more like this?

```
def get_vectors(
    array_type: str = "da",
    dtype: str = "i8",
    size= (100, 100),
    chunk=(20, 10),
):
    rs = da.random.RandomState(0)
    x = rs.randint(0, 3, size=size, chunks=chunk, dtype=dtype)
    return x if array_type == "da" else np.asarray(x)
```

It makes the tests harder to follow IMO when the dask version is returning ints in [0, 2] and the numpy arrays are floats in [0, 1].

sgkit/tests/test_distance.py
```diff
@@ -127,19 +148,25 @@ def test_distance_ndarray() -> None:
...
 def test_missing_values(
     metric: MetricTypes,
     metric_func: typing.Callable[[ArrayLike, ArrayLike], np.float64],
     dtype: str,
+    target: TargetTypes,
 ) -> None:
+    skip_gpu_tests_if_no_gpu(target)
     x = get_vectors(array_type="np", dtype=dtype)
 
     ri_times = np.random.randint(5, 20)
```
eric-czech, line 172: I think something like this here would be more readable:

```python
x = np.random.choice(range(-3, 3), size=(30, 30))
if np.issubdtype(dtype, np.floating):
    x = np.where(x < 0, np.nan, x)
```

eric-czech:
I left a few more suggestions but overall looks good!  Thanks again for seeing it through all these permutations.

Remind me again what I'm looking for in running this myself?  Do we have any idea if these tests run on a GPU?

aktech:
> Remind me again what I'm looking for in running this myself? Do we have any idea if these tests run on a GPU?

You would need to have a machine which has a GPU and is [detectable](https://github.com/pystatgen/sgkit/pull/498/files#diff-bfebf015a7d0a911d66743edcf9806a15b4661814667c0683fef0af8a6222b35R18) and run the tests as you would for anything: 
```
pytest -v --cov=sgkit --cov-report=html sgkit/tests/test_distance.py
```

and check if the GPU tests are passing, if the tests are unable to detect GPU then it will skip them. I tested these on my machine (with GPU) and it works.

Also, thanks for the review. I am away until Monday. I'll get back to them next week.

eric-czech:
> You would need to have a machine which has a GPU and is detectable and run the tests as you would for anything
> I tested these on my machine (with GPU) and it works.

I meant more like what am I looking for if it works for you?

aktech:
> I meant more like what am I looking for if it works for you?

Just a validation/sanity check basically, that the tests works on someone else's GPU as well apart from mine.

jakirkham:
It would be good to capture this in a blogpost at some point if anyone here has interest. Alistair started some work on this previously in PR ( https://github.com/dask/dask-blog/pull/50 ), which is likely very close to being ready

jakirkham:
cc @gmarkall (in case this is of interest 🙂)

aktech:
> It would be good to capture this in a blogpost at some point if anyone here has interest. Alistair started some work on this previously in PR ( dask/dask-blog#50 ), which is likely very close to being ready

Indeed, there is a lot to consume here. I was thinking about the same earlier https://github.com/pystatgen/sgkit/pull/498#issuecomment-817881279

gmarkall:
> cc @gmarkall (in case this is of interest slightly_smiling_face)

Thanks! I'd been talking to @aktech about it already, but perhaps I should have written to him here rather than in a private chat.

ravwojdyla:
@eric-czech for some reason GH doesn't allow me to leave a comment under https://github.com/pystatgen/sgkit/pull/498#discussion_r614206050, so I will reply here. I agree, would be nice to have a single convention, and I'm +1 to having an issue + dev doc. On the check itself, I actually am +1 to the current implementation from @aktech, since it covers both mypy and non-mypy uses in a concise way.

eric-czech:
> I agree, would be nice to have a single convention, and I'm +1 to having an issue + dev doc. On the check itself, I actually am +1 to the current implementation from @aktech, since it covers both mypy and non-mypy uses in a concise way.

Cool so https://github.com/pystatgen/sgkit/pull/498#discussion_r614206050 can stay as is @aktech but can you add an issue describing that check and stating that we should be doing it everywhere?

sgkit/distance/api.py
aktech: One of the motivations of not using device was: `device` is also a synonym of GPU, isn't it?

tomwhite: The new Python array API uses "device" in this context: https://data-apis.org/array-api/latest/design_topics/device_support.html#device-support. So it seems like the right term to be using here.

aktech: Ok, great. Sounds good.

sgkit/distance/api.py
aktech: Numba released a bugfix version 0.53.1, we don't need this workaround anymore.

sgkit/distance/metrics.py
aktech: > How did you optimize it by the way? Quick local benchmarks with small array inputs?

Yep, mostly like that, with small arrays.

sgkit/distance/metrics.py
aktech: I'll remove this Todo, this was a note just for myself to try cupy.

jeromekelleher:
@aktech, can you ping me when you're finished with @eric-czech's comments so I can do my review please?

sgkit/distance/metrics.py
aktech: Quite good I would say, this is the maximum number of threads you can have in a block 

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#features-and-technical-specifications__technical-specifications-per-compute-capability

These apply to compute capability 2.0 and higher and all GPUs NVIDIA has shipped in the past 10+ years have compute capability > 3.0 as far as I know.

One way to get compute capability programmatically is: 

```python
In [1]: from numba import cuda
In [2]: device = cuda.get_current_device()
In [3]: device.compute_capability
Out[3]: (7, 5)
```

The above is for the Quansight GPU, I was working on, do you think I should add this check in the code rather?

eric-czech: Nope, no need for the check as far as I'm concerned.  Thanks for those details!

All set on this point.

aktech:
@eric-czech I think I have addressed all the comments, except the following:

```diff

-    rs = da.random.RandomState(0)

-    x = rs.randint(0, 3, size=size, chunks=chunk).astype(dtype)

-    return x if array_type == "da" else np.asarray(x)

+    if array_type == "da":

+        rs = da.random.RandomState(0)

+        x = rs.randint(0, 3, size=size).astype(dtype).rechunk(chunk)

+    else:

+        x = np.random.rand(size[0], size[1]).astype(dtype)

+    return x



-    if array_type == "da":

-        rs = da.random.RandomState(0)

-        x = rs.randint(0, 3, size=size).astype(dtype).rechunk(chunk)

-    else:

-        x = np.random.rand(size[0], size[1]).astype(dtype)

-    return x

+    rs = da.random.RandomState(0)

+    x = rs.randint(0, 3, size=size, chunks=chunk).astype(dtype)

+    return x if array_type == "da" else np.asarray(x)
```

This seems to fail on Windows, I don't have a windows machine to debug at
the moment, let me know if it is possible to address this later.
https://github.com/pystatgen/sgkit/runs/2419912014?check_suite_focus=true

This PR has become non-responsive on the browser, maybe due to its size (or
maybe my bandwidth is low at the moment). I am trying to reply via email
instead.

cc @jeromekelleher

aktech:
@eric-czech I think I have addressed all the comments, except the following:

```diff
-    rs = da.random.RandomState(0)
-    x = rs.randint(0, 3, size=size, chunks=chunk).astype(dtype)
-    return x if array_type == "da" else np.asarray(x)
+    if array_type == "da":
+        rs = da.random.RandomState(0)
+        x = rs.randint(0, 3, size=size).astype(dtype).rechunk(chunk)
+    else:
+        x = np.random.rand(size[0], size[1]).astype(dtype)
+    return x


-    if array_type == "da":
-        rs = da.random.RandomState(0)
-        x = rs.randint(0, 3, size=size).astype(dtype).rechunk(chunk)
-    else:
-        x = np.random.rand(size[0], size[1]).astype(dtype)
-    return x
+    rs = da.random.RandomState(0)
+    x = rs.randint(0, 3, size=size, chunks=chunk).astype(dtype)
+    return x if array_type == "da" else np.asarray(x)
```

This seems to fail on Windows, I don't have a windows machine to debug at
the moment, let me know if it is possible to address this later.
https://github.com/pystatgen/sgkit/runs/2419912014?check_suite_focus=true

~This PR has become non-responsive on the browser, maybe due to its size (or
maybe my bandwidth is low at the moment). I am trying to reply via email
instead. (Update: Seems like my internet is slow)~

EDIT: Reposted this comment and deleted the old comment as the diff was not rendering properly.

cc @jeromekelleher

sgkit/distance/api.py
```diff
@@ -133,9 +154,19 @@ def _pairwise(f: ArrayLike, g: ArrayLike) -> ArrayLike:
     def _aggregate(x_chunk: ArrayLike, **_: typing.Any) -> ArrayLike:
         """Last function to be executed when resolving the dask graph,
         producing the final output. It is always invoked, even when the reduced
-        Array counts a single chunk along the reduced axes."""
+        Array counts a single chunk along the reduced axes.
+
+        Parameters
+        ----------
+        x
```
eric-czech, line 161:
```suggestion
        x_chunk
```

sgkit/tests/test_distance.py
eric-czech: What happened with this one?  I don't see a resolution for it.

sgkit/tests/test_distance.py
eric-czech: I don't see changes for this one.

eric-czech:
Hey @aktech, I didn't see any changes for a few of them.  On that Windows failure:

> This seems to fail on Windows, I don't have a windows machine to debug atthe moment, let me know if it is possible to address this later.

Ok but at a minimum we should have an issue describing it as well as a change to set the numpy random seed in addition to the dask random seed.

sgkit/tests/test_distance.py
aktech: I see I posted the same diff twice in here: https://github.com/pystatgen/sgkit/pull/498#issuecomment-825735250

This suggestion was the second one, which broken on windows.

sgkit/distance/api.py
```diff
@@ -8,12 +8,14 @@
 from sgkit.typing import ArrayLike
 
 MetricTypes = Literal["euclidean", "correlation"]
+DeviceTypes = Literal["cpu", "gpu"]
 
 
 def pairwise_distance(
```
eric-czech, line 14: Can you also add this to the top-level API in https://github.com/pystatgen/sgkit/blob/master/sgkit/__init__.py?

sgkit/tests/test_distance.py
aktech: I added this in here: https://github.com/pystatgen/sgkit/pull/498/files#diff-bfebf015a7d0a911d66743edcf9806a15b4661814667c0683fef0af8a6222b35R104

Probably a mistake, you requested in the correlation test and I added in the euclidean test, I think there is value in adding it in both, I'll add.

eric-czech: I see, can you please collapse both of them into something like this?

```python
@pytest.mark.parametrize(
    "size, chunk, device, metric",
    [
        pytest.param(shape, chunks, device, metric, marks=pytest.mark.gpu if device == "gpu" else None)
        for shape in [(30, 30), (15, 30), (30, 15)]
        for chunks in [(10, 10), (5, 10), (10, 5)]
        for device in ["cpu", "gpu"]
        for metric in ["euclidean", "correlation"]
    ]
)
def test_pairwise(size, chunk, device, metric):
    ... # create `x`
    actual = pairwise_distance(x, metric=metric, device=device)
    expected = pdist(x, metric=metric)
    np.testing.assert_almost_equal(actual, expected)
```

aktech: Makes sense, done thanks!

aktech:
> Hey @aktech, I didn't see any changes for a few of them. On that Windows failure:

I responded to those comments, I updated the diff here in this comment https://github.com/pystatgen/sgkit/pull/498#issuecomment-825735250 (I had accidentally pasted the same diff twice) to reflect the second suggestion for missing values.

eric-czech:
> I responded to those comments, I updated the diff here in this comment #498 (comment) (I had accidentally pasted the same diff twice) to reflect the second suggestion for missing values.

Ok last thing then would be to add an issue for it and then add the random seed set for numpy too, i.e. 

```
    if array_type == "da":
         rs = da.random.RandomState(0)
         x = rs.randint(0, 3, size=size).astype(dtype).rechunk(chunk)
    else:
+       rs = np.random.RandomState(0)
         x = rs.rand(size[0], size[1]).astype(dtype)
```

After that, this is all set from my end!

aktech:
Created the issue for the same here: https://github.com/pystatgen/sgkit/issues/543 and made the suggested change.

eric-czech (approved):
Great work @aktech!  It's all yours @jeromekelleher.

sgkit/distance/api.py
```diff
@@ -60,6 +62,8 @@ def pairwise_distance(
 
         Omit to let dask heuristically decide a good default. A default can
         also be set globally with the split_every key in dask.config.
+    device
+        The architecture to run the calculation on, either of cpu or gpu
```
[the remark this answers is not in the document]
jeromekelleher: What happens if I call ``device="gpu"`` and I don't have a GPU? We should define the behaviour here. Maybe we should raise a custom error class, so that this can be handled programatically? (So, client code could try to run on the GPU, and if that fails, run on CPU)?

jeromekelleher (changes requested):
This is awesome, thanks @aktech! I have some very minor documentation suggestions.

The only real issue is handling the failure of GPU calculations gracefully, I think.

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,231 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    zero = types.uint32(0)
+
+    while i < m:
+        if x[i] >= zero and y[i] >= zero:
+            v0 += x[i]
+            v1 += y[i]
+            v2 += x[i] * x[i]
+            v3 += y[i] * y[i]
+            v4 += x[i] * y[i]
+            v5 += 1
```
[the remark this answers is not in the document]
aktech: Aggresive typecasting of all the variables is done to improve performance. See this: https://github.com/pystatgen/sgkit/pull/498#issuecomment-804650369

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,231 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+    of the given pair of vectors.
+    """
+    square_sum = types.float32(0)
+
+    zero = types.uint32(0)
+    a_shape_0 = types.uint32(a.shape[types.uint32(0)])
+    i = types.uint32(0)
+
+    while i < a_shape_0:
+        if a[i] >= zero and b[i] >= zero:
```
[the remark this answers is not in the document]
aktech: Same as this https://github.com/pystatgen/sgkit/pull/498/files#r626468971

Sounds good, I think.

sgkit/distance/metrics.py
```diff
@@ -172,3 +175,231 @@ def correlation_reduce(v: ArrayLike, out: ArrayLike) -> None:  # pragma: no cove
...
+            square_sum += (a[i] - b[i]) ** types.uint32(2)
```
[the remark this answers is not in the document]
aktech: This was not intentional, it's just what I happened to use.
To be honest, I don't know the the root cause of why it's more performant to explicitly type everything apart from of course the obvious reason, that the compiler won't have to waste time to infer types.

sgkit/distance/api.py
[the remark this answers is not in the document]
aktech: It would return `CudaSupportError` maybe we can catch that and return a custom sgkit `CUDANotFoundError` maybe?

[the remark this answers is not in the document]
jeromekelleher: CudaSupportError is fine, that's a comprehensible error. We can add a local error later if it seems needed. Thanks.

sgkit/distance/metrics.py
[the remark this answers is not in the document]
jeromekelleher: So it makes sense that we make sure that variables are **declared** to have specific types, but I don't see why the compiler needs this extra info each time we *update* the variable. We've told it that ``i`` is a uint32 (which is good), but it seems pretty weird to have to *keep* telling it that it's a uint32. If we apply this pattern consistently, we'll have very ugly code: for example, why not
```
v0 = types.float32(v0 + x[i])
...
v5 = types.float32(v5 + types.float32(1))
```
I don't see why we're making a special case for the index loop variable here, which should be very easy for the compiler to know "remains" a uint32, unless the literal ``1`` is always upcast to a float64 (which would be weird).

sgkit/distance/metrics.py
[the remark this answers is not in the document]
jeromekelleher: I don't mind if the compiler takes a tiny bit longer to compile the code - that seems worth having code without these inline typecasts on variable updates. Unless it significantly affects the actual run-time performance I'd prefer we didn't have them.

jeromekelleher (commented):
I'd like to get to the bottom of whether these inline variable-updating typecasts actually make any difference @aktech. Having clean code is and important goal of the project so we should try to keep the code Pythonic except where there's a compelling perf motivation. We're setting the template for later work here in this PR, so I want to make sure that the assumptions we're making and patterns we're establishing are good.

sgkit/distance/metrics.py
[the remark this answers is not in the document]
leofang: It is not about compilation time, it's about runtime performance.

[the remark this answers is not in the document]
jeromekelleher: Thanks @leofang - I'd like to understand what's going on here so I appreciate you taking the time to jump in here. I can see how we're helping out the compiler here to make good code in this case because it's an exponentiation. I don't understand the loop increment above though.

So, the question is if ``i`` is declared as a uint32, why we need to write ``i = types.uint32(i + types.uint32(1))`` rather than ``i += 1``?

I'm concerned because the patterns established in the PR will be copied in subsequent GPU kernels, and it does make the code quite a lot less easy to read. Declaring variables as particular types is totally fine, but  ``i = types.uint32(i + types.uint32(1))`` obscures the meaning of the code and makes mistakes much more likely.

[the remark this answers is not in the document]
gmarkall: Numba infers the type of integer constants as being pointer-sized, i.e. 64-bit on most systems. This is point 2 in this section of the documentation: https://numba.readthedocs.io/en/latest/proposals/integer-typing.html#proposal-predictable-width-conserving-typing - the rationale for this typing is outlined in the document, but it can be summarised as "choose types that makes the typing behaviour seem as consistent and predictable as possible to the end user" - the previous ethos was "start small, and grow as required".

The impact of the current behaviour on the CUDA target is that judicious use of explicit types is needed, as in this line - if it were just `square_sum += (a[i] - b[i]) ** 2`, then `2` would be inferred as an `int64`, then the `**` would operate on `float32 ** int64`, which would then have both operands cast to use the `float64 ** float64` implementation of `**`, and then that would force `square_sum` to have the `float64` type everywhere, which is what slows things down.

[This document](https://github.com/numba/nvidia-cuda-tutorial/blob/main/session-2/session-2-with-notes.pdf) contains various examples and explanations relevant to the CUDA target.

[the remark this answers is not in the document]
gmarkall: > So, the question is if `i` is declared as a uint32, why we need to write `i = types.uint32(i + types.uint32(1))` rather than `i += 1`?

It's not *declared* as a `uint32` - `i` refers to a `uint32` value at that point in the function, but if a value of a different type is stored in `i` somewhere else, then Numba has to unify the set of different types that `i` can store. This usually means using the biggest type that is used, or possibly something bigger than any of them - for example, if it stored an `int32` in one location and a `float32` on another, then the resulting unified type would be `float64`, because Numba tries to avoid losing precision (in this case, a `float32` doesn't have enough precision to represent all the digits of a large `int32` value).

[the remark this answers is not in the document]
jeromekelleher: Great, thanks @gmarkall, that makes a lot of sense and the reference material is super helpful. So, in the loop counter example, if we wrote
```
i = types.uint32(0)
while stuff:
    i += 1
```
``i`` would be promoted to int64, because is 1 an int64 even though we've explicitly declared it to be a uint32?

So, then it should be sufficient to write
```
i = types.uint32(0)
while stuff:
    i += types.uint32(1)
```
and ``i = types.uint32(i + types.uint32(1))`` is excessive?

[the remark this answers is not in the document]
gmarkall: > `i` would be promoted to int64, because is 1 an int64 even though we've explicitly declared `i` to be a uint32?

The code doesn't *explicitly declare* anything about `i` - it assigns it a `uint32` value on the first line. The type of the value it holds on the first line doesn't constrain the types of the values that might be assigned to it later. (for completeness / correctness, I would note though that using incompatible types can result in a unification error)



> So, then it should be sufficient to write
> 
> ```
> i = types.uint32(0)
> while stuff:
>     i += types.uint32(1)
> ```
> 
> and `i = types.uint32(i + types.uint32(1))` is excessive?

Unfortunately rule 3 of https://numba.readthedocs.io/en/latest/proposals/integer-typing.html#proposal-predictable-width-conserving-typing requires that you write `i = types.uint32(i + types.uint32(1))` - if you just write `i += types.uint32(1)`, the `+=` on two `uint32` values results in a `uint64`.

As an example of how to investigate this sort of thing, if you run:

```python
from numba import njit, types

@njit
def f(n):
    i = types.uint32(0)
    while i < types.uint32(n):
        i += types.uint32(1)
    return i

# Call the function so we get a compiled version to inspect
f(10)

# Inspect the result of Numba's type inference
f.inspect_types()
```

you will get a lot of output (the entire Numba IR and associated typing will be dumped) but the relevant parts are:

* The cast of the literal `0` down to `uint32`:
  ```
  #   i = call $4load_method.1($const6.2, func=$4load_method.1, args=[Var($const6.2, nocast.py:5)], kws=(), vararg=None, target=None)  :: (int64,) -> uint32
  ```
  * The phi node where the two different `i` values meet at the top of the loop (a phi node selects a variable from two different control flow paths - in this case there's an incoming `i` value from just before the while loop, and one from the loop body). This is a `uint64`, because one of the incoming values was a `uint64`:
  ```
  #   i.2 = phi(incoming_values=[Var(i, nocast.py:5), Var(i.1, nocast.py:7)], incoming_blocks=[0, 26])  :: uint64
  ```
* The cast of the literal `1` down to `uint32`:
  ```
  #   $34call_method.4 = call $30load_method.2($const32.3, func=$30load_method.2, args=[Var($const32.3, nocast.py:7)], kws=(), vararg=None, target=None)  :: (int64,) -> uint32
  ```
* The addition, which is typed as a `uint64`:
  ```
  #   $36inplace_add.5 = inplace_binop(fn=<built-in function iadd>, immutable_fn=<built-in function add>, lhs=i.2, rhs=$34call_method.4, static_lhs=Undefined, static_rhs=Undefined)  :: uint64
  ```
* The assignment of the result of the addition (a `uint64`) to `i` on the third line of the function (which must therefore also be a `uint64`):
  ```
      #   i.1 = $36inplace_add.5  :: uint64
  ```

Numba's typing works by propagating type information through the function until there are no more updates. You can see each step of the inference by running the code with `NUMBA_DEBUG_TYPEINFER=1` - each step is delimited by a banner stating `--- propagate ---`. The Numba IR and the current typing is then dumped at each step. Looking at the first step can often help pinpoint where widening is coming from. For the example above, the IR and first step look like:

```
$ NUMBA_DEBUG_TYPEINFER=1 python nocast.py 
---------------------------------SSA IR DUMP: f---------------------------------
label 0:
    n = arg(0, name=n)                       ['n']
    $2load_global.0 = global(types: <module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>) ['$2load_global.0']
    $4load_method.1 = getattr(value=$2load_global.0, attr=uint32) ['$2load_global.0', '$4load_method.1']
    $const6.2 = const(int, 0)                ['$const6.2']
    i = call $4load_method.1($const6.2, func=$4load_method.1, args=[Var($const6.2, nocast.py:5)], kws=(), vararg=None, target=None) ['$4load_method.1', '$const6.2', 'i']
    jump 12                                  []
label 12:
    i.2 = phi(incoming_values=[Var(i, nocast.py:5), Var(i.1, nocast.py:7)], incoming_blocks=[0, 26]) ['i', 'i.1', 'i.2']
    $14load_global.1 = global(types: <module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>) ['$14load_global.1']
    $16load_method.2 = getattr(value=$14load_global.1, attr=uint32) ['$14load_global.1', '$16load_method.2']
    $20call_method.4 = call $16load_method.2(n, func=$16load_method.2, args=[Var(n, nocast.py:5)], kws=(), vararg=None, target=None) ['$16load_method.2', '$20call_method.4', 'n']
    $22compare_op.5 = i.2 < $20call_method.4 ['$20call_method.4', '$22compare_op.5', 'i.2']
    bool24 = global(bool: <class 'bool'>)    ['bool24']
    $24pred = call bool24($22compare_op.5, func=bool24, args=(Var($22compare_op.5, nocast.py:6),), kws=(), vararg=None, target=None) ['$22compare_op.5', '$24pred', 'bool24']
    branch $24pred, 26, 42                   ['$24pred']
label 26:
    $28load_global.1 = global(types: <module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>) ['$28load_global.1']
    $30load_method.2 = getattr(value=$28load_global.1, attr=uint32) ['$28load_global.1', '$30load_method.2']
    $const32.3 = const(int, 1)               ['$const32.3']
    $34call_method.4 = call $30load_method.2($const32.3, func=$30load_method.2, args=[Var($const32.3, nocast.py:7)], kws=(), vararg=None, target=None) ['$30load_method.2', '$34call_method.4', '$const32.3']
    $36inplace_add.5 = inplace_binop(fn=<built-in function iadd>, immutable_fn=<built-in function add>, lhs=i.2, rhs=$34call_method.4, static_lhs=Undefined, static_rhs=Undefined) ['$34call_method.4', '$36inplace_add.5', 'i.2']
    i.1 = $36inplace_add.5                   ['$36inplace_add.5', 'i.1']
    jump 12                                  []
label 42:
    $44return_value.1 = cast(value=i.2)      ['$44return_value.1', 'i.2']
    return $44return_value.1                 ['$44return_value.1']

-----------------------------------propagate------------------------------------
---- type variables ----
[$14load_global.1 := Module(<module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>),
 $16load_method.2 := class(uint32),
 $20call_method.4 := uint32,
 $22compare_op.5 := bool,
 $24pred := bool,
 $28load_global.1 := Module(<module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>),
 $2load_global.0 := Module(<module 'numba.core.types' from '/home/gmarkall/numbadev/numba/numba/core/types/__init__.py'>),
 $30load_method.2 := class(uint32),
 $34call_method.4 := uint32,
 $36inplace_add.5 := uint64,
 $44return_value.1 := uint32,
 $4load_method.1 := class(uint32),
 $const32.3 := Literal[int](1),
 $const6.2 := Literal[int](0),
 arg.n := int64,
 bool24 := Function(<class 'bool'>),
 i := uint32,
 i.1 := uint64,
 i.2 := uint32,
 n := int64]
```

Numba IR is in Static Single Assignment (SSA) form, where variables are only assigned to once - the original Python variables are suffixed with an index to convert them to SSA form, so we have multiple instances of `i` - `i` (the original assignment), `i.1` (the assignment inside the loop), and `i.2` (the value of `i` at the start of the loop body coming either from outside the loop, or the end of a previous iteration). We can see that `i` started out holding a `uint32` (`i := uint32,`), as did `i.2` (` i.2 := uint32,`), but `i.1` held a `uint64` (`i.1 := uint64,`) because it was assigned the `uint64` result of the addition (` $36inplace_add.5 := uint64,`). In a subsequent step the `uint64` type of `i.1` propagated into `i.2` through the control flow from the end of the loop back to the top of it:

```
-----------------------------------propagate------------------------------------
... <snip> ...
 'i': uint32,
 'i.1': uint64,
 'i.2': uint64,
```

Finally, having investigated all this, we fix the problem with the cast of the result of the addition, so our code reads:

```python
from numba import njit, types

@njit
def f(n):
    i = types.uint32(0)
    while i < types.uint32(n):
        i = types.uint32(i + types.uint32(1))
    return i

print(f(10))
f.inspect_types()
```

then when we dump type inference debugging inference, the final types are (relevant info only, other bits omitted):

```
---------------------------------Variable types---------------------------------
... <snip > ...
 'i': uint32,
 'i.1': uint32,
 'i.2': uint32,
 'n': int64}
```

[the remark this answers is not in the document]
jeromekelleher: Aha! Thank you so much @gmarkall, this is incredibly helpful. I am entirely satisfied :smile: 

Numba is, literally, awesome :star_struck:

jeromekelleher:
OK, @aktech, I think I'm happy with all the changes. Can you ping back when the comment stuff has been updated please so we can merge?

sgkit/distance/metrics.py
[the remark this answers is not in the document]
leofang: Apology @jeromekelleher, I was distracted and couldn't get back promptly, but looks like you are in good hands now 😅 Thanks @gmarkall for jumping in and providing such a detailed explanation! (I actually already linked to Graham's posts in my first reply in this PR.)

jeromekelleher:
I think this is basically ready to go @aktech?

aktech:
Give me a couple of hours, I can take a quick look by then.

tomwhite:
Is it worth squashing the commits on this one when merging?

aktech:
Yes, there are plenty of not so useful "fix" commits, unless someone has a
reason not to.

jeromekelleher:
Yep, I'd vote for squashing this down a bit.

aktech:
I think this is good to go now. Passing [GPU Tests (via Cirun)](https://github.com/aktech/sgkit/runs/2544046510?check_suite_focus=true) for reference.

Issue #499, opened 2021-03-18
Cohort observed Heterozygosity


timothymillar:
In diploids the observed heterozygosity at a locus is typically described as the proportion of individuals who are heterozygous. This is generalised to polyploids in Hardy 2016 ([paper](https://onlinelibrary.wiley.com/doi/abs/10.1111/1755-0998.12431)) who defines the individual heterozygosity at a locus as the probability of drawing a pair of IBS alleles without replacement. For diploids this results in a value of 1.0 (het) or 0.0 (hom) and for ployploids it a values in the interval [0, 1]. Observed heterozygosity is then calculated as the mean of individual heterozygosities.

Individual and observed heterozygosity could be implemented as a pair of functions returning "call_heterozygosity" and "cohort_heterozygosity" respectively. Note that the above paper contains some examples suitable for unit tests.

Some related discussion on the [pystatgen discourse](https://discourse.pystatgen.org/t/calculation-of-expected-and-observed-heterozygosity/89).

jeromekelleher:
Sounds good to me @timothymillar - are you volunteering to implement? :wink:

PR #505, opened 2021-03-23
Cohort Observed Heterozygosity

timothymillar:
Implements #499 

Some thoughts:
- This implementation aims to match that of the existing `diversity` method. One difference is that individual heterozygosity is undefined for haploid calls and hence any haploid calls won't contribute to the cohort observed heterozygosity.
- I have followed the use of `np.sum` in the windowed version of  `diversity` [(line)](https://github.com/pystatgen/sgkit/blob/master/sgkit/stats/popgen.py#L104). I couldn't find much discussion on this choice but I assume it's to avoid taking the mean prior to its use in other statistics?
- I'm not super happy with the length of `'stat_observed_heterozygosity'` as a variable name but `'stat_heterozygosity'` could easily be mistaken for the expected heterozygosity.
- I'm not sure if `aggregation.py` is the best place for `individual_heterozygosity` but it is an aggregation of sorts and it doesn't really fit in `popgen.py` as there is no use of cohorts.

I also noted that there is a bit of discrepancy in naming conventions with capitalized names used for Dask arrays in `aggregation.py ` and lower case used in `popgen.py`. I have followed the style used in each file for now.

### DIFF
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -712,3 +712,73 @@ def infer_sample_ploidy(
 
     new_ds = create_dataset({variables.sample_ploidy: sample_ploidy})
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+def individual_heterozygosity(
+    ds: Dataset,
+    *,
+    call_allele_count: Hashable = variables.call_allele_count,
+    merge: bool = True,
+) -> Dataset:
+    """Compute per sample observed heterozygosity.
+
+    Individual heterozygosity is defined as the probability that
+    two alleles drawn at random without replacement are not identical
+    in state. Therefore, individual heterozygosity is defined for
+    diploid and polyploid calls but will return nan in the case of
+    haploid calls.
+
+    Parameters
+    ----------
+    ds
+        Dataset containing genotype calls.
+    call_allele_count
+        Input variable name holding call_allele_count as defined by
+        :data:`sgkit.variables.call_allele_count_spec`.
+        If the variable is not present in ``ds``, it will be computed
+        using :func:`count_call_alleles`.
+    merge
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
+    Returns
+    -------
+    A dataset containing :data:`sgkit.variables.call_heterozygosity_spec`
+    of per genotype observed heterozygosity with shape (variants, samples)
+    containing values within the inteval [0, 1] or nan if ploidy < 2.
+
+    Examples
+    --------
+
+    >>> import sgkit as sg
+    >>> ds = sg.simulate_genotype_call_dataset(n_variant=4, n_sample=2, seed=1)
+    >>> sg.display_genotypes(ds) # doctest: +NORMALIZE_WHITESPACE
+    samples    S0   S1
+    variants
+    0         1/0  1/0
+    1         1/0  1/1
+    2         0/1  1/0
+    3         0/0  0/0
+
+    >>> sg.individual_heterozygosity(ds)["call_heterozygosity"].values # doctest: +NORMALIZE_WHITESPACE
+    array([[1., 1.],
+           [1., 0.],
+           [1., 1.],
+           [0., 0.]])
+    """
+    ds = define_variable_if_absent(
+        ds, variables.call_allele_count, call_allele_count, count_call_alleles
+    )
+    variables.validate(ds, {call_allele_count: variables.call_allele_count_spec})
+
+    AC = da.asarray(ds.call_allele_count)
+    K = AC.sum(axis=-1)
+    # use nan denominator to avoid divide by zero with K - 1
+    K2 = da.where(K > 1, K, np.nan)
+    AF = AC / K2[..., None]
+    HI = (1 - da.sum(da.power(AF, 2), axis=-1)) * (K / (K2 - 1))
+    new_ds = create_dataset(
+        {variables.call_heterozygosity: (("variants", "samples"), HI)}
+    )
+    return conditional_merge_datasets(ds, new_ds, merge)
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -720,13 +720,13 @@ def individual_heterozygosity(
     call_allele_count: Hashable = variables.call_allele_count,
     merge: bool = True,
 ) -> Dataset:
-    """Compute per sample observed heterozygosity.
+    """Compute per call individual heterozygosity.
 
-    Individual heterozygosity is defined as the probability that
-    two alleles drawn at random without replacement are not identical
-    in state. Therefore, individual heterozygosity is defined for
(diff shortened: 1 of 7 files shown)


timothymillar:
I'm not sure what exactly is causing the crashes with the initial push but it seems to relate to the uneven chunk sizes in the sample dimension for `test_observed_heterozygosity` and `test_observed_heterozygosity__windowed`. The test itself passes but shortly after python crashes with `Fatal Python error: Aborted` or `Windows fatal exception: code 0xc0000374` on linux and windows respectively. The timing of the crash is variable but occurs after the test has passed. My guess is that this is a Dask issue however I can't debug right now.

For now I have removed the test case with uneven chunk size.

sgkit/stats/aggregation.py
```diff
@@ -712,3 +712,73 @@ def infer_sample_ploidy(
...
+def individual_heterozygosity(
+    ds: Dataset,
+    *,
+    call_allele_count: Hashable = variables.call_allele_count,
+    merge: bool = True,
+) -> Dataset:
+    """Compute per sample observed heterozygosity.
+
+    Individual heterozygosity is defined as the probability that
+    two alleles drawn at random without replacement are not identical
```
jeromekelleher, line 726: Maybe insert "from an individual at a given site" after "at random"?

sgkit/stats/aggregation.py
```diff
@@ -712,3 +712,73 @@ def infer_sample_ploidy(
...
+    merge
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
+    Returns
+    -------
+    A dataset containing :data:`sgkit.variables.call_heterozygosity_spec`
+    of per genotype observed heterozygosity with shape (variants, samples)
+    containing values within the inteval [0, 1] or nan if ploidy < 2.
```
jeromekelleher, line 749: inteval -> "interval"

sgkit/stats/aggregation.py
```diff
@@ -712,3 +712,73 @@ def infer_sample_ploidy(
...
+        ds, variables.call_allele_count, call_allele_count, count_call_alleles
+    )
+    variables.validate(ds, {call_allele_count: variables.call_allele_count_spec})
+
+    AC = da.asarray(ds.call_allele_count)
+    K = AC.sum(axis=-1)
+    # use nan denominator to avoid divide by zero with K - 1
+    K2 = da.where(K > 1, K, np.nan)
+    AF = AC / K2[..., None]
+    HI = (1 - da.sum(da.power(AF, 2), axis=-1)) * (K / (K2 - 1))
```
jeromekelleher, line 780: Is it worth calling ``da.power(AF, 2)``? I'm not sure how sophisticated the method is, but it could be more efficient to just call ``AF ** 2`` (or ``AF * AF``).

Ah, or am I being stupid, and you need to do this for Dask to be used properly? (Sorry, my Dask knowledge is rudimentary)

sgkit/stats/popgen.py
```diff
@@ -910,3 +914,156 @@ def Garud_H(
...
+            out[j] = np.nan
+
+
+def observed_heterozygosity(
+    ds: Dataset,
+    *,
+    call_heterozygosity: Hashable = variables.call_heterozygosity,
+    merge: bool = True,
+) -> Dataset:
+    """Compute per cohort observed heterozygosity.
```
jeromekelleher, line 970: Could either link to the documentation for the other heterozygosity function, or give the definition here again.

sgkit/tests/test_popgen.py
```diff
@@ -456,3 +458,132 @@ def test_Garud_h__raise_on_no_windows():
...
+    ho = observed_heterozygosity(ds)["stat_observed_heterozygosity"]
+    np.testing.assert_almost_equal(
+        ho,
+        np.array(
+            [
+                [1 / 4, 2 / 3, 2 / 3],
+                [1 / 2, 5 / 3, np.nan],
+            ]
+        ),
+    )
```
jeromekelleher, line 589: It would be nice to add a test to check that we're computing the same value as scikit allele for a few simulated examples. Can we compare to https://scikit-allel.readthedocs.io/en/v0.9.1/stats.html#allel.stats.heterozygosity_observed ?

jeromekelleher (commented):
Looks great, thanks @timothymillar! Some minor comments.

jeromekelleher:
@tomwhite any thoughts on what we should do about the crash @timothymillar has hit above? Is it worth trying to isolate a case where it happens?

sgkit/stats/aggregation.py
timothymillar: No I think you are correct, `AF ** 2` is fine. I'm not that familiar with the internals of Dask either.

sgkit/stats/aggregation.py
```diff
@@ -712,3 +712,73 @@ def infer_sample_ploidy(
...
     return conditional_merge_datasets(ds, variables.validate(new_ds), merge)
+
+
+def individual_heterozygosity(
+    ds: Dataset,
+    *,
+    call_allele_count: Hashable = variables.call_allele_count,
+    merge: bool = True,
+) -> Dataset:
+    """Compute per sample observed heterozygosity.
```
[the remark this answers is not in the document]
timothymillar: That would be clearer! though they are essentially the same thing when referring to an individual.

timothymillar:
Thanks for reviewing @jeromekelleher!
Just a note that the comparison to scikit-allel is only implemented for diploid genotypes because scikit-allel treats any level of polyploid heterozygosity as being fully heterozygous e.g. hi(AAAB) == hi(AABB) == hi(ABCD) == 1. This is ultimately due to the `is_het` method which returns boolean values.

sgkit/stats/popgen.py
```diff
@@ -910,3 +914,162 @@ def Garud_H(
...
+    n_cohorts = len(out)
+    for i in range(n_samples):
+        h = hi[i]
+        if not np.isnan(h):
+            c = cohorts[i]
+            out[c] += h
+            _[c] += 1
+    for j in range(n_cohorts):
+        n = _[j]
+        if n:
```
jeromekelleher, line 958: I'm a bit of an anti-Falsey fundamentalist, but I think it'd be clearer to say ``if n != 0`` here wouldn't it?

sgkit/tests/test_popgen.py
```diff
@@ -456,3 +458,165 @@ def test_Garud_h__raise_on_no_windows():
...
+
+
```
jeromekelleher, line 591: could parmetrize across a few seeds as well? It's a tricky balance between being thorough and excessive.

jeromekelleher (approved):
Looks great, thanks @timothymillar! A couple of more random nitpicky comments, feel free to ignore.

sgkit/stats/popgen.py
timothymillar: Agreed, `if n != 0` is much easier to read.

tomwhite:
One minor comment - it would be worth adding the new functions to `docs/api.rst`.

jeromekelleher:
Happy to merge when you are @tomwhite

PR #508, opened 2021-03-23
Field parsing should use supplied max_alt_alleles #507

### LINKED ISSUE #507: max_alt_alleles is ignored when parsing VCF fields
The fixed value of `DEFAULT_MAX_ALT_ALLELES` is always used, which can cause an error due to a dimension size mismatch.

tomwhite:
Fixes #507


jeromekelleher (approved):
LGTM

PR #510, opened 2021-03-24
Dask crash related to uneven chunk size in test_observed_heterozygosity

timothymillar:
Don't merge. 
This is a demonstration of the bug mentioned in #505 [(comment)](https://github.com/pystatgen/sgkit/pull/505#issuecomment-805604674).

cc @tomwhite 

This crash occurs (only?) when running the full unit test suit, usually after `test_observed_heterozygosity` and `test_observed_heterozygosity__windowed` have both passed. On my machine it most commonly occurs around `sgkit/tests/test_preprocessing.py::test_filter_partial_calls__mixed_ploidy` and `sgkit/tests/test_regenie.py::test_regenie__glow_comparison` though I think this is due to timing rather than an interaction with those tests.

This crash is highly reproducible on my machine (GNU/Linux 4.4.0-19041-Microsoft x86_64 in WSL1 on Ryzen 4700u).
The crash seems to specifically relate to the uneven chunk size on the sample dimension in the third iteration of both tests `((3, 1), (4, 2), (1, 3))`.

Example output from my machine: [(gist)](https://gist.github.com/timothymillar/fbc4a58562a9bf080a0a14de8e20b263)

I have tried chunk variations including `((3, 1), (3, 3), (1, 3))` and `((3, 1), (2, 2, 2), (1, 3))` which both seem reliably avoid the crash over several repetitions. I also *think* I had a single crash with  `((3, 1), (6,), (1, 3))` though if that wasn't my mistake then it is a less reproducible example.

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -15,6 +15,7 @@ from .stats.aggregation import (
     count_call_alleles,
     count_cohort_alleles,
     count_variant_alleles,
+    individual_heterozygosity,
     sample_stats,
     variant_stats,
 )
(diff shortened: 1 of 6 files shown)


jeromekelleher:
Thanks @timothymillar, this is very helpful.

tomwhite:
I wonder if this could be related to numba. Does it still happen if run as `NUMBA_DISABLE_JIT=1 pytest`?

timothymillar:
I get the same [result](https://gist.github.com/timothymillar/fbc4a58562a9bf080a0a14de8e20b263#file-example3-txt) with `NUMBA_DISABLE_JIT=1 pytest` (repeatably).

tomwhite:
Thanks for trying that @timothymillar.

cc-ing @jrbourbeau and @mrocklin who may have seen something like this before, or how to diagnose further.

PR #513, opened 2021-03-31
In gwas_linear_regression allow empty covariates #509

### LINKED ISSUE #509: Allow empty covariate list in gwas_linear_regression
In #463, the linear regression is first run with no covariates (i.e. only with an intercept). This is not possible with the current implementation of `gwas_linear_regression`, so we should add that capability. (There should be a check to see that `add_intercept` is True in the case of no covariates.)

tomwhite:
Fixes #509

### DIFF
--- a/sgkit/stats/association.py
+++ b/sgkit/stats/association.py
@@ -209,9 +209,15 @@ def gwas_linear_regression(
 
     G = _get_loop_covariates(ds, dosage=dosage, call_genotype=call_genotype)
 
-    X = da.asarray(concat_2d(ds[list(covariates)], dims=("samples", "covariates")))
-    if add_intercept:
-        X = da.concatenate([da.ones((X.shape[0], 1), dtype=X.dtype), X], axis=1)
+    if len(covariates) == 0:
+        if add_intercept:
+            X = da.ones((ds.dims["samples"], 1), dtype=float)
+        else:
+            raise ValueError("add_intercept must be True if no covariates specified")
+    else:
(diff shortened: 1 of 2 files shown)


jeromekelleher:
One a side note here @tomwhite, I was looking to see if we had full test coverage here and I'm not seeing a codecov report. Is it just this PR or is this broken generally?

tomwhite:
@jeromekelleher coverage reports are being run and uploaded (https://app.codecov.io/gh/pystatgen/sgkit/pulls?page=1&state=open&order=-pullid) and coverage is 100%. However, there seem to be some problems reported as "Missing base report" and "Unable to find commit in Github", which is perhaps why we don't get reports posted on all PRs here. Opened #514

sgkit/stats/association.py
```diff
@@ -209,9 +209,15 @@ def gwas_linear_regression(
 
     G = _get_loop_covariates(ds, dosage=dosage, call_genotype=call_genotype)
 
-    X = da.asarray(concat_2d(ds[list(covariates)], dims=("samples", "covariates")))
-    if add_intercept:
-        X = da.concatenate([da.ones((X.shape[0], 1), dtype=X.dtype), X], axis=1)
+    if len(covariates) == 0:
+        if add_intercept:
+            X = da.ones((ds.dims["samples"], 1), dtype=float)
```
eric-czech, line 214: Any objection to float32 here to avoid promotion if `G` is 32-bit as well?

tomwhite: Good call

PR #521, opened 2021-04-15
Updates for msprime 1.0

jeromekelleher:
Minimal updates to use the msprime 1.0 release, and update our testing code the use the new APIs (which generate more suitable data for us).

I hit two separate issues while doing this. I suggest we file issues for these, and merge this much @tomwhite?

There's more we can do in terms of generating cohorts more realistically, but these should probably be done in follow-ups.

### DIFF
--- a/sgkit/tests/test_popgen.py
+++ b/sgkit/tests/test_popgen.py
@@ -5,6 +5,7 @@ import dask.array as da
 import msprime  # type: ignore
 import numpy as np
 import pytest
+import tskit  # type: ignore
 import xarray as xr
 from allel import hudson_fst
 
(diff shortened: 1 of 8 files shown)


tomwhite (commented):
These changes look good @jeromekelleher. I think the new `simulate_ts` function just needs type annotations then it should be ready to go in.

jeromekelleher:
Any idea what's happening [here](https://github.com/pystatgen/sgkit/runs/2352176242?check_suite_focus=true#step:6:24) @tomwhite? Seems like some issue with autodoc typehints? I can't reproduce locally.

tomwhite:
I see this locally if I use sphinx-autodoc-typehints 1.12.0 (which was released earlier today), but not with 1.11.1. Perhaps we can pin that version in requirements-doc.txt.

sgkit/tests/test_popgen.py
```diff
@@ -309,9 +345,10 @@ def test_Fst__windowed(sample_size, n_cohorts, chunks):
         )  # scikit-allel has final window missing
 
 
+@pytest.mark.skip("Bug in Tajimas D for n=100?")
```
tomwhite, line 348: Can we run the cases that work in the meantime?

jeromekelleher: Yes - I was being lazy. How do you skip a particular example with parametrize?

tomwhite: I don't know if it's possible. But we could remove `100` from the list, as long as we have an issue open (which we do: #523).

jeromekelleher: Ah, actually no, it's failing for all n > 2, so it's probably simplest to keep skipping it rather than giving a false sense of security here.

jeromekelleher:
Hmm, still getting the failure after pinning to sphinx-autodoc-typehints-1.11.1 @tomwhite

tomwhite:
Looks like it's this problem: https://github.com/theislab/scanpydoc/pull/32, now fixed but not released. How about using `git+https://github.com/theislab/scanpydoc.git#egg=scanpydoc` as the dependency in requirements-doc.txt? (That will need  sphinx-autodoc-typehints 1.12.0.)

sgkit/tests/test_popgen.py
tomwhite: OK, I agree it's best to skip then.

tomwhite:
This is failing because the coverage is no longer 100% since some tests are being excluded. I had a look to see if I could see how to fix the underlying problems.
* The Fst test (`test_Fst__windowed`) passes - I couldn't reproduce the problem here. Perhaps it's OK now?
* The Tajimas D test fails. I found it's because the number of segregating sites is different: for the 10 sample test, in tskit the number of segregating sites is 31, and in sgkit it is 28. The odd thing is that there are 29 variants (sites) - so how can there be 31 segregating sites? (I'm probably missing something though!)

tomwhite:
@jeromekelleher any thoughts on my segregating sites question above?

jeromekelleher:
Following up now @tomwhite, thanks for the nudge

tomwhite:
I'm using OSX - so perhaps that's causing the difference.

jeromekelleher:
I've had a  look at the Tajima's D tests @tomwhite, and I'm not sure what's going on. I've enabled the test for n=2 so we get some coverage, so maybe we could merge this much (assuming that gives coverage)?

We're clearly not covering all the corner cases with these popgen stats currently, so I think we'll just need to follow up and beef the testing up a bit.

jeromekelleher:
> The Tajimas D test fails. I found it's because the number of segregating sites is different: for the 10 sample test, in tskit the number of segregating sites is 31, and in sgkit it is 28. The odd thing is that there are 29 variants (sites) - so how can there be 31 segregating sites? (I'm probably missing something though!)

That's odd - can you paste in more details? (Maybe we should move this to the issue #523?)

tomwhite (approved):

tomwhite:
> That's odd - can you paste in more details? (Maybe we should move this to the issue #523?)

I'll follow up in #523

jeromekelleher:
OK, happy to merge when you are @tomwhite. I'll follow up with the Fst one now and  make a PR - that one looks easy enough to resolve.

Issue #31, opened 2020-07-09
LD-prune implementation


eric-czech:
The prototype LD-prune implementation at [gwas-analysis/lib/method/core.py](https://github.com/related-sciences/gwas-analysis/blob/master/notebooks/platform/xarray/lib/method/core.py#L13) extends [skallel.locate_unlinked](https://scikit-allel.readthedocs.io/en/stable/stats/ld.html?highlight=locate_unlinked#allel.locate_unlinked) to be parallelizable by dask, runnable on GPU or CPU, and also adds support for windowing based on genomic distance (it is equivalent for fixed-size windows/steps).

This could be moved over as-is or it may be worth discussing some of the aspects of the implementation first.  

Either way, we should wait on resolutions to https://github.com/pystatgen/sgkit/issues/3 and https://github.com/pystatgen/sgkit/issues/29 so we know what target API to refactor towards.

jeromekelleher:
Can potentially use tskit's [ld calculator](https://tskit.readthedocs.io/en/latest/tutorial.html#calculating-ld) for validation of LD computations. The API is quite old and a bit limited, though. We're working on extending the functionality, but it'll be a while before it's ready.

eric-czech:
I was planning to start adding in pieces of this soon, and I wanted to better explain my plan for that first.

The prototype versions of this I had written reduce LD pruning into three separate problems, each of which I know are usable in at least one other context as well:

1. Determining what variants are included in windows for comparison
    - This is very similar to [hal.locus_windows](https://hail.is/docs/0.2/linalg/utils/index.html#hail.linalg.utils.locus_windows) except that I'm also computing similar windows that correspond to chunks rather than individual variants
    - This requires that all of the window definitions for a single contig fit in memory

2. Computing a sparse LD matrix
    - This becomes straightforward once the windows are defined and I had lifted [scikit_allele.rogers_huff_r](https://scikit-allel.readthedocs.io/en/stable/stats/ld.html#allel.rogers_huff_r) verbatim to do it
    - @jeromekelleher/@alimanfoo  do you know if there are any important differences between the tskit r2 calculation and the one in scikit-allel?
    - This requires that chunking for genotypes is only along one axis
      - The solution discussed [here on the Dask discourse](https://dask.discourse.group/t/creating-sparse-output-from-dense-inputs-to-matrix-multiplication/26) would be much better (and much more difficult)
3. Finding a maximal independent set (MIS) between variants with LD over a threshold
[...]

PR #532, opened 2021-04-20
LD prune

tomwhite:
This is a first draft of #31, which is a port of (some of) the LD-prune code written by @eric-czech in https://github.com/related-sciences/gwas-analysis/tree/master/notebooks/platform/xarray/lib. It's not complete, but I thought I'd post my progress so far.

* The implementation uses the existing `windows` function in sgkit, so it only supports fixed size "units"-based indexing (number of variants, not base pair positions). However, when more flexible indexing is added (#341), then LD prune will be able to use that (hopefully with no changes).
* The core of the LD matrix calculation (`_ld_matrix_jit`) iterates over all the windows in a chunk, and within each window iterates over all the variant pairs in the window. The original implementation arranged things slightly differently, with each window broken up into multiple "axis intervals", which effectively unrolled the loop of the pairwise comparisons. The approach here makes it possible to decouple the array chunking from the window definition, so you can use the existing Dask array chunks (the original implementation calculates its own chunks).
* The relevant unit tests have been ported, including a concordance check with scikit-allel's `locate_unlinked` function. This uses Hypothesis to generate lots of combinations of input array, windows, and chunk sizes. To get an exact match, I had to declare the numba types for the Rogers Huff R function to be single-precision floats. (There are slight differences if they are left as the default double precision.)
* The biggest outstanding issue is that the `ld_matrix` function returns a Dask dataframe, rather than an Xarray Dataset - which would be more consistent with the rest of sgkit. It would also be better to provide `scores` as a Dataset variable, rather than an array, since its dimension could then be validated.
* Documentation is yet to be added. As a part of this PR I will add function docs (and expose `ld_matrix`, `maximal_independent_set`, and `ld_prune` in the public API).


jeromekelleher:
Looks great @tomwhite. One quick thought: I wonder if we could pull out a PR to just compute LD (r2) first? We'll definitely want this and it would reduce the size of this PR a bit.

tomwhite:
@jeromekelleher good suggestion. I'll need to write some unit tests for that function, which will be a good thing.

tomwhite:
I've rebased this to include #536, and I've added a way to optionally specify scores as a dataset variable.

I had a quick look at having `ld_matrix` return an Xarray Dataset rather than a Dask DataFrame, but I think that the computation would be eager, since the number of rows in the result is unknown (there is a bit of discussion about it here: https://github.com/pydata/xarray/issues/3929).

It also wasn't particularly obvious what the variable names would be: `ld_variant_0` and `ld_variant_1` for `i` and `j`, perhaps, but `maximal_independent_set` is not just for LD calculations, so these might not be quite right for that.

So I'm not sure what to do about this - we could perhaps not expose `ld_matrix` or `maximal_independent_set` as public methods - or we could have them as public utility methods that use Dask DataFrames.

tomwhite:
* As discussed in the [meeting](https://github.com/pystatgen/sgkit/discussions/553), I've left the return value of `ld_matrix` and the input to `maximal_independent_set` as a Dask dataframe.
* I've also left the variables in the dataframe as `i`, `j` etc, since they don't form part of a larger dataset, so they an be considered "local" variables. Happy to change them if there were some better names for them.
* I've changed `ld_prune` to return a dataset where the variants in linkage disequilibrium have been removed. This is similar to what Hail does. For more control over the process, it's straightforward to use `ld_matrix` followed by `maximal_independent_set` to find the variants to drop.
* I've added documentation and an example.

This still depends on #536.

PR #536, opened 2021-04-21
Rogers Huff R

tomwhite:
Here's the part of #31 and #532 to compute r2, as suggested by @jeromekelleher.

### DIFF
--- a/sgkit/stats/ld.py
+++ b/sgkit/stats/ld.py
@@ -7,6 +7,7 @@ from numba import njit
 from sgkit.typing import ArrayLike
 
 
+# Note that float32 types are used in this function to get an exact match with scikit-allel's equivalent function.
 @njit(nogil=True, fastmath=True, cache=True, locals={"m0": numba.float32, "m1": numba.float32, "v0": numba.float32, "v1": numba.float32, "cov": numba.float32, "n": numba.int32})  # type: ignore
 def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
     """Rogers Huff R
--- a/sgkit/stats/ld.py
+++ b/sgkit/stats/ld.py
@@ -1,0 +1,59 @@
+import math
+
+import numba
+import numpy as np
+from numba import njit
+
+from sgkit.typing import ArrayLike
+
+
+@njit(nogil=True, fastmath=True, cache=True, locals={"m0": numba.float32, "m1": numba.float32, "v0": numba.float32, "v1": numba.float32, "cov": numba.float32, "n": numba.int32})  # type: ignore
+def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
+    """Rogers Huff R
+
+    From https://github.com/cggh/scikit-allel/blob/961254bd583572eed7f9bd01060e53a8648e620c/allel/opt/stats.pyx
+    """
+    # initialise variables
+    m0 = m1 = v0 = v1 = cov = 0.0
+    n = 0
+
+    # iterate over input vectors
+    for i in range(len(gn0)):
+        x = gn0[i]
+        y = gn1[i]
+        # consider negative values as missing
+        if x >= 0 and y >= 0:
+            n += 1
+            m0 += x
+            m1 += y
+            v0 += x ** 2
+            v1 += y ** 2
+            cov += x * y
+
+    # early out
+    if n == 0:
+        return np.nan
+
+    # compute mean, variance, covariance
+    m0 /= n
+    m1 /= n
+    v0 /= n
+    v1 /= n
+    cov /= n
+    cov -= m0 * m1
+    v0 -= m0 * m0
+    v1 -= m1 * m1
+
+    d = math.sqrt(v0 * v1)
+    if d < np.finfo(np.float32).tiny:
+        return np.nan
+
+    # compute correlation coeficient
+    r = cov / d
+
+    return r
+
+
+@njit(nogil=True, fastmath=True, cache=True)  # type: ignore
+def rogers_huff_r2_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
+    return rogers_huff_r_between(gn0, gn1) ** 2  # type: ignore
--- a/sgkit/tests/test_ld.py
+++ b/sgkit/tests/test_ld.py
@@ -1,20 +1,28 @@
+import allel
 import numpy as np
 
 from sgkit.stats.ld import rogers_huff_r2_between, rogers_huff_r_between
 
 
 def test_rogers_huff_r_between():
-    gna = np.array([0, 1, 2])
-    gnb = np.array([0, 1, 2])
-    assert rogers_huff_r_between(gna, gnb) == 1.0
-    assert rogers_huff_r2_between(gna, gnb) == 1.0
+    gna = np.array([[0, 1, 2]])
+    gnb = np.array([[0, 1, 2]])
+    assert rogers_huff_r_between(gna[0], gnb[0]) == 1.0
+    assert rogers_huff_r2_between(gna[0], gnb[0]) == 1.0
+    assert allel.rogers_huff_r_between(gna, gnb) == rogers_huff_r_between(
+        gna[0], gnb[0]
+    )
 
-    gna = np.array([0, 1, 2])
-    gnb = np.array([2, 1, 0])
-    assert rogers_huff_r_between(gna, gnb) == -1.0
-    assert rogers_huff_r2_between(gna, gnb) == 1.0
+    gna = np.array([[0, 1, 2]])
+    gnb = np.array([[2, 1, 0]])
+    assert rogers_huff_r_between(gna[0], gnb[0]) == -1.0
+    assert rogers_huff_r2_between(gna[0], gnb[0]) == 1.0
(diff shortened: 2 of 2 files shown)


sgkit/stats/ld.py
```diff
@@ -0,0 +1,59 @@
...
+        return np.nan
+
+    # compute correlation coeficient
+    r = cov / d
+
+    return r
+
+
+@njit(nogil=True, fastmath=True, cache=True)  # type: ignore
+def rogers_huff_r2_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
```
jeromekelleher, line 58: We need to check the array bounds here, don't we? What happens if we get different length arrays?

sgkit/stats/ld.py
```diff
@@ -0,0 +1,59 @@
...
+    if d < np.finfo(np.float32).tiny:
+        return np.nan
+
+    # compute correlation coeficient
+    r = cov / d
+
+    return r
+
+
+@njit(nogil=True, fastmath=True, cache=True)  # type: ignore
```
jeromekelleher, line 57: There's no point in jitting this, is there?

sgkit/stats/ld.py
```diff
@@ -0,0 +1,59 @@
+import math
+
+import numba
+import numpy as np
+from numba import njit
+
+from sgkit.typing import ArrayLike
+
+
+@njit(nogil=True, fastmath=True, cache=True, locals={"m0": numba.float32, "m1": numba.float32, "v0": numba.float32, "v1": numba.float32, "cov": numba.float32, "n": numba.int32})  # type: ignore
```
jeromekelleher, line 10: Does it make much difference to declare these as float32s? I'd have thought it would be safer to do the calc in double precision, and I'd be surprised if it made any perf difference. Particularly for accumulating values in a CPU register, I can't see any point in doing this in single precision.

sgkit/tests/test_ld.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+
+from sgkit.stats.ld import rogers_huff_r2_between, rogers_huff_r_between
+
+
```
jeromekelleher, line 5: Good to compare with scikit allele on a range of simulated datasets.

jeromekelleher (commented):
The calculation itself looks great - I'm sure it'll be super fast. 

Is this intended to ultimately be a public interface? If so, we'll need to think about how to structure things a bit. Ideally we'll want a method to compute the ``r2`` and ``r`` measures of LD (which will probably need to be computed independently for some estimators; I assume it's OK to just sqrt r2 to get r here). So, we'd want to have some function akin to ``diversity`` which takes an "estimator" as an argument. I don't know what a good name for this function would be - maybe ``linkage_disequilibrium`` is the least bad option.

So, like ``linkage_disequilibrium(measure="r2", estimator="rogers_huff")`` or maybe ``ld_r2(estimator="rogers_huff")`` would be better. I'd have to dig out some text books to give a better idea (I've cowritten a book chapter on LD, but can mainly remember that it's very complicated!)

That's how I'd proceed here: make the public interface for computing LD, and then we can call into that ourselves for the LD pruning stuff.

tomwhite:
Thanks for the review @jeromekelleher! This function was not intended to be a public interface - all of the public methods in sgkit so far operate on Datasets, and it's not clear (to me) how r2 would be exposed in a Dataset form. It's used in the `ld_matrix` function (which would be public) in #532 over pairs of variants, so not very easy for users to apply in another way.

sgkit/stats/ld.py
tomwhite: Probably only needed if it's a public method (see other comment).

sgkit/stats/ld.py
tomwhite: I didn't measure the effect of jitting here, since it came from @eric-czech's original implementation.

sgkit/tests/test_ld.py
tomwhite: +1

sgkit/stats/ld.py
tomwhite: Ah I forgot to mention here that I needed to declare the local variables as float32 to get concordance with scikit-allel's `locate_unlinked` function. I agree it shouldn't make much of a performance difference though. See #532.

jeromekelleher:
OK, thanks @tomwhite. So, if this is an internal method I think we just need 

- a bit more developer aimed comments explaining stuff like the choice of float32 (which I'm suspicious of - maybe we just trying to be wrong in the same way as scikit-allele here? If there's any difference beyond ~6 digits between the float and double versions, then I can't see any way that the float version can be "more right".) 
- a bit more test coverage, especially poking at the awkward corner cases (coalescent sims are pretty good at exposing these)

How does that sound?

sgkit/stats/ld.py
jeromekelleher: Unless there's a loop in the code, or we're calling it from numba-jitted code, then marking it as njit can only slow down load times, I think. Numpy will implement the squaring just as well, won't it?

Or is there an advantage to running ``a ** 2`` as numba vs ``a * a`` in numpy (or ``a ** 2`` in numpy) @eric-czech?

tomwhite:
Thanks @jeromekelleher. I have added some more documentation about float32, and changed the tests to also check against scikit-allel.

For a more comprehensive test, I agree using simulated data from msprime would be ideal here, using something similar to what we do in `test_popgen.py`. We could even integrate it with hypothesis to generate more extensive sets of test data. I think it might be worth factoring out a test utility for that, perhaps in a separate issue?

sgkit/stats/ld.py
eric-czech: If I recall we don't have a choice b/c you can't call a function from inside another jitted function unless it is also jitted.  The function calling `rogers_huff_r2_between` just isn't in this PR yet (the jitted LD methods will call it).

eric-czech (approved):

sgkit/stats/ld.py
```diff
@@ -0,0 +1,60 @@
...
+    m1 /= n
+    v0 /= n
+    v1 /= n
+    cov /= n
+    cov -= m0 * m1
+    v0 -= m0 * m0
+    v1 -= m1 * m1
+
+    d = math.sqrt(v0 * v1)
+    if d < np.finfo(np.float32).tiny:
```
eric-czech, line 49: I believe my reason for adding this was that without it, you can end up with results outside of [0, 1].  This may not fix that in all cases though.  We may want to consider instead returning `nan` if `d` is too small (like this) but also clip `r` to [0, 1].  It seems odd to let `r` be either in [0, 1] or `inf` but I'm good with that too.  Perhaps there's a better way?

sgkit/stats/ld.py
jeromekelleher: Ah, makes sense.

sgkit/stats/ld.py
jeromekelleher: A lot of the time it's simpler and more consistent to just let the floating point error happen and IEEE define what the exact semantics should be. What happens if we just don't make the check here? I think it's fine that r is in [0, 1] except for cases where we hit against numerical limits (and use the IEEE semantics for those limits)

eric-czech: That works for me.  I'd like to suggest that we then add a note in the docs that inf is a possibility to avoid users tripping up on something like `xr.DataArray([np.inf]).isnull().item(0) == False # or np.isnan(np.inf)`, but I'm sure we've not documented it in other places.  I'd still be weakly in favor of documenting it anyways since I know I would appreciate it as a user.

jeromekelleher: Sounds good @eric-czech, good for us to document and test these corner cases. In general, I guess we should encourage people to use ``np.isfinite`` rather than ``not np.isnan`` to make client code robust, though.

eric-czech: 👍 

I wish Xarray had made that as prominent as `DataArray.isnull` rather than having to do `DataArray.pipe(xr.ufuncs.isfinite)`.

hammer: @eric-czech file an issue upstream?

eric-czech: https://github.com/pydata/xarray/issues/5261

We should talk to Quansight about getting a bot to file Xarray issues like this automatically 😉

tomwhite: +1 to removing the if clause. If we also replace the division as follows

```
-    r = cov / d
+    r = np.divide(cov, d)
```

Then it would return `np.inf` in some cases. However, I have not been able to find what these cases are, so I'm not sure how to document them. (An input containing `np.inf` will return `np.nan` for example, since when computing the covariance `np.inf - np.inf` is calculated, which is `np.nan`.)

So we could just say that users should be prepared to handle `np.nan` and infinities, or we could ensure that the return value is always a finite float or `np.nan` by adding:

```
+    if np.isinf(r):
+        return np.nan
```

Thoughts?

jeromekelleher: I think we can probably just say that the method will return non-finite values (IEEE inf and NaNs) under certain situations (e.g AN EXAMPLE), and the user should be prepared to deal with that by checking np.isfinite. Checking isfinite isn't any worse than check isnan.

The trouble with putting in special cases for numerical corner cases is that you nearly always miss some. Simpler in the long run to just let the maths and IEEE semantics do the work.

I don't feel that strongly about it though, feel free to ignore this.

Issue #523, opened 2021-04-15
Problem with Tajimas D tests


jeromekelleher:
Since #521 the Tajima's D tests have been skipped. We were getting different values > n=2.

tomwhite:
Continuing the discussion from https://github.com/pystatgen/sgkit/pull/521#issuecomment-823405576. For n=10, this change

```
diff --git a/sgkit/tests/test_popgen.py b/sgkit/tests/test_popgen.py
index a5a9f9a..de27946 100644
--- a/sgkit/tests/test_popgen.py
+++ b/sgkit/tests/test_popgen.py
@@ -349,9 +349,12 @@ def test_Fst__windowed(sample_size, n_cohorts, chunks):
 
 # Skipping values > 2, because of this issue:
 # https://github.com/pystatgen/sgkit/issues/522
-@pytest.mark.parametrize("sample_size", [2])
+@pytest.mark.parametrize("sample_size", [10])
 def test_Tajimas_D(sample_size):
     ts = simulate_ts(sample_size)
+    print()
+    print(ts)
+    print("segregating_sites", ts.segregating_sites(span_normalise=False))
     ds = ts_to_dataset(ts)  # type: ignore[no-untyped-call]
     ds, subsets = add_cohorts(ds, ts, cohort_key_names=None)  # type: ignore[no-untyped-call]
     n_variants = ds.dims["variants"]
```
prints
```
╔═══════════════════════╗
║TreeSequence           ║
╠═══════════════╤═══════╣
║Trees          │      7║
╟───────────────┼───────╢
║Sequence Length│  100.0║
╟───────────────┼───────╢
║Sample Nodes   │     10║
╟───────────────┼───────╢
║Total Size     │5.5 KiB║
╚═══════════════╧═══════╝
╔═══════════╤════╤══════════╤════════════╗
║Table      │Rows│Size      │Has Metadata║
╠═══════════╪════╪══════════╪════════════╣
║Edges      │  36│1012 Bytes│          No║
╟───────────┼────┼──────────┼────────────╢
║Individuals│  10│ 172 Bytes│          No║
╟───────────┼────┼──────────┼────────────╢
║Migrations │   0│   4 Bytes│          No║
╟───────────┼────┼──────────┼────────────╢
║Mutations  │  36│   1.0 KiB│          No║
╟───────────┼────┼──────────┼────────────╢
║Nodes      │  23│ 556 Bytes│          No║
╟───────────┼────┼──────────┼────────────╢
[...]

PR #542, opened 2021-04-23
Fix issues in Tajima's D calculation

jeromekelleher:
This is wip for fixing #523, and a few other issues that have cropped up while I was in there.

The original issue was easy enough to fix, but two other things have popped up:

1. It would be easiest if we produced exactly the same answer as tskit in division by zero cases. Tskit has taken the approach of letting the errors happen and using numpy to suppress the warnings, so we can get -inf, nan or inf, depending on the input. However, Dask doesn't support this, it seems (https://github.com/dask/dask/issues/3245), which is awkward. Not sure how to proceed here.
2. It's not clear to me that we're handling windows properly in Tajima's D.

@tomwhite, good to get your thoughts on this. Am I doing the windows-of-size-one thing correctly?

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -554,8 +554,10 @@ def Tajimas_D(
 
     ac = ds[variant_allele_count]
 
-    # count segregating
-    S = ((ac > 0).sum(axis=1) > 1).sum()
+    # count segregating. Note that this uses the definition in tskit,
+    # which is the number of alleles - 1. In the biallelic case this
+    # gives us the number of non-monomorphic sites.
+    S = ((ac > 0).sum(axis=1) - 1).sum()
 
     # assume number of chromosomes sampled is constant for all variants
     # NOTE: even tho ac has dtype uint, we promote the sum to float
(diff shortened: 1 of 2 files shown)


sgkit/stats/popgen.py
```diff
@@ -554,8 +554,10 @@ def Tajimas_D(
 
     ac = ds[variant_allele_count]
 
-    # count segregating
-    S = ((ac > 0).sum(axis=1) > 1).sum()
+    # count segregating. Note that this uses the definition in tskit,
+    # which is the number of alleles - 1. In the biallelic case this
+    # gives us the number of non-monomorphic sites.
+    S = ((ac > 0).sum(axis=1) - 1).sum()
```
tomwhite, line 560: It looks like windowing has not been implemented for Tajima's D yet. The segregating sites variable (`S`) needs to be calculated per-window (using `window_statistic`) rather than having a total sum calculated (which assumes a single window).

sgkit/tests/test_popgen.py
```diff
@@ -359,6 +357,22 @@ def test_Tajimas_D(sample_size):
...
+    print("d = ", ts_d)
+    np.testing.assert_allclose(d, ts_d)
+
+
+@pytest.mark.parametrize("sample_size", [2, 3, 5, 10, 100])
+def test_Tajimas_D_per_site(sample_size):
+    ts = simulate_ts(sample_size, random_seed=1234)
+    ds = ts_to_dataset(ts)  # type: ignore[no-untyped-call]
+    ds, subsets = add_cohorts(ds, ts, cohort_key_names=None)  # type: ignore[no-untyped-call]
+    ds = window(ds, size=1)  # by site
```
tomwhite, line 369: By site is the default behaviour, so you should be able to miss out this line.

PR #560, opened 2021-05-10
Rogers Huff R

tomwhite:
New version of #536 following the main branch rename.

### DIFF
--- a/sgkit/stats/ld.py
+++ b/sgkit/stats/ld.py
@@ -7,6 +7,7 @@ from numba import njit
 from sgkit.typing import ArrayLike
 
 
+# Note that float32 types are used in this function to get an exact match with scikit-allel's equivalent function.
 @njit(nogil=True, fastmath=True, cache=True, locals={"m0": numba.float32, "m1": numba.float32, "v0": numba.float32, "v1": numba.float32, "cov": numba.float32, "n": numba.int32})  # type: ignore
 def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
     """Rogers Huff R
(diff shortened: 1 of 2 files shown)


tomwhite:
Rebased, and updated to include suggestions from @jeromekelleher here: https://github.com/pystatgen/sgkit/pull/536#discussion_r626705276

jeromekelleher:
Looks good @tomwhite - we could do with a few parametrized tests with simulated data and a comparison to scikit-allele though? Looks like we're just testing a few corner cases here at the moment.

tomwhite:
Thanks @jeromekelleher - I agree about more tests. See [this comment](https://github.com/pystatgen/sgkit/pull/536#issuecomment-828590034) - it needs a bit more infrastructure, so I was thinking about doing it in a separate issue.

jeromekelleher:
> For a more comprehensive test, I agree using simulated data from msprime would be ideal here, using something similar to what we do in test_popgen.py. We could even integrate it with hypothesis to generate more extensive sets of test data. I think it might be worth factoring out a test utility for that, perhaps in a separate issue?

Aha, yes, I missed this. Yes, totally agree sounds great.

jeromekelleher (approved):

jeromekelleher:
I'm still sceptical about the relative value of float32 vs float64 btw - in my experience there's really not much of a gain unless you're working with really vast arrays (or you're on a GPU), and the loss of precision can bite you on the ass badly. I'd prefer if we had a "float64 unless there's a good reason" policy.

tomwhite:
I had a further look at the float32 vs float64 issue. I put it in originally to get an exact match with scikit-allel's `rogers_huff_r_between` function in all cases. I think we can relax this though and use float64 here, and still get an exact match in almost all cases - the exception being where we have two identical arrays with a large enough columns that causes scikit-allel to return NaN and the float64 implementation to return 1.0.

I've added a test for this, and also updated the LD prune PR to avoid hitting this case.

PR #372, opened 2020-11-03
Take contigs into account for fixed-size windowing

tomwhite:
Fixes #335

PR #561, opened 2021-05-10
LD prune

tomwhite:
New version of #532 following the main branch rename.

### DIFF
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -26,6 +26,7 @@ from .stats.aggregation import (
 from .stats.association import gwas_linear_regression
 from .stats.conversion import convert_probability_to_call
 from .stats.hwe import hardy_weinberg_test
+from .stats.ld import ld_matrix, ld_prune, maximal_independent_set
 from .stats.pc_relate import pc_relate
 from .stats.pca import pca
 from .stats.popgen import (
@@ -68,6 +69,9 @@ __all__ = [
     "infer_call_ploidy",
     "infer_sample_ploidy",
     "infer_variant_ploidy",
+    "ld_matrix",
+    "ld_prune",
+    "maximal_independent_set",
     "sample_stats",
     "variant_stats",
     "diversity",
--- a/sgkit/stats/ld.py
+++ b/sgkit/stats/ld.py
@@ -1,9 +1,18 @@
 import math
+from typing import Any, Hashable, List, Optional
 
+import dask
+import dask.array as da
+import dask.dataframe as dd
 import numpy as np
+import pandas as pd
+from dask.dataframe import DataFrame
 from numba import njit
+from xarray import Dataset
 
-from sgkit.typing import ArrayLike
+from sgkit import variables
+from sgkit.typing import ArrayLike, DType
+from sgkit.window import _get_chunked_windows, _sizes_to_start_offsets, has_windows
 
 
 @njit(nogil=True, fastmath=True, cache=True)  # type: ignore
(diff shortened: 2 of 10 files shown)


sgkit/stats/ld.py
```diff
@@ -61,3 +70,377 @@ def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: n
...
+
+    >>> # Divide into windows of size five (variants)
+    >>> ds = sg.window(ds, size=5)
+
+    >>> pruned_ds = sg.ld_prune(ds)
+    >>> pruned_ds.dims["variants"]
+    6
+    """
+    ldm = ld_matrix(ds, dosage=dosage, threshold=threshold, ld_score=ld_score)
+    mis = maximal_independent_set(ldm)
```
eric-czech, line 445: What happens if the variants in the provided dataset are on different contigs?  Will the windowing prevent comparisons of variants at, say, the end of chr1 to those at the beginning of chr2?

Something that may be worth mentioning in the docs here is that this MIS algorithm will pull the entire ld_matrix data frame into memory.  The MIS steps could run separately on different contigs, but afaik there is no distributed MIS algorithm that didn't look pretty hard to do on dask.

sgkit/stats/ld.py
```diff
@@ -61,3 +70,377 @@ def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: n
...
+    """Prune variants in linkage disequilibrium (LD).
+
+    This method uses a sparse LD matrix to find a maximally independent set (MIS)
+    of variants, and returns a dataset containing only those variants.
+
+    No information about which variants are pruned is returned by this method.
+    Consider using :func:`sgkit.ld_matrix` and :func:`sgkit.maximal_independent_set`
+    to get more insight into the variants that are pruned.
+
+    Note: This result is not a true MIS if ``ld_score`` is provided and are based on
```
eric-czech, line 397:
```suggestion
    Note: This result is not a true MIS if ``ld_score`` is provided and comparisons are based on
```

sgkit/stats/ld.py
```diff
@@ -61,3 +70,377 @@ def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: n
...
 def rogers_huff_r2_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: no cover
     return rogers_huff_r_between(gn0, gn1) ** 2  # type: ignore
+
+
+def ld_matrix(
+    ds: Dataset,
+    *,
+    dosage: Hashable = variables.dosage,
+    threshold: Optional[float] = None,
+    ld_score: Optional[Hashable] = None,
```
eric-czech, line 80: 🤔  Not a strong opinion but `ld_score` is maybe a little too polysemous for what it represents here?  Perhaps `variant_score` or just `score` instead?

sgkit/__init__.py
```diff
@@ -26,6 +26,7 @@
 from .stats.association import gwas_linear_regression
 from .stats.conversion import convert_probability_to_call
 from .stats.hwe import hardy_weinberg_test
+from .stats.ld import ld_matrix, ld_prune, maximal_independent_set
```
eric-czech, line 29: 🎉   🙌

sgkit/stats/ld.py
```diff
@@ -61,3 +70,377 @@ def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: n
...
+    - ``cmp``: If ``ld_score`` is provided, this is 1, 0, or -1 indicating whether or not ``i > j`` (1), ``i < j`` (-1), or ``i == j`` (0)
+
+    Raises
+    ------
+    ValueError
+        If the dataset is not windowed.
+    """
+
+    if not has_windows(ds):
+        raise ValueError("Dataset must be windowed for ld_matrix")
```
eric-czech, line 120: Mention the `sgkit.window` function so users know what to do next?

sgkit/stats/ld.py
```diff
@@ -61,3 +70,377 @@ def rogers_huff_r_between(gn0: ArrayLike, gn1: ArrayLike) -> float:  # pragma: n
...
+    """Compute a sparse linkage disequilibrium (LD) matrix.
+
+    This method computes the Rogers Huff R2 value for each pair of variants in
+    a window, and returns those that exceed the provided threshold, as a sparse
+    matrix dataframe.
+
+    Parameters
+    ----------
+    ds
+        Dataset containing genotype dosages.
```
eric-czech, line 91: It would be nice IMO if the windowing requirement and how to add them was elevated to somewhere like this in the docstrong rather than being implicit in "Raises".

eric-czech (commented):
Love it @tomwhite!  Some of the rewritten parts are much more clear now.

The only real question I have is whether or not comparisons across contigs are possible with this code.  It may also be necessary to break up the MIS calculations by contig sooner than later to make them work on real data.  That part could be the responsibility of the user for a while though IMO.

sgkit/stats/ld.py
hammer: Yes `LD score` is a term of art with a precise definition (cf [slide 15](http://ibg.colorado.edu/cdrom2016/nealeB/LDScore/LDScore_2016_Boulder.pdf)) so we should use a different name here.

sgkit/tests/test_ld.py
```diff
@@ -38,3 +55,156 @@ def test_rogers_huff_r_between():
...
+@settings(max_examples=50, deadline=None, phases=PHASES_NO_SHRINK)
+@example(args=(np.array([[1, 1], [1, 1]], dtype="uint8"), 1, 1, 0.0, -1))
+def test_vs_skallel(args):
+    x, size, step, threshold, chunks = args
+
+    ds = simulate_genotype_call_dataset(n_variant=x.shape[0], n_sample=x.shape[1])
+    ds["dosage"] = (["variants", "samples"], da.asarray(x).rechunk({0: chunks}))
+    ds = window(ds, size, step)
+
+    ldm = ld_matrix(ds, threshold=threshold)
```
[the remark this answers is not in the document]
tomwhite: Glad you raised this, since I added the check and it found some duplicates! I'd missed a case where windows overlap, which I've fixed now.

sgkit/tests/test_ld.py
```diff
@@ -38,3 +55,156 @@ def test_rogers_huff_r_between():
...
+
+@given(args=ld_prune_args())  # pylint: disable=no-value-for-parameter
+@settings(max_examples=50, deadline=None, phases=PHASES_NO_SHRINK)
+@example(args=(np.array([[1, 1], [1, 1]], dtype="uint8"), 1, 1, 0.0, -1))
+def test_vs_skallel(args):
+    x, size, step, threshold, chunks = args
+
+    ds = simulate_genotype_call_dataset(n_variant=x.shape[0], n_sample=x.shape[1])
+    ds["dosage"] = (["variants", "samples"], da.asarray(x).rechunk({0: chunks}))
+    ds = window(ds, size, step)
```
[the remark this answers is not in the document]
tomwhite: No, I didn't see that issue. Looks like we don't have it, as you point out.

sgkit/stats/ld.py
tomwhite: > Will the windowing prevent comparisons of variants at, say, the end of chr1 to those at the beginning of chr2?

Yes, since windowing is contig-aware (#372).

> Something that may be worth mentioning in the docs here is that this MIS algorithm will pull the entire ld_matrix data frame into memory.

Done. And I opened #569 for partitioning MIS by contig.

sgkit/stats/ld.py
tomwhite: Good point. Renamed to `variant_score` (including the name of the variable).

sgkit/stats/ld.py
tomwhite: Done

sgkit/stats/ld.py
tomwhite: Done

tomwhite:
Thanks for the review @eric-czech! I think I've addressed all the feedback now.

eric-czech (approved):

eric-czech:
One last thought: Is it worth a warning here and in `window` that positions have to be sorted?

tomwhite:
> Is it worth a warning here and in `window` that positions have to be sorted?

That will be true when we have windowing based on physical position, but currently `window` only works with "index" units. I'm going to open a PR soon for windowing by physical units.

PR #562, opened 2021-05-11
Fix issues in Tajima's D calculation

jeromekelleher:
Updated version of #542

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -554,8 +554,11 @@ def Tajimas_D(
 
     ac = ds[variant_allele_count]
 
-    # count segregating
-    S = ((ac > 0).sum(axis=1) > 1).sum()
+    # count segregating. Note that this uses the definition in tskit,
+    # which is the number of alleles - 1. In the biallelic case this
+    # gives us the number of non-monomorphic sites.
+    # FIXME this needs to be windowed: # 563
+    S = ((ac > 0).sum(axis=1) - 1).sum()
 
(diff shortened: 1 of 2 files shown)


jeromekelleher:
I've updated this @tomwhite, would you mind taking a look? There's some icky compromises in there to get things working. I tried various other approaches, but all of them seemed worse. Happy to consider alternatives.

sgkit/stats/popgen.py
```diff
@@ -583,15 +586,21 @@ def Tajimas_D(
...
+
+    # FIXME there doesn't appear to be a way to do the np.errstate trick
+    # below in Dask (https://github.com/dask/dask/issues/3245) so we call
+    # compute() and turn the values into numpy arrays. Since this is a simple
+    # computation this is unlikely to be an issue in practical terms, but
+    # may lead to surprises for users if they expect all the computations
+    # done in sgkit to be lazy.
+    d = np.array(d.compute())
+    d_stdev = np.array(d_stdev.compute())
+    with np.errstate(invalid="ignore", divide="ignore"):
```
tomwhite, line 599: The Dask issue is about controlling warnings, rather than the computation not working, so I wonder if we could remove the `with` block completely, and live with the warnings. I think that would be preferable to the eager computation.

To make the test pass, you can just add the following to `test_Tajimas_D`:

```python
@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```

tomwhite (approved):

PR #567, opened 2021-05-17
Add GPU support in the CI via Cirun.io

aktech:
Add support for provisioning GPU enabled runners on the GCP, this also adds the `cirun.yml` file which contains the config for provisioning runners.

The reference documentation for the `cirun.yml` can be seen here: https://docs.cirun.io/reference/yaml.html (It might be missing a thing or two, so I'll answer any questions if required).

Notes:
- Currently the runner provisioning takes anywhere from 5-7 minutes, that time is mostly installation of nvidia-drivers. In future, I'll create a base VM image with these already installed, so that it doesn't takes this long.
- The `cirun.yml` also has a flag for using `preemptible` instance. I have disabled that for now because preemptible instances are quite less reliable, I'll see if there is anything I can do to help there.
- This will fail until the GCP service account doesn't have `ComputeAdmin` permissions

Example run on my fork: https://github.com/aktech/sgkit/actions/runs/848823856

cc @jeromekelleher @tomwhite


.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,46 @@
+name: Build GPU
+
+on:
+  push:
+  pull_request:
```
ravwojdyla, line 5: Do we want to run this on every single PR and all push events to all branches? That's probably unnecessary and potentially $$$.

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Install dependencies
+        shell: bash -l {0}
+        run: |
+          conda install -c conda-forge gsl
+          conda install -c numba/label/dev numba
+          conda install -c anaconda cudatoolkit
+
+          pip install -r requirements.txt
+          pip install -r requirements-dev.txt
```
ravwojdyla, line 41: You could make this a single:

```bash
```

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set up Python
+        uses: conda-incubator/setup-miniconda@v2
+        env:
+          CONDA: /home/runnerx/miniconda3
+        with:
+          python-version: ${{ matrix.python-version }}
+          miniconda-version: "latest"
+
+      - name: Install dependencies
+        shell: bash -l {0}
```
ravwojdyla, line 34: You can probably make this a default via:

```
  runs-on: ubuntu-latest
    defaults:
      run:
         shell: bash -l {0}
```

?

cirun.yml
```diff
@@ -0,0 +1,22 @@
+# Self-Hosted Github Action Runners on GCP via Cirun.io
```
ravwojdyla, line 1: Random question - does this file need to exist at the root of the repo? Also does it have to be visible, as opposed to say `.cirun.yml`?

ravwojdyla (changes requested):

.github/workflows/build-gpu.yml
aktech: > Do we want to run this on every single PR and all push events to all branches? That's probably unnecessary and potentially $$$.

It isn't necessary at all. We can just run once daily as well.

Here is a rough estimate of the costs:

## GPU & Instance Pricing
`n1-standard` = $0.04749975 per hour
`NVIDIA® Tesla® T4` = $0.35 per
Total = $ 0.3975 per hour ~ 0.006625 per minute

Source: 
- https://cloud.google.com/compute/gpus-pricing
- https://cloud.google.com/compute/vm-instance-pricing#n1_predefined

##  GPU Tests Run cost (per push two builds)
Each run is about 12 min x 2 (Python 3.7, 3.8) = 24 min x 0.006625 = **$0.159**

- 1 Builds a day = $0.159 x 1 x 30 ~ **$4.77 per month**
- 10 Builds a day = $0.159 x 10 x 30 ~ **$47.7 per month**

.github/workflows/build-gpu.yml
aktech: Sounds good!

cirun.yml
aktech: > does this file need to exist at the root of the repo? 

Yes, at the moment Cirun doesn't have the ability to scan the whole repo for config.

> Also does it have to be visible, as opposed to say .cirun.yml?

No, I can make it invisible too, that makes more sense indeed.

ravwojdyla: > Yes, at the moment Cirun doesn't have the ability to scan the whole repo for config.

Even in `.github/cirun.yml` would be great and make sense.

But it's not a big deal, just curious :)

.github/workflows/build-gpu.yml
ravwojdyla: I don't have the build stats on hand, but the current config is definitely an overkill. A lagom setup could be:

```yaml
on:
  workflow_dispatch:
  push:
    branches:
      - main
  pull_request:
```

Which is PR changes + push to `main` + manual trigger. And if we want to save $, we could remove the PR trigger. But I will let other comment how much we are willing to spend here. ping @hammer

cirun.yml
aktech: > Even in .github/cirun.yml would be great and make sense.

Makes sense, I initially considered it, but was worried in case it conflicts with anything github related, but now looking at the docs I think it shouldn't. I'll incorporate this soon, but cannot promise in this PR.

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,49 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+  schedule:
+    # Run at the end of every day
+    - cron: "0 0 * * *"
```
ravwojdyla, line 11: If we run it on pus to `main` and all PR. Scheduled run is probably not necessary? wdyt?

aktech: I agree with you. This was done for consistency with [CPU nightly runs](https://github.com/pystatgen/sgkit/blob/main/.github/workflows/build.yml#L8). I don't have a strong opinion on either.

aktech:
@hammer I guess this is waiting on your input here: https://github.com/pystatgen/sgkit/pull/567#discussion_r633478474

.github/workflows/build-gpu.yml
hammer: I'm fine with a minimal setup for now so removing the PR trigger and just firing on pushes to `main` and manually works for me.

aktech:
@hammer @ravwojdyla changed the triggers.

.github/workflows/build-gpu.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  build:
+
+    runs-on: [self-hosted, gpu]
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
```
ravwojdyla, line 18: In the meantime we have also added 3.9 support

.cirun.yml
```diff
@@ -0,0 +1,22 @@
...
+    # Cheapest VM on GCP, with GPU attachable
+    instance_type: n1-standard-1
+    # Ubuntu-20.4, can be seen from "gcloud compute images list"
+    machine_image: ubuntu-minimal-2004-lts
+    # preemptible instances seems quite less reliable.
+    preemptible: false
+    # Number of runners to provision on every trigger on Actions job
+    # 2 because testing on Python 3.7 & 3.8
+    # See .github/workflows/build-gpu.yml
+    count: 2
```
ravwojdyla, line 18: Curious: if this was set to 1, would the tests effectively run sequentially?

ravwojdyla (approved):

.cirun.yml
aktech: Yes, unless the machine gets killed for being idle for too long.

.github/workflows/build-gpu.yml
aktech: Yep, adding.

tomwhite:
@aktech I think this needs rebasing now the build issues we had last week have been fixed (#591).

aktech:
This can be merged now.

ravwojdyla:
thank you @aktech 👍

PR #577, opened 2021-05-19
Add support for Python 3.9 in the CI

### LINKED ISSUE #310: Add Python 3.9 in the CI
Since Python 3.9 is released: https://docs.python.org/3/whatsnew/3.9.html

It would be good to start testing this in CI as early as we can, we should add something to mark this is **expected to fail**, but would be good to know what's failing.

aktech:
The tests seems to be passing.

Fixes #310

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


aktech:
@tomwhite All tests seems to pass in here.

tomwhite (approved):
That's great! Thanks @aktech.

Issue #340, opened 2020-10-20
Windowing functions should be lazy


tomwhite:
See https://github.com/pystatgen/sgkit/pull/303#discussion_r507906940

tomwhite:
Currently, window variables (`window_{contig,start,stop}`) are numpy arrays, which as Eric points out in the comment do not scale well to 100M variants. I think there are two things to do:
1. Make the windowing functions construct Dask arrays for the window variables so that they can be computed more efficiently.
2. Make window computations (`window_statistic` and `ld_matrix`) use Dask arrays.

For 1. we need a Dask implementation of [`np.searchsorted`](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html), both for finding contig boundaries, and for finding windows by genomic position. See https://github.com/dask/dask/pull/7696

2 is more work, since we can't materialize the window arrays as we do now. Instead, the strategy is probably: rechunk the window arrays to match the target variable chunks (e.g. genotypes for computing some of the popgen stats), then use `map_overlap` to process the target variable and the window definitions chunk by chunk. We'll also need a Dask delayed version of this in cases where we can't use `map_overlap`, such as when using `window_by_position` where the overlap depth is not bounded. This is similar to the current implementation in `ld_matrix`, which has the additional complication of returning a Dask dataframe, rather than a Dask array like `map_overlap` does.

So it might be worth having a couple of utility functions, `map_windows` and a more flexible Dask delayed version.

jeromekelleher:
It's not obvious to me that there really is a scaling issue here - have we done some experiments to see what sort of time these windowing functions take? Sounds like the dask versions are doing to be fairly tricky so might be worth doing a few quick checks before embarking on it?

tomwhite:
Closing this for the time being, can re-open if we see scalability issues.

Issue #434, opened 2021-01-05
Genomic coordinate systems


tomwhite:
Are genomic positions in sgkit 0-based or 1-based? We should decide and document the coordinate system conventions that sgkit uses.

[Please correct any mistakes I've made below!]

When reading VCF, PLINK, and BGEN files we leave the position variables unchanged, so they use the convention of the underlying file format. These are as follows:

- VCF is 1-based, fully closed http://samtools.github.io/hts-specs/VCFv4.3.pdf
- PLINK (.bim) is 1-based https://www.cog-genomics.org/plink/2.0/formats#bim
- BGEN _(I can't find documentation saying which convention this uses)_

In [`simulate_genotype_call_dataset`](https://pystatgen.github.io/sgkit/latest/generated/sgkit.simulate_genotype_call_dataset.html#sgkit.simulate_genotype_call_dataset) we create positions starting at 0, which implies they are 0-based. So this is inconsistent with reading the file formats above.

For comparison here are the conventions that other libraries and tools use.

- scikit-allel is 1-based. VCF POS is not changed, and functions are documented as being 1-based ([example](https://scikit-allel.readthedocs.io/en/stable/stats/diversity.html#allel.sequence_diversity)).
- tskit is 0-based. See [discussion here](https://tskit.readthedocs.io/en/latest/python-api.html#tskit.TreeSequence.write_vcf).
- Glow is 0-based, at least for VCF. It [converts VCF file positions to 0-based internally](https://github.com/projectglow/glow/blob/f6476a924153b02fa29634580f1b467a4d48b714/core/src/main/scala/io/projectglow/vcf/VCFLineToInternalRowConverter.scala#L153), (see this section marked ['Important'](https://glow.readthedocs.io/en/latest/etl/variant-data.html)).
- Hail is 1-based. See [docs](https://hail.is/docs/0.2/genetics/hail.genetics.Locus.html#hail.genetics.Locus.position), and this [warning](https://hail.is/docs/0.2/methods/impex.html#hail.methods.import_bed).

Whichever convention we adopt, we will likely need to support multiple ways of specifying intervals for selection, or windowing, which may use either convention. (This [article](https://gatk.broadinstitute.org/hc/en-us/articles/360035531852-Intervals-and-interval-lists) has a good summary of options.)

[...]

PR #581, opened 2021-05-25
Rename window to window_by_index, and add window_by_position

tomwhite:
This is a draft implementation of the design sketched in https://github.com/pystatgen/sgkit/issues/341#issuecomment-847140397

There is still work to change existing `window` references to `window_by_index`. Also, the windowing code does not use Dask arrays, see #340 for follow on work to fix that.

### DIFF
--- a/sgkit/window.py
+++ b/sgkit/window.py
@@ -91,9 +91,6 @@ def window_by_variant(
     return _window_per_contig(ds, variant_contig, merge, _get_windows, size, step)
 
 
-window = window_by_variant
-
-
 def window_by_position(
     ds: Dataset,
     *,
--- a/sgkit/window.py
+++ b/sgkit/window.py
@@ -1,9 +1,10 @@
-from typing import Any, Callable, Iterable, Optional, Tuple, Union
+from typing import Any, Callable, Hashable, Iterable, Optional, Tuple, Union
 
 import dask.array as da
 import numpy as np
 from xarray import Dataset
 
+from sgkit import variables
 from sgkit.utils import conditional_merge_datasets, create_dataset
 from sgkit.variables import window_contig, window_start, window_stop
 
@@ -12,27 +13,32 @@ from .typing import ArrayLike, DType
 # Window definition (user code)
 
 
-def window(
+def window_by_index(
     ds: Dataset,
+    *,
     size: int,
     step: Optional[int] = None,
+    variant_contig: Hashable = variables.variant_contig,
     merge: bool = True,
 ) -> Dataset:
-    """Add fixed-size windowing information to a dataset.
+    """Add window information to a dataset, measured by number of variants.
 
     Windows are defined over the ``variants`` dimension, and are
     used by some downstream functions to calculate statistics for
-    each window.
+    each window. Windows never span contigs.
 
     Parameters
     ----------
     ds
         Genotype call dataset.
     size
-        The window size (number of variants).
+        The window size, measured by number of variants.
     step
         The distance (number of variants) between start positions of windows.
         Defaults to ``size``.
+    variant_contig
+        Name of variable containing variant contig indexes.
+        Defined by :data:`sgkit.variables.variant_contig_spec`.
     merge
         If True (the default), merge the input dataset and the computed
         output variables into a single dataset, otherwise return only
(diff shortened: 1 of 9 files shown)


sgkit/window.py
```diff
@@ -12,27 +13,32 @@
 # Window definition (user code)
 
 
-def window(
+def window_by_index(
```
jeromekelleher, line 16: Not a strong opinion here, but have you considered ``window_by_variant``? The ``index`` suffix here might refer to a number of things, and there would be a nice symmetry then between ``window_by_variant`` and ``window_by_position``.

sgkit/window.py
```diff
@@ -43,12 +49,84 @@ def window(
...
+
+
+window = window_by_index
+
+
+def window_by_position(
+    ds: Dataset,
+    *,
+    size: int,
+    offset: int = 0,
```
jeromekelleher, line 70: could we add a ``num`` (or similar) option which is mutually exclusive to ``size`` that gives the number of windows? This would be really useful for plotting - I want to plot around 200 windows, e.g. 

So the ``size`` argument corresponds to numpy.arange and the ``num`` argument corresponds to ``numpy.linspace``? 

Haven't thought through how ``offset`` fits in here.

jeromekelleher (commented):
LGTM! Some comments.

sgkit/window.py
tomwhite: I hadn't thought of it, but `window_by_variant` might be better.

sgkit/window.py
tomwhite: `window_by_index` uses `np.arange`, so perhaps a `num` argument makes more sense there?

jeromekelleher: Yes, sorry, this is what I meant

jeromekelleher: You could want "num" in either - maybe I want ``n`` windows with the same number of variants (or as close as possible) or I want ``n`` equally spaced windows along the genome. Both are valid things, I think, and would be useful.

tomwhite: True. I think we could add `num` later, since it's fairly easy to work around. For `window_by_index`, you can set `size=ds.dims["variants"] // num`. For `window_by_position` you can work out the size by adding up the lengths of the contigs and dividing by `num`.

However, this does raise a difference in the types of windowing when measuring by genomic position. The `window_by_position` function here produces one window per variant (position), which is what we want for computing `ld_matrix` (which came from Eric's implementation, and was influenced by Hail's [locus_windows](https://hail.is/docs/0.2/linalg/utils/index.html#hail.linalg.utils.locus_windows)). This is subtly different to chopping the genome up into equally spaced windows measured by number of base pairs. I think we need both, but I'm not sure what the second one looks like from an API point of view (or what it should be called!) - particularly as it probably involves #434 as well.

tomwhite: Thinking about this more, I think the two ways of looking at windowing when measuring by genomic position can be generalized. I've introduced a variable called `window_start_position` that can be used to set window starts - so setting it to `variant_position` for example will create a window per variant (the `ld_matrix` case). Alternatively, leaving it unset will produce equally-spaced windows of size `size` across the genome (Jerome's case).

How does this look?

tomwhite:
When discussing this issue in [last week's meeting](https://github.com/pystatgen/sgkit/discussions/604), it was clear that there was some confusion about naming and what these functions do. I've added some examples and more documentation which hopefully should help.

jeromekelleher (commented):
Sorry for the delay in getting back on this @tomwhite, but I really wanted to get what you're doing here. I think I'm still missing something basic though - see the comment above.

jeromekelleher:
Dammit, github lost my comment! Basically I'm confused about why ``window_start`` and ``window_stop`` aren't genome coordinates, and why we won't have a ``window_index`` variable (which gives the index of the window for each variant). OK, it's a bit more storage space, but it would be simpler to work with, wouldn't it?

For window_by_position, consider the following example. What if we wanted to plot diversity in windows along the genome for two different datasets. If the window start/stop coords are based on the variant positions, then this will be quite messy.

I feel like I must be misunderstanding something though, so apologies for being slow here..

tomwhite:
I think the answer to why `window_start` and `window_stop` aren't genome coordinates is because we started with windowing by variant (index) and not taking coordinates into account. (I think of them as "internal" variables that other methods use.) I'm not sure that storing them as genome coordinates helps much though, since they would have to be converted to indexes at some point anyway. And we can support genome coordinates by using `np.searchsorted` to compute  `window_start` and `window_stop` from coordinates (as this PR does).

> What if we wanted to plot diversity in windows along the genome for two different datasets. If the window start/stop coords are based on the variant positions, then this will be quite messy.

Would using `window_by_position` on each dataset, then plotting each (on the same chart) not work? It might be worth trying an example to see how it looks.

BTW I'm away all next week, but look forward to discussing this more when I get back!

jeromekelleher:
> Would using window_by_position on each dataset, then plotting each (on the same chart) not work? It might be worth trying an example to see how it looks.

It probably would, but what if I wanted to compare the values directly? I think it's worth at least thinking through the alternative approach to windowing where we store the index of the window each variant belongs to.

> BTW I'm away all next week, but look forward to discussing this more when I get back!

Have fun! And yes, I think it's definitely worth talking this through in more detail...

tomwhite:
> I think it's worth at least thinking through the alternative approach to windowing where we store the index of the window each variant belongs to.

Windows can overlap, so a variant may fall in an arbitrary number of windows. So I'm not sure how useful it would be as a representation.

hammer:
> I think it's worth at least thinking through the alternative approach to windowing where we store the index of the window each variant belongs to.

This sounds to me like a partitionining, which I suppose could be considered a special case of windowing. I can imagine certain workloads where we'd want to request a partioning that obeys certain characteristics, so perhaps we should file a separate issue to design an API and representation for that workload?

tomwhite:
> I think it'd definitely want the windows to be strictly comparable across the two datasets, which they wouldn't be if the coordinates didn't align.

We can do this within a single contig with something like the following:

```
>>> import sgkit as sg
>>> import numpy as np
>>> 
>>> # first dataset
>>> ds1 = sg.simulate_genotype_call_dataset(n_variant=5, n_sample=3, seed=0)
>>> ds1["variant_position"] = (["variants"], np.array([1, 4, 6, 8, 12]))
>>> ds1 = sg.window_by_position(ds1, size=5, offset=1)
>>> ds1
<xarray.Dataset>
Dimensions:             (windows: 3, variants: 5, alleles: 2, samples: 3, ploidy: 2)
Dimensions without coordinates: windows, variants, alleles, samples, ploidy
Data variables:
    window_contig       (windows) int64 0 0 0
    window_start        (windows) int64 0 2 4
    window_stop         (windows) int64 2 4 5
    variant_contig      (variants) int64 0 0 0 0 0
    variant_position    (variants) int64 1 4 6 8 12
    variant_allele      (variants, alleles) |S1 b'C' b'T' b'C' ... b'A' b'A'
    sample_id           (samples) <U2 'S0' 'S1' 'S2'
    call_genotype       (variants, samples, ploidy) int8 0 0 1 0 1 ... 0 1 1 1 1
    call_genotype_mask  (variants, samples, ploidy) bool False False ... False
Attributes:
    contigs:  [0]
>>> 
>>> # second dataset with different number of variants, and none in first window
>>> ds2 = sg.simulate_genotype_call_dataset(n_variant=4, n_sample=3, seed=1)
>>> ds2["variant_position"] = (["variants"], np.array([6, 11, 13, 17]))
>>> ds2 = sg.window_by_position(ds2, size=5, offset=1)
>>> ds2
<xarray.Dataset>
Dimensions:             (windows: 4, variants: 4, alleles: 2, samples: 3, ploidy: 2)
Dimensions without coordinates: windows, variants, alleles, samples, ploidy
Data variables:
    window_contig       (windows) int64 0 0 0 0
    window_start        (windows) int64 0 0 1 3
    window_stop         (windows) int64 0 1 3 4
    variant_contig      (variants) int64 0 0 0 0
    variant_position    (variants) int64 6 11 13 17
    variant_allele      (variants, alleles) |S1 b'T' b'C' b'T' ... b'A' b'T'
    sample_id           (samples) <U2 'S0' 'S1' 'S2'
    call_genotype       (variants, samples, ploidy) int8 1 0 1 0 1 ... 0 1 1 0 0
    call_genotype_mask  (variants, samples, ploidy) bool False False ... False
Attributes:
    contigs:  [0]
>>> 
>>> # positions of variants in window 2 (0-based) for each dataset
>>> ds1.sel(dict(variants=slice(ds1.window_start[2].values, ds1.window_stop[2].values))).variant_position.values
array([12])
>>> ds2.sel(dict(variants=slice(ds2.window_start[2].values, ds2.window_stop[2].values))).variant_position.values
array([11, 13])
```

So it should be possible to plot diversity for two datasets with something similar, although doing it for multiple contigs would be more complicated.

jeromekelleher:
> Windows can overlap, so a variant may fall in an arbitrary number of windows. So I'm not sure how useful it would be as a representation.

Right, I think this is what was bothering me and what I hadn't grokked. So if windows can overlap (and yes, it's better that they can), then what we have is a good approach.

We'll probably want some recipes and maybe some utility functions for doing simple plotting like I was talking about above, but these can be added later. Thanks for your patience!

jeromekelleher (approved):

tomwhite:
> Right, I think this is what was bothering me and what I hadn't grokked. So if windows can overlap (and yes, it's better that they can), then what we have is a good approach.

Thanks @jeromekelleher. (It wasn't until I started trying to implement your suggestion that I realised that overlapping windows caused a problem.)

PR #582, opened 2021-05-25
Document installation using conda

### LINKED ISSUE #518: Packaging
There's a number of things going on with packaging, so I thought it would be good to bring things together into a meta-issue here.

Packaging is complicated in Python, and to reach the largest audience we need to support both pip and conda in a robust and reliable manner across Linux, OSX and Windows. Ideally, we would like to automatically pull in all our dependencies and always ship a fully-featured library, but unfortunately this simply isn't possible given our dependencies. We will have to compromise. 

See also: 
- #427 conda package development and status
- #494 discussion of whether we should have IO libraries as optional in pip or not
- (Please add any further issues that would be relevant here)

# Pypi

We have been assuming up to now that the main distribution platform is pypi, and that binary wheels will solve our problems with complex dependencies. 

Binary wheels on pypi are very helpful, but we have to assume that users will sometimes need to build from source. For example, it will probably be a year or more before ARM Macs are well supported in the conda and pypi binary infrastructures, and new versions of Python will come along between releases. 

# Conda

Conda packages will only install binary packages, so at least we don't have to worry about users needing to compile complicated packages 

# Cyvcf2

There are packaging issues with this on both conda and pypi. On pypi, it seems to be quite hard to compile from source:
```
$ pip install cyvcf2 --no-binary cyvcf2
...
 x86_64-linux-gnu-gcc: error: cyvcf2/cyvcf2.c: No such file or directory
 x86_64-linux-gnu-gcc: fatal error: no input files
 compilation terminated.
 error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
```
On conda, there is only a package available on ``bioconda`` (which means we would also have to use bioconda, if we have a hard dependency on it)

# bed-reader

The main issue with bed-reader is that it is partly written in Rust, so that source compiles depend on the Rust toolchain:

```
$ time pip install bed-reader --no-binary bed-reader
Collecting bed-reader
  Downloading bed_reader-0.2.5.tar.gz (2.4 MB)
     |████████████████████████████████| 2.4 MB 929 kB/s 
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... error
    ERROR: Command errored out with exit status 1:
     command: /home/jk/tmp/mucking/venv/bin/python3 /tmp/tmpo6io6p31 prepare_metadata_for_build_wheel /tmp/tmplq_7qd7u
         cwd: /tmp/pip-install-sog5acr9/bed-reader
    Complete output (6 lines):
    
    Cargo, the Rust package manager, is not installed or is not on PATH.
    This package requires Rust and Cargo to compile extensions. Install it through
    the system's package manager or via https://rustup.rs/
    
    Checking for Rust toolchain....
    ----------------------------------------
ERROR: Command errored out with exit status 1: /home/jk/tmp/mucking/venv/bin/python3 /tmp/tmpo6io6p31 prepare_metadata_for_build_wheel /tmp/tmplq_7qd7u Check the logs for full command output.
```

At least the error is pretty comprehensible, but users shouldn't need to install the Rust toolchain to install sgkit.

Hopefully we can make a conda-forge package for bed-reader, which would make it possible to add it as a full dependency in our conda package (see https://github.com/conda-forge/staged-recipes/pull/14471)

# cbgen

cbgen also doesn't currently build from source in a clean venv:
```
$ pip install cbgen --no-binary cbgen

 building 'cbgen._ffi' extension
  creating build/temp.linux-x86_64-3.8/build
  creating build/temp.linux-x86_64-3.8/build/temp.linux-x86_64-3.8
  x86_64-linux-gnu-gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O2 -Wall -g -fstack-protector-strong -Wformat -Werror=format-security -g -fwrapv -O2 -g -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/home/jk/tmp/mucking/venv/include -I/usr/include/python3.8 -c build/temp.linux-x86_64-3.8/cbgen._ffi.c -o build/temp.linux-x86_64-3.8/build/temp.linux-x86_64-3.8/cbgen._ffi.o
  build/temp.linux-x86_64-3.8/cbgen._ffi.c:570:14: fatal error: bgen/bgen.h: No such file or directory
    570 |     #include "bgen/bgen.h"
        |              ^~~~~~~~~~~~~
  compilation terminated.
  error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
  ----------------------------------------
```

But, it does have a conda-forge package, which is great: https://anaconda.org/conda-forge/cbgen

# What do we do?

Conda doesn't have the concept of an "optional" extra, so our choices are simple here: either we create a bioconda package for sgkit and ship with all dependencies **or** we create a conda-forge package and when someone tries to use VCF functionality we raise an error that says "Please install cyvcf2: either `conda install -c bioconda cyvcf2` or `pip install cyvcf2`"

Hopefully we will have conda-forge packages for the other upstreams, and so we can ship a package with full dependencies on them.

The situation for pypi is more complex, but I think probably either keeping the current "optional extras" setup or just raising an error asking the user to install the specific package (like the one I suggest for cyvcf2 above) is the simplest and most reliable route. I accept that this isn't an optimal user experience for someone who wants to jump right in and analyse their VCF data (or whatever), but it's less bad than failing to install at all with a cryptic compiler error. 

I think a reasonable requirement would be that sgkit should install and be functional by running ``pip install sgkit`` on a ARM Mac. Nothing loses users more quickly than something that won't install, and there's nothing more frustrating that having to do compiler workarounds for software that you're not going to use, just to get the package to install. 

(Does anyone have one of these? It would probably be a good dogfooding exercise for one of us to go through the pain that will be involved in using it.)

tomwhite:
Fixes #518

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -19,9 +19,10 @@ Installation
 Sgkit can read standard genetic file formats, including VCF, PLINK, and BGEN.
 
 If sgkit has been installed using conda, support for reading BGEN and PLINK is included, but
-VCF is not, and should be installed using pip as follows::
+VCF is not because there is no Windows support for cyvcf2, the library we use for reading VCF data.
+If you are using Linux or a Mac, please install cyvcf2 using the following to enable VCF support::
 
-    $ pip install 'sgkit[vcf]'
+    $ conda install -c bioconda cyvcf2
 
 If sgkit has been installed using pip, then support for reading these formats is
 not included, and requires additional dependencies, which can be installed
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -16,8 +16,15 @@ Reading genetic data
 Installation
 ------------
 
-Sgkit can read standard genetic file formats, including VCF, PLINK, and BGEN. Support for reading
-these formats is not installed by default, and requires additional dependencies, which can be installed
+Sgkit can read standard genetic file formats, including VCF, PLINK, and BGEN.
+
+If sgkit has been installed using conda, support for reading BGEN and PLINK is included, but
+VCF is not, and should be installed using pip as follows::
+
+    $ pip install 'sgkit[vcf]'
+
+If sgkit has been installed using pip, then support for reading these formats is
+not included, and requires additional dependencies, which can be installed
 as an "extra" feature using pip, as follows.
 
 To install sgkit with BGEN support::
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -10,9 +10,13 @@ Getting Started
 Installation
 ------------
 
-You can install sgkit with `pip` - there is no conda package yet. Python 3.7 or 3.8 is required.
+You can install sgkit with pip. Python 3.7, 3.8, or 3.9 is required::
 
     $ pip install sgkit
+
+Alternatively, you can use conda::
+
+    $ conda install -c conda-forge sgkit
 
 ..
 
--- a/docs/vcf.rst
+++ b/docs/vcf.rst
@@ -30,7 +30,7 @@ Installation
 VCF support is an "extra" feature within sgkit and requires additional
 dependencies, notably ``cyvcf2``.
 
-To install sgkit with VCF support using pip::
+To install sgkit with VCF support using pip (there is no conda package)::
 
     $ pip install 'sgkit[vcf]'
 


ravwojdyla (approved):

docs/user_guide.rst
```diff
@@ -16,8 +16,15 @@ Reading genetic data
...
 ------------
 
-Sgkit can read standard genetic file formats, including VCF, PLINK, and BGEN. Support for reading
-these formats is not installed by default, and requires additional dependencies, which can be installed
+Sgkit can read standard genetic file formats, including VCF, PLINK, and BGEN.
+
+If sgkit has been installed using conda, support for reading BGEN and PLINK is included, but
+VCF is not, and should be installed using pip as follows::
+
+    $ pip install 'sgkit[vcf]'
```
jeromekelleher, line 24: I'm not sure this is the best advice - why not be up front and say "VCF support is not included because there is no Windows support for cyvcf2, the library we use for reading VCF data. If you are using Linux or a Mac, please install cyvcf2 using ``conda install -c bioconda cyvcf2`` to enable VCF support."

Unless we explain the situation clearly we'll end up answering endless questions from users with failing installations. (I.e., someone on a Mac who doesn't have a working C compiler and tries to install with pip, like above)

jeromekelleher (commented):
Looks good, but I think we should recommend conda users install cyvcf2 from bioconda.

docs/user_guide.rst
tomwhite: Thanks for the suggestion - I've updated the text.

jeromekelleher (approved):

PR #591, opened 2021-05-31
Fix change in make_meta in Dask 2021.5.1 #590

### LINKED ISSUE #590: Main build failing in bgen after upgrade to Dask 2021.5.1
See e.g. https://github.com/pystatgen/sgkit/runs/2706847930?check_suite_focus=true

tomwhite:
Fixes #590. It wasn't sufficient to just rename `make_meta` to `make_meta_util`, I had to also set `parent_meta`. 

I haven't tried to do anything for older versions of Dask - is is worth setting a minimum version in requirements?

The incompatible change was made in https://github.com/dask/dask/pull/7586.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy
 xarray
-dask[array]
-distributed
+dask[array] != 2021.5.1
+distributed != 2021.5.1
 dask-ml
 scipy
 typing-extensions
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,8 +28,8 @@ python_requires = >=3.7
 install_requires =
     numpy
     xarray
-    dask[array]
-    distributed
+    dask[array] != 2021.5.1
+    distributed != 2021.5.1
     dask-ml
     scipy
     zarr


eric-czech:
What do you think of doing something like https://github.com/rapidsai/cudf/pull/8368/files#diff-8cccf94d2261588b89ab7cd02573abfe37a19aab98fb6d210184c5ce293e9b3eR24-R28 (as @jakirkham suggested in https://github.com/dask/dask/pull/7586#issuecomment-849033120)?

I'm fine with a minimum version, it seems inevitable.  I do hope we can avoid needing to program in switches like https://github.com/dask/dask-ml/blob/3ef1c84fe02889d354247f72f476ae49a42bf321/dask_ml/_compat.py#L21-L26 (from dask-ml), but that may also be hard to avoid in some situations.  FWIW those different flags are used, it total, ~35 times in the dask-ml repo so it is apparently not easy to maintain backwards compatibility over longer time periods.

tomwhite:
> What do you think of doing something like https://github.com/rapidsai/cudf/pull/8368/files#diff-8cccf94d2261588b89ab7cd02573abfe37a19aab98fb6d210184c5ce293e9b3eR24-R28 (as @jakirkham suggested in [dask/dask#7586 (comment)](https://github.com/dask/dask/pull/7586#issuecomment-849033120))?

I was hoping to avoid this. It would need to be more than a change in the import, because there is an extra argument (`parent_meta`) that's needed now. The bigger problem is testing, by having two paths we'd need to have CI set up to test various Dask versions. We've been a big vague on this so far - we basically only test on the latest versions of our dependencies - which is OK for a young project I think, so I'd like to continue with that for a while if possible.

> I'm fine with a minimum version, it seems inevitable.

I'll add that.

> I do hope we can avoid needing to program in switches like https://github.com/dask/dask-ml/blob/3ef1c84fe02889d354247f72f476ae49a42bf321/dask_ml/_compat.py#L21-L26 (from dask-ml), but that may also be hard to avoid in some situations. FWIW those different flags are used, it total, ~35 times in the dask-ml repo so it is apparently not easy to maintain backwards compatibility over longer time periods.

Not pretty. But at least there is an attempt to maintain backwards compatibility, which is laudable.

galipremsagar (approved):

quasiben:
FYI, we are talking about a different way of handling these changes in https://github.com/dask/dask/pull/7743

tomwhite:
Thanks for pointing out that issue @quasiben.

tomwhite:
Dask is now being fixed to preserve the old behaviour (https://github.com/dask/dask/pull/7743), so I think we can just avoid using Dask 2021.5.1. I've updated this PR to fix the main build, and our upstream build should start working again once https://github.com/dask/dask/pull/7743 is merged.

jrbourbeau:
Just a heads up that https://github.com/dask/dask/pull/7743 has been merged (thanks again for testing that PR out @tomwhite)

PR #592, opened 2021-05-31
Fix tests failing due to Xarray dims ordering change

### LINKED ISSUE #583: Upstream failing due to Xarray dims ordering change
Xarray no longer sorts dimension by name (https://github.com/pydata/xarray/pull/4753). This affects the display representation, which is causing some upstream tests to fail. Example [failure](https://github.com/pystatgen/sgkit/runs/2661210916?check_suite_focus=true):

```
Differences (unified diff with -expected +actual):
    @@ -1,5 +1,5 @@
     <xarray.Dataset>
    -Dimensions:                  (alphas: 5, blocks: 2, contigs: 2, outcomes: 5, samples: 50)
    -Dimensions without coordinates: alphas, blocks, contigs, outcomes, samples
    +Dimensions:                  (blocks: 2, alphas: 5, samples: 50, outcomes: 5, contigs: 2)
    +Dimensions without coordinates: blocks, alphas, samples, outcomes, contigs
     Data variables:
         regenie_base_prediction  (blocks, alphas, samples, outcomes) float64 0.33...
```

I'm not sure how to fix this so that the tests work against both Xarray 0.18.2 and the main branch.

The `test_gwas_linear_regression__validate_statistics` test is also failing due to the same change, but the error is a numerical mismatch, which suggests the dimensions are somehow getting mixed up.

tomwhite:
There are two parts to this fix: one that removes an assumption of dimension ordering (in `test_gwas_linear_regression__validate_statistics`), and another that ignores the two doctests that fail due to the `repr` having changed. These ignores can be removed once the next release of Xarray is made, so I'll open an issue to do that if this fix looks OK.

Fixes #583

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -67,7 +67,8 @@ bgen =
 fail_under = 100
 
 [tool:pytest]
-addopts = --doctest-modules --ignore=validation --cov-fail-under=100
+# TODO: remove ignored doctests below once Xarray>0.18.2 is released (with https://github.com/pydata/xarray/pull/4753)
+addopts = --doctest-modules --ignore=validation --cov-fail-under=100 --ignore=sgkit/stats/regenie.py --ignore=sgkit/stats/pca.py
 norecursedirs = .eggs build docs
 filterwarnings =
     error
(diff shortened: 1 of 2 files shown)


tomwhite:
Failing due to #590

eric-czech (approved):

Issue #452, opened 2021-02-03
Move to NumPy's ArrayLike and DtypeLike


tomwhite:
Introduced in NumPy 1.20.0: https://numpy.org/doc/stable/release/1.20.0-notes.html#numpy-is-now-typed

These would replace our types in `sgkit.typing`.

PR #594, opened 2021-06-01
Use numpy.typing.DTypeLike

tomwhite:
This fixes half of #452. NumPy's `ArrayLike` isn't a drop-in replacement for ours, since it represents something that can be converted to a NumPy array (including lists, for example), whereas we use it in the sense of "something that looks like an array" (which could be a Dask array, for example).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
     - name: Check for Sphinx doc warnings
+      if: contains(matrix.python-version, '3.8')
       run: |
         cd docs
         make html SPHINXOPTS="-W --keep-going -n"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -94,7 +94,11 @@ exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
 
 autosummary_generate = True
 
-nitpick_ignore = [("py:class", "sgkit.display.GenotypeDisplay")]
+nitpick_ignore = [
+    ("py:class", "sgkit.display.GenotypeDisplay"),
+    ("py:class", "numpy.typing._dtype_like._DTypeDict"),
+    ("py:class", "numpy.typing._dtype_like._SupportsDType"),
+]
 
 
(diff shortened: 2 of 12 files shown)


eric-czech (approved):

tomwhite:
The test is failing due to some interaction between numpy.typing and scanpydoc.

tomwhite:
> The test is failing due to some interaction between numpy.typing and scanpydoc.

I think it is failing only on Python 3.7, since `SupportsIndex` is not there (introduced in Python 3.8), and Python 3.7 when called from scanpydoc doesn't use `typing-extensions`.

One fix might be to require Python >=3.8 for the doc build, so we could skip "Check for Sphinx doc warnings" in `build.yml` when running Python 3.7. (Not sure how easy it is to do that though.)

ravwojdyla (approved):

.github/workflows/build.yml
```diff
@@ -31,6 +31,7 @@ jobs:
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
     - name: Check for Sphinx doc warnings
+      if: contains(matrix.python-version, '3.8')
```
ravwojdyla, line 34: 👍

tomwhite: @ravwojdyla thanks for your (offline) suggestion to do this. The workflow syntax works perfectly. But the build is failing on 3.8 too, so I need to look into that!

tomwhite:
Various extra changes were needed to get Sphinx to build the docs. The type annotation is very off-putting for users though:

dtypes

I was expecting the annotation to be `DTypeLike`, linked to https://numpy.org/devdocs/reference/typing.html#numpy.typing.DTypeLike.

This problem was discussed in Pandas in https://github.com/pandas-dev/pandas/issues/33025. It might be worth investigating further to see if we can improve this.

ravwojdyla:
@tomwhite nice, should we merge this and investigate/improve in a separate PR/issue?

tomwhite:
@ravwojdyla I'd rather wait and see if there's a way to avoid the off-putting type signatures in docs, as I don' think this is an improvement as it stands.

PR #610, opened 2021-06-15
Deploy distribution artifacts to test PyPi on a tagged commit #606

### LINKED ISSUE #606: Deploy distribution artifacts to test PyPi on a tagged commit
Add a GitHub action that builds the distribution artifacts and deploys to test PyPi whenever a tagged commit is made.

References:
* https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
* https://github.com/pypa/gh-action-pypi-publish
* https://github.com/tskit-dev/tskit/blob/main/.github/workflows/wheels.yml

tomwhite:
Fixes #606 

This workflow builds and tests wheels for each combination of the following (except VCF on Windows):
* IO: BGEN, PLINK, VCF
* Python: 3.7, 3.8, 3.9
* OS: Linux, MacOS, Windows

If successful it will push to test PyPi if the commit is tagged. I've tested this works by running it in the `test` branch in this repository. (See https://github.com/pystatgen/sgkit/runs/2830563755?check_suite_focus=true and https://test.pypi.org/project/sgkit/0.3.0a2/)

If there is no tag the wheels are still built and tested for every push to main - do we want that?

### DIFF
--- a/.github/scripts/test_sgkit.py
+++ b/.github/scripts/test_sgkit.py
@@ -1,0 +1,5 @@
+import sgkit as sg
+
+if __name__ == "__main__":
+    ds = sg.simulate_genotype_call_dataset(n_variant=100, n_sample=50, n_contig=23)
+    print(ds)
--- a/.github/scripts/test_sgkit_bgen.py
+++ b/.github/scripts/test_sgkit_bgen.py
@@ -1,0 +1,11 @@
+import urllib.request
+
+from sgkit.io.bgen import read_bgen
+
+if __name__ == "__main__":
+    urllib.request.urlretrieve(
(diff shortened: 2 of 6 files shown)


jeromekelleher (approved):
LGTM!

jeromekelleher:
> If there is no tag the wheels are still built and tested for every push to main - do we want that?

I think we do - building the artefacts is a small cost and it's good to know early when your packaging breaks.

jeromekelleher:
Also good to add a ``twine check --strict`` somewhere in our CI pipelines, it does catch a good few problems.

PR #616, opened 2021-06-29
Xarray consolidate metadata change

### LINKED ISSUE #615: Use consolidated writes/reads by default
Upstream is failing since https://github.com/pydata/xarray/pull/5252 went in, which enables consolidated writes/reads by default.

See e.g. https://github.com/pystatgen/sgkit/runs/2927778139?check_suite_focus=true

```
E               RuntimeWarning: Failed to open Zarr store with consolidated metadata, falling back to try reading non-consolidated metadata. This is typically much slower for opening a dataset. To silence this warning, consider:
E               1. Consolidating metadata in this existing store with zarr.consolidate_metadata().
E               2. Explicitly setting consolidated=False, to avoid trying to read consolidate metadata, or
E               3. Explicitly setting consolidated=True, to raise an error in this case instead of falling back to try reading non-consolidated metadata.

/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/xarray/backends/zarr.py:371: RuntimeWarning
=========================== short test summary info ============================
FAILED sgkit/tests/test_regenie.py::test_regenie__glow_comparison - RuntimeWa...
FAILED sgkit/tests/test_vcfzarr_reader.py::test_vcfzarr_to_zarr[None-True-False-True]
FAILED sgkit/tests/test_vcfzarr_reader.py::test_vcfzarr_to_zarr[None-True-True-False]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__target_chunks[target_chunks0]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__target_chunks[target_chunks1]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__target_chunks[target_chunks2]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__target_chunks[target_chunks3]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_from_zarr__self_consistent
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__probability_encoding[uint8]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__probability_encoding[uint16]
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_rechunk_bgen__variable_packing
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_unpack_variables__invalid_gp_dims
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_bgen_to_zarr[None] - Run...
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_bgen_to_zarr[region1] - ...
FAILED sgkit/tests/io/bgen/test_bgen_reader.py::test_bgen_to_zarr[region2] - ...
FAILED sgkit/tests/io/vcf/test_vcf_reader.py::test_vcf_to_zarr__multiple_partitioned_by_size[True]
FAILED sgkit/tests/io/vcf/test_vcf_reader.py::test_vcf_to_zarr__multiple_partitioned_by_size[False]
```

tomwhite:
Fixes #615.

This uncovered a bug, where VCF part files with no variants were not being written, which manifested as an error with the new version of Xarray when it tried to consolidate their metadata. This change includes a fix so that empty part files are written (containing zero variants).

### DIFF
--- a/sgkit/io/bgen/bgen_reader.py
+++ b/sgkit/io/bgen/bgen_reader.py
@@ -518,6 +518,8 @@ def rechunk_bgen(
         )
         rechunked.execute()
 
+    zarr.consolidate_metadata(output)
+
     ds: Dataset = xr.open_zarr(output, concat_characters=False)  # type: ignore[no-untyped-call]
     if pack:
         ds = unpack_variables(ds)
--- a/sgkit/io/vcf/utils.py
+++ b/sgkit/io/vcf/utils.py
@@ -19,16 +19,23 @@ def ceildiv(a: int, b: int) -> int:
     return -(-a // b)
 
 
-# https://dev.to/orenovadia/solution-chunked-iterator-python-riddle-3ple
+# Based on https://dev.to/orenovadia/solution-chunked-iterator-python-riddle-3ple
 def chunks(iterator: Iterator[T], n: int) -> Iterator[Iterator[T]]:
     """
     Convert an iterator into an iterator of iterators, where the inner iterators
     each return `n` items, except the last, which may return fewer.
+
+    For the special case of an empty iterator, an iterator of an empty iterator is
+    returned.
     """
 
+    empty_iterator = True
     for first in iterator:  # take one item out (exits loop if `iterator` is empty)
+        empty_iterator = False
         rest_of_chunk = itertools.islice(iterator, 0, n - 1)
         yield itertools.chain([first], rest_of_chunk)  # concatenate the first item back
+    if empty_iterator:
+        yield iter([])
 
 
 def get_file_length(
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -63,6 +63,9 @@ Similarly, the function :func:`sgkit.save_dataset` can be used to save the datas
 This can be used to save a newly-loaded dataset, or a dataset with new variables that are the
 result of running genetic methods on the data, so it can be loaded again later.
 
+By default, Zarr datasets created by sgkit (and Xarray) have `consolidated metadata <http://xarray.pydata.org/en/stable/user-guide/io.html#consolidated-metadata>`_,
+which makes opening the dataset faster.
+
 BGEN
 ----
 
--- a/sgkit/io/utils.py
+++ b/sgkit/io/utils.py
@@ -74,14 +74,10 @@ def zarrs_to_dataset(
 
     datasets = []
     for path in urls:
-        try:
-            dataset = xr.open_zarr(  # type: ignore[no-untyped-call]
-                fsspec.get_mapper(path, **storage_options), concat_characters=False
-            )
-            datasets.append(dataset)
-        except zarr.errors.GroupNotFoundError:
-            # ignore empty directory, occurs if no variants in region
-            pass
+        dataset = xr.open_zarr(  # type: ignore[no-untyped-call]
+            fsspec.get_mapper(path, **storage_options), concat_characters=False
+        )
+        datasets.append(dataset)
 
     # Combine the datasets into one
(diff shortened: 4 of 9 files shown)


sgkit/io/bgen/bgen_reader.py
```diff
@@ -518,6 +518,8 @@ def rechunk_bgen(
         )
         rechunked.execute()
 
+    zarr.consolidate_metadata(output)
```
ravwojdyla, line 521: does this need to be document in the user guide?

sgkit/io/vcf/utils.py
```diff
@@ -19,16 +19,23 @@ def ceildiv(a: int, b: int) -> int:
     return -(-a // b)
 
 
-# https://dev.to/orenovadia/solution-chunked-iterator-python-riddle-3ple
+# Based on https://dev.to/orenovadia/solution-chunked-iterator-python-riddle-3ple
 def chunks(iterator: Iterator[T], n: int) -> Iterator[Iterator[T]]:
```
ravwojdyla, line 23: Nice.

For future I dig up some implementation I did some time again, in case there's any interest in that:

```
from itertools import chain, islice

import six
import typing


def batch(iterable: typing.Iterable, size: int) -> typing.Generator: 
    """
    For example:
    ```
    list(batch([1, 2, 3], 2)) -> [[1, 2], [3]]
    ```
    """
    if size <= 0:
        raise ValueError("Can't iterate over batch size <= 0")
    if iterable is None:
        raise ValueError("Iterable can't be None")
    if hasattr(iterable, "__len__"):
        # leverage len to do efficient batching
        length = len(iterable)  # type: ignore
        it = iter(iterable)
        for i in range(0, length, size):
            yield [k for k in islice(it, size)]
    else:
        # fallback to generator based method
        for chunk in batch_gen(iterable, size):
            yield chunk
        return


def batch_gen(generator: typing.Iterable, size: int, lazy: bool=False) -> typing.Generator:
    """
    For example:
    ```
    def foo_gen():
        for x in range(5):
            yield x

    list(batch(foo_gen(), 2)) -> [[0, 1], [2, 3], [4]]
    ```

    NOTE: default implementation is eager on the materialization of the elements, you can use lazy parameter to
    get batches of generators of elements, but keep in mind that lazy implementation is probably not thread safe, and
    batch element generator has to be exhausted before retrieving the next batch.
    """
    if size <= 0:
        raise ValueError("Can't iterate over batch size <= 0")
    if generator is None:
        raise ValueError("generator can't be None")
    it = iter(generator)
    while True:
        batch_it = islice(it, size)
        try:
            chunk_gen = chain([six.next(batch_it)], batch_it)
        except StopIteration:
            return
        if lazy:
            yield chunk_gen
        else:
            yield [x for x in chunk_gen]
```

ravwojdyla (approved):

sgkit/io/vcf/utils.py
tomwhite: Thanks @ravwojdyla! Is the batch code more efficient do you think? Something worth evaluating in the future perhaps.

ravwojdyla: @tomwhite afair, it might be if the iterable has length (`hasattr(iterable, "__len__")`).

jeromekelleher:
This is over my head - happy to defer to you guys here

sgkit/io/vcf/utils.py
tomwhite: Ah, we don't know the number of variants in a particular region, so I don't think we can use this optimization unfortunately.

PR #620, opened 2021-07-06
Issue a warning if the number of alt alleles exceeds the maximum specified

### LINKED ISSUE #487: Issue warning if more alternate alleles than max_alt_alleles when reading VCF
From https://github.com/pystatgen/sgkit/pull/471#discussion_r586239899:

> Users should know if their ALTs are being dropped, and ideally it should be discoverable later if they have been truncated. Could we add some metadata tracking when this has happened?

tomwhite:
Fixes #487. The maximum number of alleles found while parsing is stored in `ds.attrs["max_alt_alleles_seen"]` (in all cases).

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -24,6 +24,7 @@ from cyvcf2 import VCF, Variant
 from numcodecs import PackBits
 
 from sgkit import variables
+from sgkit.io.dataset import load_dataset
 from sgkit.io.utils import zarrs_to_dataset
 from sgkit.io.vcf import partition_into_regions
 from sgkit.io.vcf.utils import build_url, chunks, temporary_directory, url_filename
@@ -48,6 +49,12 @@ try:
 except ImportError:  # pragma: no cover
     warnings.warn("Cannot import Blosc, falling back to no compression", RuntimeWarning)
     DEFAULT_COMPRESSOR = None
+
+
+class MaxAltAllelesExceededWarning(UserWarning):
+    """Warning when the number of alt alleles exceeds the maximum specified."""
+
+    pass
 
 
 @contextmanager
@@ -362,6 +369,7 @@ def vcf_to_zarr_sequential(
         # Remember max lengths of variable-length strings
         max_variant_id_length = 0
         max_variant_allele_length = 0
+        max_alt_alleles_seen = 0
 
         # Iterate through variants in batches of chunk_length
 
@@ -413,6 +421,7 @@ def vcf_to_zarr_sequential(
                 variant_position[i] = variant.POS
 
                 alleles = [variant.REF] + variant.ALT
+                max_alt_alleles_seen = max(max_alt_alleles_seen, len(variant.ALT))
                 if len(alleles) > n_allele:
                     alleles = alleles[:n_allele]
                 elif len(alleles) < n_allele:
@@ -457,6 +466,7 @@ def vcf_to_zarr_sequential(
             if add_str_max_length_attrs:
                 ds.attrs["max_length_variant_id"] = max_variant_id_length
                 ds.attrs["max_length_variant_allele"] = max_variant_allele_length
+            ds.attrs["max_alt_alleles_seen"] = max_alt_alleles_seen
 
             if first_variants_chunk:
                 # Enforce uniform chunks in the variants dimension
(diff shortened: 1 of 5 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -705,6 +716,15 @@ def vcf_to_zarr(
         field_defs=field_defs,
     )
 
+    # Issue a warning if max_alt_alleles caused data to be dropped
+    ds = load_dataset(output)
+    max_alt_alleles_seen = ds.attrs["max_alt_alleles_seen"]
+    if max_alt_alleles_seen > max_alt_alleles:
```
eric-czech, line 722: I think it's worth noting in the docstring for `max_alt_alleles` that this is only applicable to the array of base changes defining the alleles, and not any of the actual call data.

eric-czech (approved):

sgkit/io/vcf/vcf_reader.py
tomwhite: Good point - I've updated the doc.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -640,7 +652,8 @@ def vcf_to_zarr(
         specified ploidy will raise an exception.
     max_alt_alleles
         The (maximum) number of alternate alleles in the VCF file. Any records with more than
-        this number of alternate alleles will have the extra alleles dropped.
+        this number of alternate alleles will have the extra alleles dropped (the `variant_allele`
+        variable will be truncated). Call genotype fields will however be unaffected.
```
jeromekelleher, line 656: So, in this case the user may get an index error if they use ``alleles[genotypes[..]]``, right? It might not be obvious to people that this just means they won't be able to find out the allelic value that corresponds to the genotype value. 

Code may break downstream in unexpected ways because of this as we violating a basic invariant of the data model. Maybe it would be better to mark subsequent alleles as missing data (-1)? At least well-written code would deal with this in a predictable way, whereas having to deal with arbitrary missing alleles at every step could be quite tedious.

jeromekelleher (commented):
See my comment about marking subsequent alleles as missing data - I think this would be easier to document and lead to fewer bugs in the long run.

sgkit/io/vcf/vcf_reader.py
tomwhite: Any thoughts on this @eric-czech?

ravwojdyla: +1 handling call data. I wonder if there should be a flag to handle the alt allele overflow - with options: `error`, `turncate`, `freq_truncate`?

tomwhite: Another thought I had was that we could store the entire ALT field as another variable (of type string), which could be used for reconstructing more alleles without having to regenerate the whole dataset. This might be orthogonal to what we are discussing here though.

I also wondered if we could use [Zarr ragged arrays](https://zarr.readthedocs.io/en/stable/tutorial.html#ragged-arrays), but I don't think you can combine them with variable length strings - and even if you could they don't play well with the general array data model, which assumes fixed array dimensions.

jeromekelleher: Ragged arrays would probably be the best solution in the long run, but that's a big change, as you say. Maybe we should break that discussion off into an issue, so we can explore the pros and cons?

tomwhite: Opened #634

tomwhite: Thinking about this more, if we stored the ALT field as another variable (just a single comma-delimited string, not a ragged array), then I think there is more of a case for keeping all the allele values in the genotype field, since it would be possible to find the allele string if needed. (This would also allow the fixed length `alleles` field to be regenerated if needed, without having to read all the data in from scratch.)

We could document the contract here (i.e. the possibility of indexing out of the `alleles` array), and also a pattern of how to use `xr.where` to ensure the genotypes don't index out of the `alleles` array. This would effectively do the truncation when accessing the data.

I'm also a bit wary of marking some genotypes as missing with value -1, since then it's not possible to distinguish if they were missing in the original file, or because they were truncated.

tomwhite:
Rebased, with no other changes for the time being.

sgkit/io/vcf/vcf_reader.py
jeromekelleher: What about the overflow issue @tomwhite? #640

tomwhite: Good point! That's a strong argument for truncating.

jeromekelleher: Indeed - clinches it in my view. We don't want two different behaviours when having different flavours of "allele out of bounds", do we?

tomwhite: Now fixed in the latest commit.

jeromekelleher (approved):
LGTM!

Issue #642, opened 2021-07-26
Use PackBits filter for boolean fields for Zarr storage


tomwhite:
When writing Zarr files we should enable the [PackBits](https://numcodecs.readthedocs.io/en/stable/packbits.html) filter for boolean fields.

For some 1000 genomes data I saw `call_genotype_phased` go from 2.8 MB to 212 K, after enabling it.

jeromekelleher:
Nice! I guess we should provide fairly open access to the Zarr compressors and filters too - it's good to be able to fiddle with these for specific applications.

tomwhite:
Fixed in #644

PR #644, opened 2021-07-28
Add compressor option to VCF functions.

### LINKED ISSUE #641: Add compressor option to vcf_to_zarr
Like the one in [bgen_to_zarr](https://pystatgen.github.io/sgkit/latest/generated/sgkit.io.bgen.bgen_to_zarr.html).

tomwhite:
Fixes #641 and #642

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -1,5 +1,6 @@
 import functools
 import itertools
+import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path
@@ -20,7 +21,7 @@ import fsspec
 import numpy as np
 import xarray as xr
 from cyvcf2 import VCF, Variant
-from numcodecs import Blosc, PackBits
+from numcodecs import PackBits
 
 from sgkit import variables
 from sgkit.io.utils import zarrs_to_dataset
@@ -39,6 +40,14 @@ from sgkit.utils import max_str_len
 DEFAULT_MAX_ALT_ALLELES = (
     3  # equivalent to DEFAULT_ALT_NUMBER in vcf_read.py in scikit_allel
 )
+
+try:
+    from numcodecs import Blosc
+
+    DEFAULT_COMPRESSOR = Blosc(cname="zstd", clevel=7, shuffle=Blosc.AUTOSHUFFLE)
+except ImportError:  # pragma: no cover
+    warnings.warn("Cannot import Blosc, falling back to no compression", RuntimeWarning)
+    DEFAULT_COMPRESSOR = None
 
 
 @contextmanager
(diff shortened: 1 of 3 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -20,6 +20,7 @@
 import numpy as np
 import xarray as xr
 from cyvcf2 import VCF, Variant
+from numcodecs import Blosc, PackBits
```
jeromekelleher, line 23: This can be problematic from experience - I'm not sure about Blosc, but Zstd can fail to import from numcodecs if there's a compile problem. Unfortunately, early on numcodecs allowed installations to proceed even if some bits failed to compile, and they're stuck with that behaviour now :(

https://github.com/tskit-dev/tsinfer/issues/161

sgkit/io/vcf/vcf_reader.py
```diff
@@ -331,6 +332,10 @@ def vcf_to_zarr_sequential(
     region: Optional[str] = None,
     chunk_length: int = 10_000,
     chunk_width: int = 1_000,
+    compressor: Optional[Any] = Blosc(
```
jeromekelleher, line 335: Better to have a DEFAULT_COMPRESSOR module variable, so that we can at least deal with missing compressors? Maybe it should be None in the signature also, so that we can have the freedom to choose different compressors for different fields, if we want to? (We probably will, I would imagine)

sgkit/io/vcf/vcf_reader.py
```diff
@@ -458,15 +463,24 @@ def get_chunk_size(dim: Hashable, size: int) -> int:
...
+                    default_encoding[var] = dict(
+                        chunks=var_chunks, compressor=compressor
+                    )
+                    if ds[var].dtype.kind == "b":
+                        default_encoding[var]["filters"] = [PackBits()]
+
+                # values from function args (encoding) take precedence over default_encoding
+                encoding = encoding or {}
+                merged_encoding = {**default_encoding, **encoding}
+                print(merged_encoding)
```
jeromekelleher, line 481: print

jeromekelleher (commented):
LGTM, minor API suggestions.

sgkit/io/vcf/vcf_reader.py
tomwhite: Ah, I wasn't aware of that. I've put the import for Blosc in a try/except block now, which issues a warning and falls back to no compression if the import fails.

sgkit/io/vcf/vcf_reader.py
tomwhite: Changed to have a `DEFAULT_COMPRESSOR` variable, but I haven't made None the default, since it is used to signify no compression. Having different compressors for different fields can be achieved by setting `encoding` - users can do this now, and the library could change the defaults in future if we wanted.

sgkit/io/vcf/vcf_reader.py
tomwhite: Fixed

sgkit/io/vcf/vcf_reader.py
jeromekelleher: Sounds good. I guess we can check if the default was specified by saying ``compressor is DEFAULT_COMPRESSOR``, right, since this is a singleton?

jeromekelleher (approved):

Issue #635, opened 2021-07-22
Add IBS based kinship estimator of Weir and Goudet (2017)


timothymillar:
I have been using the kinship estimator developed by [Weir and Goudet (2017)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5560808/) ($\hat{\beta}_{jj'}^i$ in table 3) and have found it very accurate and convenient. This method is based on IBS probabilities and is particularly accurate when samples are closely related, e.g. samples from a single pedigree (see [Goudet, Kay and Weir 2018](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6220858/)). I already have a rough implementation in sgkit.

eric-czech:
Nice!  Do you often need IBD estimates as well?  A method to compute either with some invariance to population structure would be awesome.  We added [pc_relate](https://pubmed.ncbi.nlm.nih.gov/26748516/) a while back, which could produce IBD sharing probabilities, but it requires PCs for population correction and those PCs ideally require sample independence (so we have a chicken and egg scenario).  

KING comes up most frequently as a solution for this problem -- do you know off the top of your head if this method is comparable?

jeromekelleher:
This sounds like a great idea to me @timothymillar!

timothymillar:
@eric-czech currently I'm using the W & G method to identify samples whose relationships don't match those expected based on pedigree (due to miss-sampling, rogue pollen etc.). This doesn't require a real IBD value, just something proportional to it. The W & G method calculates kinship relative to to the current (i.e. sampled) population which they refer to as "within population kinship" (analogous to F_IS). This means that the values identified can be negative (hence not actual IBD) with the non-diagonal (non-self) values summing to zero.

W & G also provide a method (eq. 11) to correct the above estimates based on the set of least related individuals. This can produce positive estimates that could be interpreted as IBD values. However, they point out that the resulting values are still relative to the sampled population (the least related samples could still be closely related/inbred). So the applicability of the corrected estimator probably depends on the population and hypothesis in question.

Thanks for pointing out the KING method, I'm not familiar with it but I'll look into it. W & G describe the KING-robust method as "estimating ibd between individuals relative to that within individuals".

eric-czech:
Thanks @timothymillar, good to know and still sounds promising (I think).  I only know the KING methods well enough to know they're not particularly easy to implement and are very common as early stage QC filters in human genetics pipelines.  

[...]

PR #647, opened 2021-08-04
Weir-Goudet kinship estimator

timothymillar:
Implements #635 including the equation (`Weir_Goudet_beta_corrected`) to correct estimates.

Some thoughts:

- The `infer_call_allele_frequencies` method is quite trivial but it seems likely that the `call_allele_frequency` array will be reused elsewhere. 
- I've started a new sub-module `ibs.py` due to the loose fit with existing sub-modules and the possibility of implementing more functionality from Weir and Goudet 2017. I'm happy to merge into an existing sub-module if that is preferred.
- This tests `Weir_Goudet_beta` against values calculated with function `beta.dosage` (by Goudet) of the R package `hierfstat`. `beta.dosage` only supports bi-allelic variants where the implementation in this PR also supports multi-allelic.
- This tests `Weir_Goudet_beta_corrected` against pedigree calculated kinship, this has lower than desired accuracy so I may need create a test case with more variants or verify another way.
- `Weir_Goudet_beta` must be run prior to `Weir_Goudet_beta_corrected` because selection of the `beta0` parameter may be dependent on population structure etc. It may be possible to have a smart default in future but that would require some care.
- Still need to add methods to docs.


sgkit/stats/aggregation.py
```diff
@@ -327,6 +327,77 @@ def count_cohort_alleles(
     return conditional_merge_datasets(ds, new_ds, merge)
 
 
+def infer_call_allele_frequencies(
```
jeromekelleher, line 330: ``infer`` feels like the wrong word here, but I see you're following the established practices so I'm not sure.

@tomwhite, any thoughts?

This is a nice well-defined operation that we could pull out in a separate PR to make this one a bit lighter - what do you think @timothymillar?

sgkit/stats/ibs.py
```diff
@@ -0,0 +1,260 @@
...
+@guvectorize(  # type: ignore
+    [
+        "void(float32[:,:], float32[:,:], float32[:,:])",
+        "void(float64[:,:], float64[:,:], float64[:,:])",
+    ],
+    "(x,a),(y,a)->(x,y)",
+    nopython=True,
+    cache=True,
+)
+def identity_by_state(
```
jeromekelleher, line 26: Not something we need to follow up on here, but I bet this would work well on the GPU backend.

sgkit/stats/ibs.py
```diff
@@ -0,0 +1,260 @@
...
+    for i in range(x):
+        for j in range(y):
+            ibs = 0.0
+            for a in range(n):
+                ibs += af_x[i, a] * af_y[j, a]
+            out[i, j] = ibs
+    return
+
+
+def pairwise_ibs(
```
jeromekelleher, line 68: Would this function fit into the existing frameworks for pairwise distances? It's sort-of the same thing, isn't it?

sgkit/stats/ibs.py
```diff
@@ -0,0 +1,260 @@
...
+    )
+    return conditional_merge_datasets(ds, new_ds, merge)
+
+
+def Weir_Goudet_beta_corrected(
+    ds: Dataset,
+    *,
+    beta0: float,
+    stat_Weir_Goudet_beta: Hashable = variables.stat_Weir_Goudet_beta,
+    merge: bool = True,
```
jeromekelleher, line 202: Seems unnecessary to have a separate function for this - we could simply add the ``beta0`` parameter to ``Weir_Goudet_beta`` - if it's None, we do the uncorrected version and otherwise we do this.

jeromekelleher (commented):
This looks awesome @timothymillar. I've made some comments above - some of them are probably ignorant because I haven't thought hard enough about what pairwise IBS is, but good to get your thoughts on how this corresponds to the general pairwise distance problem.

Also might be worth considering pulling out the allele frequency stuff into a separate PR as an easy, quick win.

sgkit/stats/ibs.py
timothymillar: That was my first thought as well. The issue is that one usually needs to see the values of beta calculated from the data in order to pick a sensible value for `beta0`, which would require running the function twice. It's actually tempting to remove the correction function completely and just describe the process in the docstring of `Weir_Goudet_beta` with an example (it's quite simple).

sgkit/stats/aggregation.py
timothymillar: Agreed (I think I started the use of infer...). I'm not sure how people feel about having a method `call_allele_frequencies` which returns variable `call_allele_frequency`? There seems to be a pattern of plural for the method and singular for the variable.

There is an unexposed method `allele_frequency` which calculates `variant_allele_frequency`.  We could potentially rename this method to `variant_allele_frequencies` to match.

Happy to start a separate PR.

sgkit/stats/ibs.py
timothymillar: Good question. `pairwise_ibs` is currently calculating an "distance" matrix per variant (variants, samples, alleles) -> (variants, samples, samples). `pairwise_distance`  seems to only handle 2d matrices (M, N) -> (M, M) and doesn't vectorize along the additional axis (unless I'm missing something?) . 

`pairwise_ibs` should ideally be (variants, samples, alleles) -> (samples, samples), however I wanted to investigate a windowed version of this which is why I left the reduction until later.

I think it would be possible to implement `pairwise_ibs` as (variants, samples) -> (samples, samples) using the `pairwise_distance` framework however this would only handle biallelic dosages and I would like to handle the multi-allelic case.

timothymillar:
Thanks for reviewing @jeromekelleher, I think the next steps should be:
- separate PR for `call_allele_frequencies` (or alternate name)
- remove `Weir_Goudet_beta_corrected` for now, this can be a separate PR if necessary 
- investigate using the `pairwise_distance` framework for `pairwise_ibs`

sgkit/stats/ibs.py
jeromekelleher: Ah, thanks for clearing up @timothymillar. I think what you have currently is good, there's no point in trying to shoehorn it into the pairwise distance stuff, I just wanted to make sure we weren't duplicating code or ideas.

sgkit/stats/ibs.py
jeromekelleher: That does seem like a sensible approach. I don't think 
```python
x = func()
y = func(starting_point=x)
```
is any worse than 
```python
x = func()
y = func_with_starting_point(x)
```
though? Anyway, pulling the correction out of this PR seems like a good approach for the moment.

sgkit/stats/aggregation.py
tomwhite: No strong thoughts on this one. I'd be fine with functions called `call_allele_frequencies` and `variant_allele_frequencies`.

jeromekelleher:
@timothymillar - just to let you know that a bunch of folks are on vacation at the moment, so things are moving slowly. But we'll definitely get to reviewing and merging this great stuff soon :smile:

timothymillar:
No worries @jeromekelleher, there's no urgency for me.

timothymillar:
Still to add:
- more tests (multi-allelic case)
- docstring examples

I also wanted to consider windowing though I'm not sure that it would be very intuitive because the default should be to treat the entire genome as a single window.

sgkit/tests/test_ibs.py
```diff
@@ -0,0 +1,153 @@
...
+        [1, ((100,) * 10, (4, 4, 4, 4), (2,))],
+        [1, ((270, 330, 400), (7, 5, 4), (1, 1))],
+        [2, ((1000,), (3,), (2,))],
+        [2, ((100,) * 10, (3,), (2,))],
+        [2, ((100,) * 10, (2, 1), (1, 1))],
+    ],
+)
+def test_Weir_Goudet_beta__hierfstat(sim, chunks):
+    # load reference data calculated with function beta.dosage
+    # from hierfstat R library
```
tomwhite, line 129: It would be good to add a note somewhere (here or next to the data file) saying how the reference data was generated.

sgkit/stats/ibs.py
```diff
@@ -0,0 +1,194 @@
...
+    return
+
+
+def pairwise_ibs(
+    ds: Dataset,
+    *,
+    call_allele_frequency: Hashable = variables.call_allele_frequency,
+    merge: bool = True,
+) -> Dataset:
+    """Compute per variant pairwise IBS probabilities between samples.
```
tomwhite, line 74: Nit: I know IBS is a commonly-understood term, but it may be worth spelling it out here, i.e. `IBS (identity by state)`

sgkit/variables.py
```diff
@@ -386,6 +386,15 @@ def _check_field(
     )
 )
 
+ibs_probabilities, ibs_probabilities_spec = SgkitVariables.register_variable(
```
tomwhite, line 389: Should this be `ibs_probability`? Variables in sgkit are singular, see `call_genotype_probability` for example.

tomwhite (commented):
This looks great from a code structure point of view. I've made just a few small suggestions.

sgkit/variables.py
timothymillar: Good point, I think `variant_ibs_probability` may be best as there is an IBS matrix per variant.

timothymillar:
Thanks for the comments @tomwhite

timothymillar:
Hi @tomwhite this may require a re-review.
I have made some changes to the calculation of IBS probabilities. It now computes a single IBS matrix rather than a matrix per variant. This is *much* more memory efficient with large numbers of samples and variants. I have also added some multiallelic test cases and an example of correcting beta to improve approximation of kinship.

timothymillar:
The one thing I would still like to consider is windowed versions of these functions (IBS and beta). I don't think these would fit with well with the same API used in `popgen.py` because the popgen methods default to computing statistics per variant if no windows are present. Beta should default to computing a single matrix for the whole genome. Would the following be OK?
- `Weir_Goudet_beta(ds) -> (samples_0, samples_1)`
- `Weir_Goudet_beta(ds, windowed=True) -> (windows/variants, samples_0, samples_1)`

Or perhaps:
- `Weir_Goudet_beta(ds) -> (samples_0, samples_1)`
- `Weir_Goudet_beta(windowed_ds) -> (windows, samples_0, samples_1)`

Either of these options would also require a windowed version of their associated variables unless the defaults were shape `(1, samples_0, samples_1)` which I think may be confusing.

tomwhite:
Is there ever a need to compute per-variant statistics in this case? If so, then it might be difficult to express that with either of the proposed APIs, since producing a window per variant is not something we've done (or optimized for).

Regarding the default of computing per-variant stats for the popgen functions - I think generally users want windows, but there is no default way to define windows, so starting with per-variant is reasonable. Computing stats over the whole genome (i.e. one big window) can be very inefficient, since most of the implementations assume each window is processed by one task.

The use of `concatenate=True` in the blockwise call can be problematic for memory usage, so we've tried hard to avoid that pattern. See https://github.com/dask/dask/issues/6874 for an example. I wonder if we can avoid it in this case.

timothymillar:
Thanks @tomwhite, I'm wondering if I'm trying to do too much in a single function. Calculating beta for the entire genome is useful on its own for exploring population structure. However, W&G also define a number of F-stat estimators based on beta. These F-stats make more sense using the same windowed API as used in `popgen.py`. An implementation of these F-stat estimators would automatically compute windowed beta and couldn't make use of the "whole genome beta" which suggests to me that these should at least be separate variables in the schema. But I'm unsure whether both variables should be computed from a single function with a flag (`windowed=True/False`) or if it's perhaps better to have separate functions?

I don't think there is a need to compute the per variant stats, just as a default if no windows are defined. Thanks for pointing out the issue with `concatenate=True`, I'll look for a workaround.

tomwhite:
> Calculating beta for the entire genome is useful on its own for exploring population structure.

Is that still true for very large genomes, or does it become less useful then? If so, then it may not be worth putting a lot of effort into working on a way of making the computation scale for entire genome beta, particularly if it's generally done on windowed data in practice.

> But I'm unsure whether both variables should be computed from a single function with a flag (`windowed=True/False`) or if it's perhaps better to have separate functions?

Different variables probably suggest different functions, given some of the discussion in #651.

timothymillar:
Ah... so it appears that the IBS gufunction can simply be replaced with `da.einsum` which would presumably take care of the `concatenate=True` issue as well. Though I'm currently having some issues related to chunking in the samples dimension.

> Is that still true for very large genomes? 

Yes beta is a (scaled) kinship estimator, it has a similar use cases to pc-relate. 

> suggest different functions

In that case I'll leave windowing out of this PR to keep it smaller. If I can figure out the chunking issue I'll squash the history and this should be good to go.

tomwhite:
> Though I'm currently having some issues related to chunking in the samples dimension.

FWIW the popgen methods don't currently support chunking in the samples dimension (e.g. see warning in https://pystatgen.github.io/sgkit/latest/generated/sgkit.diversity.html). So I think it would be OK to not support that case here if you don't currently have a need for it. 

> In that case I'll leave windowing out of this PR to keep it smaller. If I can figure out the chunking issue I'll squash the history and this should be good to go.

+1

timothymillar:
I think this ready to go now.

I've opened [dask/issues/8131](https://github.com/dask/dask/issues/8131) about the chunking issue and I've just removed the problem tests for now.

Performance with einsum is good, I also used it in place of dot product for the IBS denominator which avoids a transpose. I really should have thought of einsum earlier...

Thank again for your help @tomwhite  and @jeromekelleher!

tomwhite (approved):
This looks great @timothymillar.

Do you want to add warnings about not supporting chunking in the samples dimension (like https://pystatgen.github.io/sgkit/latest/generated/sgkit.diversity.html)?

timothymillar:
I'm happy to add the warning for now if that's your preference? I was going to see if it gets fixed upstream in the short term as it should work once that dask issue is resolved.

tomwhite:
It's always difficult to know when an upstream issue will be fixed (and this one doesn't seem to have a quick fix, judging by some of the comments). So adding a warning/ref to the upstream issue seems prudent to me in this case.

timothymillar:
Good call. I had another look into that dask issue and under some circumstances it can produce a correctly shaped array with the wrong values, so I'm going to raise NotImplementedError if the samples dimension is chunked.

tomwhite:
New change looks good to me. I think this is ready to merge now. @jeromekelleher do you agree?

jeromekelleher (approved):
LGTM!

PR #658, opened 2021-08-31
Implement region queries

tomwhite:
This makes it possible to do region queries (#25) and sample subsets on datasets.

I'm submitting it now to get feedback on whether it's the kind of thing we'd like. A slightly different API would be to expose functions like `_region_to_index` and `_sample_ids_to_index` and then the user would call Xarray's `isel` directly with the indexes returned from those functions.

The API only supports a single region, but we might want to add multiple regions. Also, the implementation materializes some variant and sample data in memory, but does _not_ use Xarray indexes at all - that might be something we want to investigate too (e.g. from a performance point of view).


tomwhite:
> A slightly different API would be to expose functions like _region_to_index

I'm wondering if this

```python
ds.isel(dict(variants=regions_to_index(ds, "1:2-3")))
```

is better than

```python
select(ds, regions="1:2-3")
```

It is a bit longer, but ties into the Xarray way of doing things. It means you could do other indexing operations (that `select` in this PR doesn't support) so would be more uniform. E.g. if you just wanted the first 10 samples, you could say:

```python
ds.isel(dict(samples=slice(10)))
```

and combining the two indexing operations would be more natural:

```python
ds.isel(dict(variants=regions_to_index(ds, "1:2-3"), samples=slice(10)))
```

tomwhite:
I've updated this PR with the approach described above, so there is now just a single `regions_to_indexer` function (samples can be done separately).

As @jeromekelleher pointed out in [last week's meeting](https://github.com/pystatgen/sgkit/discussions/660) the code does not support region queries that overlap indels. I've added a note about that limitation as a warning in the docs. Fixing that will require more work, using some kind of interval tree (e.g. https://brentp.github.io/interlap/).

jeromekelleher:
> ds.isel(dict(variants=regions_to_index(ds, "1:2-3")))

I think this version is much better @tomwhite. 

I totally agree that we should support parsing samtools-ish query strings, but I don't think they should be the primary interface that we recommend to users. A native Python function would be much better, which takes ``contig``, ``start`` and ``end`` (or whatever's more appopriate) as arguments. However, we're then into tricky semantic territory: the query strings are ``end`` inclusive, whereas it's much more natural in Python to use ``end`` exclusive.

What if we have a function called ``sg.pslice(ds, contig, start, stop)``, for "position slice", then we'd have

```python
ds.isel(dict(variants=sg.pslice(ds, "1", 2, 4))
```

I would argue for making a Pythonic interface first, and then supporting familiar genomics workflows on top of this.

tomwhite:
Thanks @jeromekelleher. Good point about a Pythonic interface - I've updated this PR with a possible implementation.

I think the API should support multiple regions, so I changed the function to accept a sequence of region tuples:

```python
regions_to_indexer(ds, ("0",), ("1", 2, 4))
```

I'm not sure about the `pslice` name. With a single region, it echos Python's `slice` function, but for multiple regions it doesn't so much, and the type returned is not actually a slice object, but an Xarray indexer, which could be one of several types.

jeromekelleher:
I agree the API should support multiple regions, but I think we'll often be working with cases where we only want a single slice, and this general notation could get tedious. What about
```
def slice_to_indexer(df, contig=None, start=None, stop=None):
     # defaults:
     # contig = None - if only one contig, operate on this contig, else ValueError
     # start=None - start of contig up to stop
     # stop= start up to end of contig
      ...
```

(Or maybe ``pslice_to_indexer`` to emphasise that this is on genomic positions, not indexes?)

I'm wary of calling this a "region" because of the implication that we're inheriting the samtools region semantics. Ideally, I think a function ``sg.parse_region`` that returns the equivalent slices would emphasise the difference in semantics.

I'm also wary of depending on tuples to unpack parameters in your version, but I see the motivation for keeping things concise.

Thoughts?

alimanfoo:
Just a random comment to say very glad to see work in this direction. Also pandas seems very comfortable with indexers being stop inclusive.

tomwhite:
Thanks @alimanfoo, that's good to know! We'll need to decide which stop convention to use here.

@jeromekelleher are you suggesting having two functions, `slice_to_indexer` for the single slice case, and `slices_to_indexer` for the multiple slice case (with the latter having the more cumbersome tuple syntax)?

I"m fine calling this `slice_to_indexer` or `pslice_to_indexer` and then adding a `region_to_indexer` equivalent later for the overlapping indel case (equivalent to samtools regions).

jeromekelleher:
> @jeromekelleher are you suggesting having two functions, slice_to_indexer for the single slice case, and slices_to_indexer for the multiple slice case (with the latter having the more cumbersome tuple syntax)?

Yes, I think so, but it would be good to get some wider input here.

I think the overlapping indel thing is orthogonal though: regardless of whether we specify the query as a samtools region string or as a slice of some sort, we'd expect to return the same set of variants (modulo the slightly different semantics).

timothymillar:
It looks like the specification of individual regions is converging on something similar to [pysam](https://pysam.readthedocs.io/en/latest/api.html#pysam.VariantFile.fetch) which I like. They use 0-based half open indices for python tuples and 1-based indexing for a region string [[1]](https://pysam.readthedocs.io/en/latest/api.html#pysam.HTSFile.parse_region).

I  don't mind weather there are 1 or 2 functions for single/multi-region queries but I guess the most common use-case would be a single region so it might be best to have that as simple as possible.

I like the use of xarrays `ds.sel(dict(..))` API. It also makes it trivial to write your own fancy select function if needed.

> using some kind of interval tree

I have minimal knowledge on the best structures for these queries but this [project](https://github.com/biocore-ntnu/ncls) and the associated [paper](https://academic.oup.com/bioinformatics/article/23/11/1386/199545) claim to have great performance using nested containment lists.

tomwhite:
Another approach (inspired by pyranges), would be to allow the `contig`, `start`, `end` parameters to be either scalars (for the single slice case) or lists (for the multiple slice case), thereby avoiding having two functions. For example:

```python
slice_to_indexer(ds, "1", 2, 4) # single slice: positions 2-4 on contig 1
slice_to_indexer(ds, ("0", "1"), (None, 2), (None, 4)) # all of contig 0, and positions 2-4 on contig 1
```

This could work with pyranges as follows (while not requiring it as a dependency of sgkit):

```python
import pyranges as pr
r = pr.from_dict({"Chromosome": ["0", "1"], "Start": [0, 2], "End": [5, 4]})
slices_to_indexer(ds, r.Chromosome.tolist(), r.Start.tolist(), r.End.tolist())
```

In practice, pyranges could be used to read from BED/GFF3, etc. There are also inevitable genomic coordinate system issues (#434) to consider when mixing libraries like this.

tomwhite:
I'm going to close this for the time being - we can reopen and revisit when it's needed.

Issue #643, opened 2021-07-26
Use variable-length strings for storing alleles in Zarr


tomwhite:
Currently we use fixed-length strings for storing alleles, but this is inefficient since the length is the size of the longest allele in the whole dataset.

For example, in some 1000 genomes data (chr22) I noticed that the longest allele is 414 base pairs, which means the data type is "S414" - unnecessarily large for the vast majority of variants. This meant that the `variant_allele` data took up 15 MB (compressed), rather than something like one tenth of that if a variable-length encoding were used (like scikit-allel does).

It would be worth investigating how we could use Zarr's [variable-length strings](https://zarr.readthedocs.io/en/stable/tutorial.html#string-arrays) (the number of alt alleles would remain fixed though, in contrast to #634). In particular, there may be some work to get this representation to work nicely with xarray.

timothymillar:
@tomwhite the freebayes variant caller is a good example for producing a range of allele lengths. The VCFs avialable [here](https://github.com/pfrnz/Example-Tetraploid-Potato-VCF-PRJNA414303) (see releases) were called with freebayes and have allele lengths of up to ~ 60bp. Note that these were called with `--haplotype-length -1` in an attempt to minimize variant length.

tomwhite:
> It would be worth investigating how we could use Zarr's [variable-length strings](https://zarr.readthedocs.io/en/stable/tutorial.html#string-arrays)

In the case of writing Zarr sequentially from VCF, we already do this (it can be enforced by passing `target_part_size=None` to `vcf_to_zarr`). The parallel case concatenates a collection of Zarr files, and as a byproduct of that the variable-length strings are converted to fixed-length strings. We should change that.

tomwhite:
I'm not sure how to fix this, so I opened https://github.com/pydata/xarray/discussions/5769 with a minimal example.

Issue #365, opened 2020-11-02
Add concat algorithm parameter to `vcf_to_zarr`


tomwhite:
This would allow (regular) VCF conversion to use the optimized path introduced in #324 (as the default), in addition to the internal Xarray/Dask concat and rechunk implementations that suffer from memory issues.

PR #665, opened 2021-09-13
Add concat algorithm parameter to vcf_to_zarr #365

tomwhite:
This is an implementation of #365, which will hopefully fix #661. It switches the concat algorithm to use an optimized memory-efficient version (using the same code as #324). Users don't have to change anything to get reduced memory use.

I've tested it on a smaller dataset (1kg chr22), and it reduces memory use significantly, but it still needs more testing. 

As a side effect of the way this is implemented, it also makes addressesing #643 straightforward (and in fact more natural than the special cases needed to set the fixed string lengths). Xarray is effectively bypassed, so it is easy to use Zarr's variable-length strings directly.

I wonder if we should remove the `xarray_internal` option entirely for that reason. (It's not the default, the optimized concat and rechunk algorithm is, but it's another code path to maintain.)

There are still potentially some race conditions lurking - I think I saw the problem in #486 again while testing locally, however it's very hard to reproduce.


jeromekelleher:
Haven't thought hard about this @tomwhite, but I'd be happy to get rid of the ``xarray_internal`` code path if there isn't a compelling performance reason for keeping it. What's to be gained from keeping it?

tomwhite:
> I'd be happy to get rid of the `xarray_internal` code path if there isn't a compelling performance reason for keeping it. What's to be gained from keeping it?

There's no performance reason for keeping it (it's slower and causes memory errors). If/when it's fixed upstream it would be straightforward to reintroduce it, so I think we can remove it.

tomwhite:
I think this can be merged as a first step, with the work to remove the `xarray_internal` code path done in a follow on PR.

I spent a bit of time running this on a 1GB VCF file (1kg chr2) and it worked fine.

jeromekelleher (approved):
LGTM! Merge away

PR #667, opened 2021-09-14
Use of int8 for variant_contig results in integer overflow with fragmented reference genomes #584

### LINKED ISSUE #584: Use of int8 for variant_contig results in integer overflow with fragmented reference genomes
Many (non human) reference genomes contain 1000s of contigs that have not been assembled into full chromosomes.
Currently the `variant_contig` array  is hard coded as int8 ([line](https://github.com/pystatgen/sgkit/blob/main/sgkit/io/vcf/vcf_reader.py#L267)) which results in integer overflow making it impossible to join variants to their contig.

tomwhite:
Fixes #584

I've included a test that uses a VCF with >128 contigs to check that it correctly picks an appropriate dtype. I haven't added a way to override the dtype, as I'm not sure it's really needed. We could add it, but it would clutter the parameter list a bit.


sgkit/utils.py
```diff
@@ -336,3 +336,22 @@ def hash_array(x: ArrayLike, out: ArrayLike) -> None:  # pragma: no cover
...
+
+    Returns
+    -------
+    A NumPy signed integer dtype suitable for storing the given value, or None if the value exceeds
+    the bounds of the largest dtype (``np.int64``).
+    """
+    for dtype in (np.int8, np.int16, np.int32, np.int64):
+        if np.iinfo(dtype).min <= value <= np.iinfo(dtype).max:
+            return dtype
+    return None
```
jeromekelleher, line 357: This return None case basically can't happen, so why not raise an OverflowError here rather than requiring client code check the return value (C style)?

jeromekelleher (approved):
LGTM, but I think we just raise an overflow error. If we get an input VCF with anything remotely like 2**63 rows, a whole bunch of other things will break before getting to this point!

sgkit/utils.py
tomwhite: Yes, that's much better. I've implemented this in https://github.com/pystatgen/sgkit/pull/667/commits/f63600f2a15d40663af05d0b934061229c50b3e1. Should be ready to go in now.

PR #744, opened 2021-10-19
How do I ...?

### LINKED ISSUE #474: How do I ...?
Xarray has a nice [page](http://xarray.pydata.org/en/stable/howdoi.html) with "How do I ...?" questions. It would be nice to have something with common task-based questions for sgkit.

Some possible questions:

How do I ...
* create a test dataset?
* look at the dataset summary?
* find the definition for a variable?
* subset to a set of variables?
* look at the values for the variables in a dataset?
* look at the genotypes?
* look at a particular field for a particular locus and sample?
* subset to a genomic range?
* get the list of samples?
* subset to a set of samples?
* filter variants by a given set of criteria?
* create a new variable based on others?
* get summary stats?
* find new variables added by a method?
* save results to a Zarr file?

tomwhite:
Fixes #474

I wrote two versions, one with complete examples, one more minimal (like Xarray's http://xarray.pydata.org/en/latest/howdoi.html). See attachments here:

[howdoi.pdf](https://github.com/pystatgen/sgkit/files/7372164/howdoi.pdf)

[howdoi_min.pdf](https://github.com/pystatgen/sgkit/files/7372165/howdoi_min.pdf)

Thoughts on which one to go with?


tomwhite:
Thanks @eric-czech. Good point about the caveats - perhaps I should add these now.

> I'd also be in favor of combining them so that the questions in howdoi_min.pdf are links anchored to sections like those in howdoi.pdf.

Possibly overkill? It might be a bit confusing to users to present it like that.

Issue #740, opened 2021-10-18
VCF Zarr specification


tomwhite:
Write a spec for how we represent VCF in Zarr. This is likely to be slightly different from the current sgkit implementation (field names should match the VCF field names, for example), but it should be easy to do a lightweight conversion to sgkit conventions via a utility method.

jeromekelleher:
Totally agree.

I'm not sure there's much value in having a second zarr format with difference column names though - wouldn't it be simpler to just say "this column corresponds to VCF field X"? Either that or we just use the VCF names directly in the sgkit format, perhaps making the current names available as synonyms in sgkit?

tomwhite:
I've updated the draft spec at https://github.com/tomwhite/sgkit/blob/vcfzarr-spec/vcf_zarr_spec.md following discussion on #755 to simplify the types and to remove the 32-bit limitation.

tomwhite:
Another update of the draft spec, https://github.com/tomwhite/sgkit/blob/vcfzarr-spec/vcf_zarr_spec.md, which uses the column names we already have in sgkit (to reflect earlier comments in this issue).

tomwhite:
I've created a repo for the spec at https://github.com/pystatgen/vcf-zarr-spec (as we discussed in the last meeting, #761), and a PR for the first version of the spec at https://github.com/pystatgen/vcf-zarr-spec/pull/1.

tomwhite:
Fixed in https://github.com/pystatgen/vcf-zarr-spec/pull/1

PR #755, opened 2021-10-28
Use missing and fill values from BCF

tomwhite:
Related to #740

### DIFF
--- a/sgkit/io/utils.py
+++ b/sgkit/io/utils.py
@@ -8,12 +8,7 @@ import zarr
 from ..typing import ArrayLike, DType
 from ..utils import encode_array, max_str_len
 
-# These values are based on BCF conventions
-INT32_MISSING, INT32_FILL = np.array(
-    # Equivalent to 0x80000000, 0x80000001
-    [np.iinfo(np.int32).min, np.iinfo(np.int32).min + 1],
-    dtype=np.int32,
-)
+INT32_MISSING, INT32_FILL = -1, -2
 
 FLOAT32_MISSING, FLOAT32_FILL = np.array([0x7F800001, 0x7F800002], dtype=np.int32).view(
     np.float32
--- a/sgkit/io/utils.py
+++ b/sgkit/io/utils.py
@@ -8,7 +8,7 @@ import zarr
 from ..typing import ArrayLike, DType
 from ..utils import encode_array, max_str_len
 
-INT32_MISSING, INT32_FILL = -1, -2
+INT_MISSING, INT_FILL = -1, -2
 
 FLOAT32_MISSING, FLOAT32_FILL = np.array([0x7F800001, 0x7F800002], dtype=np.int32).view(
     np.float32
--- a/sgkit/io/utils.py
+++ b/sgkit/io/utils.py
@@ -9,13 +9,19 @@ from ..typing import ArrayLike, DType
 from ..utils import encode_array, max_str_len
 
 # These values are based on BCF conventions
-INT32_MISSING, INT32_FILL = np.array([0x80000000, 0x80000001], dtype=np.int32)
+INT32_MISSING, INT32_FILL = np.array(
+    # Equivalent to 0x80000000, 0x80000001
+    [np.iinfo(np.int32).min, np.iinfo(np.int32).min + 1],
+    dtype=np.int32,
+)
+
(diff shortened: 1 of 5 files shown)


jeromekelleher:
I'm not sure I agree with this - aren't our existing fill values better? I don't see how it would help us interoperate with anyone to use the BCF values, since people very rarely interact directly with BCF.

tomwhite:
I think there are a couple of problems with the way we currently do things. First, we conflate the missing and fill values, so it's not possible to distinguish between say `DP=.` and the `DP` field not appearing for that variant. Second, the current fill values for integer and float types can appear in the data as actual values. For example, -1 is used to represent fill for integers, but that could appear in the data, so it's not possible to tell if it represents -1 (the value), missing, or fill.

These are admittedly edge cases that may not occur much in practice, but I think it would be a good idea to straighten them out. What I do think is missing here is a function to make this transparent to the user: `is_present` or something that does the low-level comparisons for all types so the user doesn't have to know about how they are encoded.

This PR will make it possible to say in the spec that _any_ VCF file that can be written as BCF can be written losslessly as VCF Zarr, which I think is a pretty strong statement to be able to make. I've been working on a branch that has an implementation and some tests to demonstrate this.

jeromekelleher:
I see - good points. I'll think a bit more about it, but I'm still worried about the dependency on 32 bit integer limits, and what happens if you want to do use, say, 16 bit values instead. Then, using a constant like -3 (or whatever) is a lot simpler and less fragile.

Also, there will be a time when genome sizes become 64 bit. There are plenty of chromosomes out there longer than a 32 bit int.

tomwhite:
> I see - good points. I'll think a bit more about it, but I'm still worried about the dependency on 32 bit integer limits, and what happens if you want to do use, say, 16 bit values instead. Then, using a constant like -3 (or whatever) is a lot simpler and less fragile.

So far, we've assumed that ints for general VCF fields are 32 bit, but we might want to allow users to say that a particular field should use 16 bit ints, as you say, since it is known to only use a small range of values.

I guess it boils down to which approach is more likely to cause data conflicts or confusion: a small magnitude, but fixed constant (-3), or a larger magnitude, but variable (depending on dtype) constant (`np.iinfo(dtype).min`)? I think the latter would be OK if users have a function to test for these special values, as the former means it's never safe to have these special values in the data.

> Also, there will be a time when genome sizes become 64 bit. There are plenty of chromosomes out there longer than a 32 bit int.

I agree that we need to cater for this. The position field is mandatory, so there are no issues with missing values. We currently hard code it to int32, but we should allow int64 to be used too (maybe the user has to specify this?).

jeromekelleher:
> I guess it boils down to which approach is more likely to cause data conflicts or confusion: a small magnitude, but fixed constant (-3), or a larger magnitude, but variable (depending on dtype) constant (np.iinfo(dtype).min)? I think the latter would be OK if users have a function to test for these special values, as the former means it's never safe to have these special values in the data.

Having gone down the road of using large values as sentinels before, I'd be strongly in favour of using small negative values. It's such a horrid mess, even when you do have functions to do the comparisons.

Think about it this way - suppose you were interfacing with this data file format using C. Which do you think would be simpler to work with, a fixed small constant or a type specific value?

I don't think we should go into this with the perspective that VCF/BCF is perfect and should be reproduced as closely as possible. Baking in fixed-size types to the data format is an unequivocal mistake in my book and not one we should reproduce. We don't have to have APIs to support arbitrary sized integers for every VCF field, but we should definitely allow for it to be done in the future.

tomwhite:
Thanks @jeromekelleher. Let's use small negative sentinels for integer types - I've updated this PR.

sgkit/io/utils.py
```diff
@@ -8,6 +8,17 @@
 from ..typing import ArrayLike, DType
 from ..utils import encode_array, max_str_len
 
+INT32_MISSING, INT32_FILL = -1, -2
```
jeromekelleher, line 11: The 32 seems redundant now - maybe just INT_MISSING?

jeromekelleher (approved):
LGTM, one suggestion

tomwhite:
Failures are due to mypy changes (see #759) and are not directly related to this issue, so I'm going to merge this.

PR #765, opened 2021-11-16
Document how to release

### LINKED ISSUE #691: Document how to release
@tomwhite wrote up some great instructions at https://github.com/pystatgen/sgkit/issues/608#issuecomment-891117795. We should probably put these instructions somewhere more permanent, though I'm not quite sure where. Perhaps as a discussion at https://github.com/orgs/pystatgen/teams/committers?

tomwhite:
Fixes #691

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -328,6 +328,98 @@ We use `mergify <https://mergify.io/>`_ to automate PR flow. A project
 to automatically merge a PR by labeling it with ``auto-merge``, and then when the PR gets
 at least one approval from a committer and a clean build it will get merged automatically.
 
+Release process
+---------------
+
+Releases are made by a project maintainer by following these steps.
+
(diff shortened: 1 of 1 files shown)


jeromekelleher (commented):
Looks good, but I'm not sure I see much value in making maintainers go through these steps manually. Couldn't we add these as an extra github action or something, so that making a release is just "use the UI on GitHub check that all the actions succeed?"

tomwhite:
@jeromekelleher you are right, we should be aiming to do all of the release steps without using a local machine. We already test the artifacts using the scripts I documented here (see https://github.com/pystatgen/sgkit/blob/main/.github/workflows/wheels.yml), so we could make the release via the GitHub UI as you suggest.

The only step that still needs to be done from a local machine is the documentation, but we should add that to the GitHub action too. I've opened #768, which we should do as a part of (or before) the next release.

I'm closing this PR.

Issue #734, opened 2021-10-12
Add function to get max sizes of fields from VCF


tomwhite:
From https://github.com/pystatgen/sgkit/issues/725#issuecomment-936338657

This would retrieve the maximum number of alt alleles, ploidy, and actual (maximum) size of Number=. fields, in a form suitable for passing to the `vcf_to_zarr` function.

tomwhite:
This exists as `zarr_array_sizes`, but it is not a part of the public API since it runs sequentially. Leaving this issue open to cover the parallel implementation.

PR #766, opened 2021-11-16
Fixes for VCF Zarr spec

tomwhite:
This PR fixes a few minor inconsistencies and omissions found while drafting a spec for the format (#740).

* Adds `variant_quality` and `variant_filter` fields.
* Stores samples as variable-length strings.
* Includes the original VCF header as a dataset attribute.
* Includes the version number of the spec as a dataset attribute.
* Adds a `zarr_array_sizes` function (#734), for testing only since it is only a sequential implementation.
* Adds a test for the spec that makes assertions that are based directly on statements from the spec.


sgkit/io/vcf/vcf_reader.py
```diff
@@ -902,3 +928,55 @@ def count_variants(path: PathType, region: Optional[str] = None) -> int:
...
+                        pass  # single value
+
+            for key, val in format_field_defs.items():
+                field_val = variant.format(key)
+                if field_val is not None:
+                    val["Number"] = max(val["Number"], field_val.shape[-1])
+
+            try:
+                ploidy = max(ploidy, variant.genotype.ploidy)
+            except Exception:
```
jeromekelleher, line 967: Does Cyvcf2 really throw a raw Exception? You'd expect it to be a TypeError or something a bit more specific. Probably worth putting a note here because I'd imagine linting tools will start warning about stuff like this at some point and it would be helpful for future devs to know the rationale

sgkit/io/vcf/vcf_reader.py
```diff
@@ -434,13 +443,23 @@ def vcf_to_zarr_sequential(
                     max_variant_allele_length, max(len(x) for x in alleles)
                 )
 
+                variant_quality[i] = (
+                    variant.QUAL if variant.QUAL is not None else FLOAT32_MISSING
+                )
+                variant_filter[i] = (
```
jeromekelleher, line 449: I'm not sure anyone really uses the FILTER field so I'm not overly bothered, but do we want to "unparse" this field like this? I guess it's something to do with the VlenUtf8 codec? Good to explain the rationale here and we'd definitely want to document that this is how the FILTER field is stored.

sgkit/tests/io/vcf/test_vcf_reader.py
```diff
@@ -1036,3 +1037,356 @@ def test_vcf_to_zarr__fields_errors(shared_datadir, tmp_path):
...
+def check_field(group, name, ndim, shape, dimension_names, dtype):
+    assert group[name].ndim == ndim
+    assert group[name].shape == shape
+    assert group[name].attrs["_ARRAY_DIMENSIONS"] == dimension_names
+    if dtype == str:
+        assert group[name].dtype == np.object_
+        assert VLenUTF8() in group[name].filters
+    else:
+        assert group[name].dtype == dtype
+
```
jeromekelleher, line 1080: Doesn't look like there's any tests for the multiple FILTER behaviour?

jeromekelleher (approved):
LGTM, minor nitpicks.

sgkit/io/vcf/vcf_reader.py
tomwhite: Fixed - it throws an AttributeException.

sgkit/io/vcf/vcf_reader.py
tomwhite: Done

sgkit/tests/io/vcf/test_vcf_reader.py
tomwhite: Done

sgkit/io/vcf/vcf_reader.py
tomwhite: Thinking about this more, perhaps it would be better to store this as an array of ints, which are offsets into a `filters` attribute - a bit like `variant_contig` being offsets into the `contigs` attribute.

jeromekelleher: That would be nicer all right - save a bit of space too I'd imagine.

tomwhite: I'm changed the implementation (and [draft spec](https://github.com/tomwhite/sgkit/blob/vcfzarr-spec/vcf_zarr_spec.md)) to encode filters in a boolean array of shape (variants, filters), where a true value in the `i`-th position means that the `i`-th filter is applied. A list of filter strings is also stored in the `filters` attribute.

How does this look?

jeromekelleher: Yeeesssss... yes, you have to do it this way because we have have a number of filters, and we would have to have a ragged array for filters to represent a number of them. So, this is better in the common not-very-many-filters case. Would be worse if there was say, hundreds of filters, but I don't think this happens (much - I bet somebody is doing it though).

What happens if (when) someone passes a filter value that's not declared in the header? It's probably not entirely valid VCF, but I bet htslib will happily accept it.

tomwhite: > Would be worse if there was say, hundreds of filters, but I don't think this happens (much - I bet somebody is doing it though).

These boolean arrays are very compressible though, so it shouldn't be a problem.

> What happens if (when) someone passes a filter value that's not declared in the header? It's probably not entirely valid VCF, but I bet htslib will happily accept it.

You are right: htslib does accept it. So we should probably handle it as we do a missing contig, by raising an error.

jeromekelleher: SGTM!

jeromekelleher:
I think this is sorted @tomwhite, good to merge?

PR #775, opened 2021-11-30
Rename 'call_genotype_non_allele' to 'call_genotype_fill'

### LINKED ISSUE #772: Rename `call_genotype_non_allele` to `call_genotype_fill`?
For consistency with a general way of specifying fill masks (https://github.com/pystatgen/vcf-zarr-spec/blob/main/vcf_zarr_spec.md#missing-and-fill-values) we might rename this variable. We could keep the "non-allele" terminology in the documentation in addition to describing it as a fill value.

Thoughts @timothymillar?

tomwhite:
Fixes #772

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -133,7 +133,7 @@ By convention, variable names are singular in sgkit. For example, ``genotype_cou
     variables.call_genotype_complete_mask_spec
     variables.call_genotype_spec
     variables.call_genotype_mask_spec
-    variables.call_genotype_non_allele_spec
+    variables.call_genotype_fill_spec
     variables.call_genotype_phased_spec
     variables.call_genotype_probability_spec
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
LGTM - I take it we didn't release call_genotype_non_allele then?

tomwhite:
Thanks @jeromekelleher. We did release call_genotype_non_allele, so this is a breaking change (albeit one that is allowed since we are still in the 0.x.y series - so we would release this as 0.5.0). However, I wonder how many people are using the variable ploidy stuff apart from @timothymillar.

jeromekelleher:
Sounds good @tomwhite - I guess we should just mark this as a breaking change in the release notes so that we don't forget about it? You never really know who's using what.

tomwhite:
I created a "breaking change" label for the issue so we can track this.

PR #776, opened 2021-12-07
Generic cohort_statistic function

### LINKED ISSUE #730: Generic functions for cohort means and sums
I've come across a couple of situations where I want to take mean of sample values for each cohort in a dataset. From what I can tell, there is not a simple way to achieve this using dask or xarray. 

We already have a numba jitted function to do this for a specific case ([`_cohort_observed_heterozygosity`](https://github.com/pystatgen/sgkit/blob/main/sgkit/stats/popgen.py#L951)) so it would make sense to generalize that to standard function in `utils.py` or similar. A higher level function could be made available which automatically calculates the cohort dimension size and runs `da.mapblocks` etc. We could also to the same for cohort sums.

timothymillar:
Fixes #730
This is inspired by the `window_statistic` function and assumes that it is reasonable to iterate over the number of cohorts in plain python.
Existing methods like `count_cohort_alleles` could also be simplified using `cohort_statistic` though there would probably be some reduction in performance.

### DIFF
--- a/sgkit/stats/popgen.py
+++ b/sgkit/stats/popgen.py
@@ -7,7 +7,7 @@ import numpy as np
 from numba import guvectorize
 from xarray import Dataset
 
-from sgkit.cohorts import _cohorts_to_array
+from sgkit.cohorts import _cohorts_to_array, cohort_statistic
 from sgkit.stats.utils import assert_array_shape
 from sgkit.typing import ArrayLike
 from sgkit.utils import (
@@ -939,52 +939,6 @@ def Garud_H(
     return conditional_merge_datasets(ds, new_ds, merge)
 
 
-@guvectorize(  # type: ignore
-    [
-        "void(float64[:], int32[:], uint8[:], float64[:])",
-        "void(float64[:], int64[:], uint8[:], float64[:])",
-    ],
-    "(n),(n),(c)->(c)",
-    nopython=True,
-    cache=True,
-)
-def _cohort_observed_heterozygosity(
-    hi: ArrayLike, cohorts: ArrayLike, _: ArrayLike, out: ArrayLike
-) -> None:  # pragma: no cover
-    """Generalized U-function for computing cohort observed heterozygosity.
-
-    Parameters
-    ----------
-    hi
-        Individual sample heterozygosity of shape (samples,).
-    cohorts
-        Cohort indexes for samples of shape (samples,).
-    _
-        Dummy variable of type `uint8` and shape (cohorts,) used to
-        define the number of cohorts.
-    out
-        Observed heterozygosity with shape (cohorts,) and values corresponding
-        to the mean individual heterozygosity of each cohort.
-    """
-    out[:] = 0
-    _[:] = 0
-    n_samples = len(hi)
-    n_cohorts = len(out)
-    for i in range(n_samples):
-        h = hi[i]
-        if not np.isnan(h):
-            c = cohorts[i]
-            if c >= 0:
-                out[c] += h
-                _[c] += 1
-    for j in range(n_cohorts):
-        n = _[j]
-        if n != 0:
-            out[j] /= n
-        else:
-            out[j] = np.nan
-
-
 def observed_heterozygosity(
     ds: Dataset,
     *,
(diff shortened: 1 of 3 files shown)


tomwhite:
LGTM. Maybe put `cohort_statistic` in `cohorts.py`?

jeromekelleher (approved):
LGTM also - nice that we can get rid of some guvectorised functions also. (I assume there's no major loss of performance here?)

timothymillar:
> Maybe put `cohort_statistic` in `cohorts.py`?

Yes ... that would be the obvious place for it :smile: 

> no major loss of performance here?

Actually, it's worse than I initially estimated. It's quite variable depending on chunking and the number of cohorts. With chunking that favors `cohort_statistic` (i.e. samples pre-chunked by cohort) the performance hit can be as low as 10% (in part due to the chunking slowing the numba version). Without chunking in the samples dimension, I'm usually seeing a 300-400% slowdown. The slowdown is primarily due to slicing the variables array into an array per cohort which is unavoidable if we want to use generic reduction functions.

I'm happy to take the alternate approach of implementing cohort-aware reductions in numba if we wish to maximize performance? I imagine that `cohort_sum`, `cohort_nansum`, `cohort_mean` and `cohort_nanmean` would cover most use-cases. `cohort_statistic` could be left to cover edge-cases.

jeromekelleher:
I think it's OK to have a perf reduction here @timothymillar, we should just make a note in the source that the current version could be improved with numba-fication. I don't have a strong opinion either way, I just wanted to raise the issue and keep a note of the perf implications of the change.

timothymillar:
Agreed @jeromekelleher we don't need to optimize this unless it becomes an issue. Though I think we need to keep an eye out for functions with similar reductions to avoid writing equivalent optimized methods in multiple places.

timothymillar:
Looks like CI on Linux is failing for an unrelated reason (`test_vcf_to_zarr__empty_region`), same issue in #777

sgkit/stats/popgen.py
```diff
@@ -1061,18 +1015,14 @@ def observed_heterozygosity(
...
-    ho = da.map_blocks(
-        _cohort_observed_heterozygosity,
-        hi,
-        sc,
-        n,
-        chunks=shape,
-        drop_axis=1,
-        new_axis=1,
-        dtype=np.float64,
+    # NOTE: Performance of cohort_statistic is substantially slower than a numba
```
jeromekelleher, line 1018: Chatting about this in the meeting today @timothymillar, we thought it'd be good if we could put a git hash in here to allow future travellers to easily find this fast implementation.

timothymillar: Good idea, done

timothymillar:
Thanks @jeromekelleher.
Just for clarity, this PR was intended to add a simple method to applying summary statistics to cohorts. I then updated `observed_heterozygosity` because it seemed redundant for it to have it's own 'cohort_nanmean' function, but I don't have strong feelings about it if others might be concerned about performance? I guess I'm not quite sure where the sgkit team sit with regard to performant vs simple code for small things like this.

For reference, I also played around with a decorator to simplify wrapping gufunctions for cohort reductions [here](https://gist.github.com/timothymillar/765adee5a4f6c14e7d218db4500caeb1).

tomwhite:
Thanks @timothymillar.

> I guess I'm not quite sure where the sgkit team sit with regard to performant vs simple code for small things like this.

We discussed it a bit yesterday (#781), and I think the general conclusion was to go with the simpler implementation until there's a need for better performance. Regarding `observed_heterozygosity` in particular, since you are probably the main user of this function we'll be led by whichever implementation you'd prefer.

timothymillar:
Thanks for the clarification @tomwhite. 

Well this has been a lesson in benchmarking carefully with dask, especially with implementations that produce different numbers of chunks! The plot below shows performance in relation to number of cohorts (the main factor). Cohorts were randomly assigned to each sample without chunking in the sample dimension. This is the worst case scenario for the new "numpy" implementation but probably quite realistic. Note that the new method actually performs *worse* with multi-threading (chunks = `(10000,) * 16, (100,))` on 8 cores).
bench2
I was a bit naive by benchmarking with only a few cohorts previously. I don't imagine many user would have more than a few cohorts, but it's probably best to not make that assumption.

I think we should **keep the existing version for now** and I'll revisit #730 another time. Sorry for taking up your time with this!

tomwhite:
Thanks for the investigation on this one @timothymillar!

PR #777, opened 2021-12-09
Add cohort_allele_frequencies method #729

### LINKED ISSUE #729: Cohort allele frequencies
Following #649 it would be good to also calculate cohort allele frequencies. This should simply be a case of dividing the cohort allele counts by their sum.

timothymillar:
Fixes #729


sgkit/stats/aggregation.py
```diff
@@ -437,6 +437,81 @@ def call_allele_frequencies(
...
+            [[0.75, 0.25],
+            [0.25, 0.75]]])
+    """
+    ds = define_variable_if_absent(
+        ds, variables.cohort_allele_count, cohort_allele_count, count_cohort_alleles
+    )
+    variables.validate(ds, {cohort_allele_count: variables.cohort_allele_count_spec})
+    AC = ds[cohort_allele_count]
+    K = AC.sum(dim="alleles")
+    # avoid divide by zero
```
jeromekelleher, line 509: Should we be returning 0 in this case or NaNs? I think in other stats we decided it was simpler to let the maths libraries decide what the outcome of a 1/0 should be, but to suppress the warnings.

We should be consistent with the behaviour of the popgen stats here I think.

jeromekelleher (commented):
LGTM, one question about the NaN corner case semantics though.

sgkit/stats/aggregation.py
timothymillar: AFAIK runtime warnings still can't be suppressed in Dask. I'll just let them be raised and suppress them in the tests where needed (as in Tajima's D).

timothymillar:
Thanks for the review  @jeromekelleher

jeromekelleher (approved):
LGTM!

tomwhite:
Looks like this is causing an existing test to fail.

timothymillar:
@tomwhite I think the test failure was due to #778 which was unrelated (change upstream). Can you restart the builds?

tomwhite:
> @tomwhite I think the test failure was due to #778 which was unrelated (change upstream). Can you restart the builds?

Yes, that looks like it was the problem. I can't see a button to re-run the build - maybe try rebasing?

PR #796, opened 2022-01-11
Upgrade sphinx_autodoc_typehints to fix doc build

tomwhite:
More churn in the doc packages (scanpydoc was upgraded to require latest sphinx_autodoc_typehints).

PR #798, opened 2022-01-11
Mergify strict mode deprecation

### LINKED ISSUE #797: Mergify strict mode deprecation
We need to fix Mergify's configuration due to https://blog.mergify.com/strict-mode-deprecation/

tomwhite:
Fixes #797

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,3 +1,15 @@
+queue_rules:
+  - name: default
+    conditions:
+      - base=main
+      - status-success=build (3.7)
+      - status-success=build (3.8)
+      - status-success=win_build (3.7)
+      - status-success=win_build (3.8)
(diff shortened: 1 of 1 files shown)


tomwhite:
Hmm not sure how to fix the mergify config here.

jeromekelleher:
I think we have a chicken and egg situation with #796 - how about we merge that first manually then rebase this?

tomwhite:
Rebasing. Still not sure I have the config right though, as the `condition` in `queue_rules` duplicates that in `pull_request_rules`. It might be OK, but seems a bit redundant.

jeromekelleher:
I doesn't seem to be rerunning, and the config looks OK to me. Maybe we should merge this manually, and see if mergify is then happy with some subsequent PRs?

PR #812, opened 2022-01-20
Pin numpy < 1.22 since numba 0.55 doesn't work with later versions

### LINKED ISSUE #809: Upstream failing due to numpy/numba conflict
With `numpy==1.22.1` and `numba==0.55.0` we get `ImportError: Numba needs NumPy 1.21 or less`.

This is not happening on the main build because another dependency is specifying a lower numpy version (I think).

tomwhite:
Fixes #809


ravwojdyla (approved):

jeromekelleher (approved):
LGTM

Issue #820, opened 2022-02-07
Investigate upgrading rechunker


tomwhite:
We are pinned to rechunker 0.3.3 due to this issues: https://github.com/pangeo-data/rechunker/issues/92.

However, the latest version of Dask (2022.01.1) doesn't work with rechunker 0.3.3 (see #819).

We should work out how to make progress here.

rabernat:
Tom I'm sorry for the friction this is causing you. We need to do some serious work on rechunker to fix the performance problem introduced in 0.4. Hopefully we can get to that very soon.

How do you know rechunker doesn't work with dask 2022.01.1. Could you open an issue for that in rechunker?

tomwhite:
Thanks Ryan! It would be great to get to the bottom of what is happening in rechunker 0.4 so it can be fixed.

I will open an issue in rechunker about the dask problem.

tomwhite:
@eric-czech do you remember what caused us to pin rechunker to 0.3.3? We use it directly in the bgen reader, so I wonder if it was when you were working with biobank data? It would be great if we could verify that we can upgrade rechunker with the new fix in main, but I don't have an sgkit test case to check it with.

eric-czech:
> @eric-czech do you remember what caused us to pin rechunker to 0.3.3?

Hm I don't and I'm pretty sure the last version I ever tested at scale on bgen data was 0.3.1 (I was using this [environment](https://github.com/related-sciences/ukb-gwas-pipeline-nealelab/blob/231d0e6e74772d1ae56205723952a5fff955aa74/envs/gwas.yaml#L32)).  I could be wrong, but I don't recall adding a pin in sgkit for any particular reason.

eric-czech:
> I don't have an sgkit test case to check it with

You're worried the current rechunking test cases for bgen_reader don't cover potential issues at a larger scale?

tomwhite:
Thanks @eric-czech!

> You're worried the current rechunking test cases for bgen_reader don't cover potential issues at a larger scale?

Yes. In yesterday's meeting we decided to upgrade to the new rechunker (0.5.0), and fix any scalability problems if they arise.

rabernat:
I'm eager to hear whether 0.5 works for you. It's hard to cover very large computations with CI, but in all of my ad hoc testing, the scalability problems from 0.4 have disappeared.

eric-czech:
> You're worried the current rechunking test cases for bgen_reader don't cover potential issues at a larger scale?
[...]

PR #837, opened 2022-04-08
Allow chunking in the samples dimention for identity_by_state

### LINKED ISSUE #836: Allow chunking in the samples dimention for method identity_by_state
The upstream issue has been (miraculously) solved https://github.com/dask/dask/issues/8131#issuecomment-1055709277. Removing the check for chunking also simplifies the code.

timothymillar:
Fixes #836

AFAICT the upstream bug was fixed in dask version 2022.01.0, so the new tests will fail on older versions

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 numpy < 1.22
 xarray
-dask[array] >= 2021.10.0
-distributed >= 2021.10.0
+dask[array] >= 2022.01.0
+distributed >= 2022.01.0
 dask-ml
 scipy
 typing-extensions
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,8 +28,8 @@ python_requires = >=3.7
 install_requires =
     numpy < 1.22
     xarray
-    dask[array] >= 2021.10.0
-    distributed >= 2021.10.0
+    dask[array] >= 2022.01.0
+    distributed >= 2022.01.0
     dask-ml
     scipy
     zarr >= 2.10.0, < 2.11.0
--- a/sgkit/stats/ibs.py
+++ b/sgkit/stats/ibs.py
@@ -48,17 +48,6 @@ def identity_by_state(
     which is a matrix of pairwise IBS probabilities among all samples.
     The dimensions are named ``samples_0`` and ``samples_1``.
 
-    Raises
-    ------
-    NotImplementedError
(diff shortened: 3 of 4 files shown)


timothymillar:
Failing on windows due to older dask version

jeromekelleher:
LGTM @timothymillar - should we pin to dask versions >= 2022.01.0 ?

tomwhite:
We are restricted to dask <= 2022.01.0 because of #820. A couple of possibilities:
1. Pin to dask == 2022.01.0 (i.e. has to be exactly this version), or
2. Pin to dask >= 2022.01.0 - if we think the rechunker scaling issue is fixed (or we're prepared to fix it if we see it again).

timothymillar:
> should we pin to dask versions >= 2022.01.0 ?

There's no hurry to merge this, it can wait until #820 is resolved. I just wanted to make sure that it's tidied up eventually.

tomwhite:
#820 is now resolved, so you should be able to update the pin to `dask >= 2022.01.0`.

jeromekelleher (approved):
LGTM

requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy < 1.22
 xarray
-dask[array] >= 2021.10.0
-distributed >= 2021.10.0
+dask[array] >= 2022.01.0
+distributed >= 2022.01.0
```
tomwhite, line 4: This needs changing in setup.cfg too. (I often forget to do that.)

tomwhite (approved):

requirements.txt
hammer: Perhaps we should add a `Dependencies` section to [Contribute to sgkit](https://pystatgen.github.io/sgkit/latest/contributing.html#contributing-to-code) that mentions this detail and any other tricky bits you can think about? Python dependency management is terrifying to me and any wisdom you can share would be much appreciated!

timothymillar: Ah, thanks

tomwhite: > Perhaps we should add a `Dependencies` section to [Contribute to sgkit](https://pystatgen.github.io/sgkit/latest/contributing.html#contributing-to-code) that mentions this detail and any other tricky bits you can think about?

Good idea. Opened #847

PR #839, opened 2022-04-11
Add basic pedigree schema

### LINKED ISSUE #838: Basic pedigree schema
See discussion in #786

A basic schema consists of:

- `sample_id` of shape `(samples, )` containing unique identifiers for samples (already implemented).
- `parent_id` of shape `(samples, parents)` containing unique identifiers of the parents of each sample where identifiers which are not found within `sample_id` are treated as unknown/missing parents.

From these we can calculate array `parent` of shape `(samples, parents)` containing the integer indices of each samples parents within the samples dimension. This should allow for faster numerical methods than using the unique identifiers directly. The justification for still including the `parent_id` array is that it is necessary to recompute the indices within the `parent` array following any alterations to the samples dimension.

Example:

```python
>>> ds = xr.Dataset()
>>> ds["sample_id"] = ["samples"], ["mom", "dad", "child"]
>>> ds["parent_id"] = ["samples", "parents"], [
...    [  ".",   "."],
...    [  ".",   "."],
...    ["mom", "dad"]
... ]
>>> # generate indices
>>> ds = parent_indices(ds)
>>> ds.parent
... array([[-1, -1],
...        [-1, -1],
...        [ 0,  1]]
>>> # subset samples
>>> ds = ds.sel(dict(samples=[True, False, True]))
>>> ds.parent
... array([[-1, -1],
...        [ 0,  1]]  # second index indicates that the child is its own parent
>>> # regenerating indices results in MergeWarning
>>> ds = parent_indices(ds)
>>> ds.parent
... array([[-1, -1],
...        [ 0, -1]]  # the child is now correctly recorded as having an unknown parent
```

timothymillar:
Fixes #838

### DIFF
--- a/sgkit/stats/pedigree.py
+++ b/sgkit/stats/pedigree.py
@@ -13,6 +13,7 @@ def parent_indices(
     *,
     sample_id: Hashable = variables.sample_id,
     parent_id: Hashable = variables.parent_id,
+    missing: Hashable = ".",
     merge: bool = True,
 ) -> Dataset:
     """Calculate the integer indices for the parents of each sample
@@ -29,6 +30,9 @@ def parent_indices(
     parent_id
         Input variable name holding parent_id as defined by
         :data:`sgkit.variables.parent_id_spec`.
+    missing
+        A value indicating unknown parents within the
+        :data:`sgkit.variables.parent_id_spec` array.
     merge
         If True (the default), merge the input dataset and the computed
         output variables into a single dataset, otherwise return only
@@ -38,6 +42,13 @@ def parent_indices(
     Returns
     -------
     A dataset containing :data:`sgkit.variables.parent_spec`.
+
+    Raises
+    ------
+    ValueError
+        If the 'missing' value is a known sample identifier.
+    KeyError
+        If a parent identifier is not a known sample identifier.
 
     Warnings
     --------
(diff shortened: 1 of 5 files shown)


sgkit/stats/pedigree.py
```diff
@@ -0,0 +1,78 @@
...
+
+from sgkit import variables
+from sgkit.utils import conditional_merge_datasets, create_dataset
+
+
+def parent_indices(
+    ds: Dataset,
+    *,
+    sample_id: Hashable = variables.sample_id,
+    parent_id: Hashable = variables.parent_id,
```
jeromekelleher, line 15: I think we should have a ``missing`` parameter which lets us specify the missing data character, and then throw an error if there's any IDs not in sample_id + missing.

How about ``missing="."``?

jeromekelleher (commented):
LGTM, but I do think we need to be careful about how we treat missing data.

See suggestion above.

@tomwhite, any thoughts here on the missingness API for this?

sgkit/stats/pedigree.py
tomwhite: This sounds reasonable to me.

timothymillar: Good suggestion! Are others happy with `"."` as the default? I just chose it for the examples based upon similarity with the VCF spec.

jeromekelleher: I think "." is as good as any. Another option would be "NA", but :shrug:

PR #841, opened 2022-04-14
Unpin rechunker now that version 0.5.0 is out

### LINKED ISSUE #820: Investigate upgrading rechunker

tomwhite:
Fixes #820.

Also remove the Dask pin as that was imposed by rechunker == 0.3.0.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,7 +12,7 @@ msprime>=1.0
 scikit-learn
 partd
 bed-reader
-rechunker == 0.3.3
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
:tada:

Issue #845, opened 2022-04-22
Pedigree based kinship estimation


timothymillar:
See discussion in #786, depends on #844

We probably want to start with separate implementations for the diploid and autopolyploid cases. The diploid implementations should only require the pedigree structure itself as specified by the `parent` array. There are fast approximate algorithms for diploids (e.g. [Kirkpatrick et al 2019](https://academic.oup.com/bioinformatics/article/35/6/1002/5085372)) but I suggest starting with the exact method and only looking at approximate methods if/when there is demand for it.

For the autopolyploid (and mixed-ploidy) case see [Kerr et al (2012)](https://pubmed.ncbi.nlm.nih.gov/22311370/) and [Hamilton and Kerr (2017)](https://pubmed.ncbi.nlm.nih.gov/29260268/). Their method requires two additional input arrays for the 𝜏 and 𝜆 parameters described within the latter paper. These respectively correspond to gametic ploidy and the additional IBD probability introduced by various meiotic processes. In each case, these should have the same shape as the `parent` array (samples * parents). I'd suggest naming them `Hamilton_Kerr_tau` and `Hamilton_Kerr_lambda`.

A basic API would be `pedigree_kinship(ds, method='diploid')` -> a Dataset with a square matrix `stat_pedigree_kinship` with dimensions `(samples_0, samples_1)`.

One consideration for the  `stat_pedigree_kinship` matrix is whether the diagonal values should contain the self-kinship coefficient or the inbreeding coefficient of each individual. My preference is to report the self-kinship and this matches our implementations of `Weir_Goudet_beta`, `identity_by_state` and `pc_relate`. It's worth being explicit about this as quite a few other tools default to reporting inbreeding on the diagonal.

PR #855, opened 2022-05-06
pedigree kinship

timothymillar:
See #845

TODO:

- [x] return dask arrays rather than numpy arrays
- [x] large diploid test case against published software
- [x] ~~consider parallelization~~


timothymillar:
This implementation has OK performance, and is designed to to work on unsorted pedigrees by calculating the sort order and iterating over samples in the corrected sort order without actually sorting the kinship matrix. This works but it does leave some performance on the table

I have also tried parallelization using numba with promising results. This is done simply by using `prange` in place of `range` for the inner loop within `kinship_diploid` or `kinship_Hamilton_Kerr`. This inner loop calculates all the off-diagonal elements within the output matrix (i.e., the cause of O(N^2) complexity). 

Benchmark results ([code](https://gist.github.com/timothymillar/91cdfd8a936ad517a16c620685acb5ba)):
pedigree_kinship_benchmark
- "autosort" is the implementation in this PR
- "presort" requires the samples to be in topological order already

In both the serial and parallel cases there is a lot of performance to be gained by pre-sorting the samples into topological order. However, I dont think we want to require this for the actual dataset because sorting huge datasets by the samples dimension is itself a slow/inefficient process. Other options would be (a) to implement both and check the topological order before applying, or (b) to sort a copy of the parent array, then compute the result and resort the resulting matrix which is inefficient.

sgkit/stats/pedigree.py
```diff
@@ -165,3 +170,524 @@ def topological_argsort(parent: ArrayLike) -> ArrayLike:  # pragma: no cover
...
+        If the parents dimension does not have a length of 2.
+    """
+    if parent.shape[1] != 2:
+        raise ValueError("Parent matrix must have shape (samples, 2)")
+    if not allow_half_founders:
+        _raise_on_half_founder(parent)
+    n = len(parent)
+    kinship = np.empty((n, n), dtype=dtype)
+    ordered = _is_pedigree_sorted(parent)
+    if ordered:
```
jeromekelleher, line 275: Wouldn't it be simpler to require that the dataset be ordered, or to have the argsorting stored in the dataset? I worry that (a) we're duplicating code here and (b) that people will end up calling ``topological_argsort`` many times when running different algorithms on the pedigree.

It wouldn't be so bad to raise an error if the sorting wasn't present, would it?

jeromekelleher (commented):
Very nice @timothymillar, really very pretty code. One structural comment.

sgkit/stats/pedigree.py
timothymillar: I did consider requiring sorted data but that requires sorting the full dataset by the samples dimension (including all genotype data and etc.) which isn't desirable. Storing the sort order is an option but it's a bit odd because the ith element in the sort order doesn't correspond to the ith sample, so it doesn't really fit with the xarray paradigm. `topological_argsort`is actually quite cheap and is trivial compared to the kinship matrix (O(N^2)).

timothymillar:
I seem to be getting sporadic test fails in CI but not locally, will need to dig into it.

sgkit/stats/pedigree.py
jeromekelleher: Yeah, good point. Could we avoid the code duplication then by having the argsort = np.arange(...) in the ordered case? I doubt it would make much perf difference (but maybe I'm wrong?)

timothymillar:
No, the perf issue is primarily due to the indexing on the "order" array within the inner loop. In the code shown in that figure I actually included the calculation of the order with topological_argsort even when it wasn't used (as in presort) to ensure that it was the inner loop being benchmarked.

timothymillar:
I was also surprised but it seems to be a general issue (unless my laptop is special), here's a minimal example:
<details>
<summary>code</summary>
<br>

```python
import timeit
import numpy as np
from numba import njit

njit()
def foo(order):
    n = len(order)
    out = np.empty(n)
    for i in range(n):
        out[i] = 1
    return out

njit()
def bar(order):
    n = len(order)
    out = np.empty(n)
    for i in range(n):
        j = order[i]
        out[j] = 1
    return out

a = np.arange(1000_000)
loops = 100

# warmup
foo(a)
bar(a)

timeit.timeit('foo(a)', globals=globals(), number=loops)/loops  # result: 0.0896
timeit.timeit('bar(a)', globals=globals(), number=loops)/loops  # result: 0.2101
```

</details>

> might be worthwhile making a copy of the parent array that's in order, in the out-of-order case?

I'll look into it but AFAICT that would require re-sorting the resulting matrix in both dimensions to match the samples order in the dataset. I think there's a question to be asked around whether we should optimize for speed or memory usage here. With the current version on my laptop, I can calculate a kinship matrix for 25,000 samples in around 1.1s or 3.3 sec for the sorted and unsorted cases respectively. The resulting matrix is around 4.65GB.

jeromekelleher:
Well, you learn something new every day. Perf stuff always surprises me, it's amazing how wrong intuition can be.

If it requires a full resorting of the dataset then it's definitely not worth it. I thought we could keep the sorted ``parents`` array and operate on that, but if not then what you have is great. Might be worth dropping in a few comments reflecting the perf journey here for future travellers though.

timothymillar:
I've removed the reporting of inbreeding coefficients because these can be calculated more efficiently without computing the full kinship matrix. I'll open a separate issue for this.

timothymillar:
This is essentially ready although I'm still getting sporadic test failures in CI. I think it relates to the parallelization...

timothymillar:
I'm stumped on the test failures. I have managed to get some failures locally but only very occasionally.
The symptoms are:
- it has only occurred in `test_pedigree_kinship__Hamilton_Kerr`
- the test fails due to some nan values in the kinship matrix, the location of these can vary
- there is no consistency to which iteration of the the test (different sample orders etc)

Things that **may** increase test failure rates (but are not required for a failure):
- running in the GitHub CI environment
- running multiple tests in parallel
- recompilation of numba functions (e.g., after clearing the cache)

One possibility is that there's a race condition in my code, but I can't identify any. There are multiple open bug reports relating to numba `parallel=True`/`prange` so I'm inclined to think this is a numba issue. Unfortunately I'm struggling to get a reliable example together for the sake of opening a new issue. 

I'd suggest we avoid parallelization via a numba for now and leave a TODO to revisit it later. I'll try pushing a few changes with a single core implementable and see if it has any issue in CI. Does anyone have thoughts on keeping the parallel diploid implementation? It hasn't failed yet although I'm concerned that it may be subject to the same underlying bug.

timothymillar:
I've identified the issue as an error in my code rather than a numba bug (sorry numba devs). It's unrelated to parallelism :facepalm:.

The issue occurred when encoding (half-)clones with a genomic contribution of tau=0 from an unknown parent (parent=-1) in combination with initializing the kinship matrix with `np.empty`. I hadn't explicitly handled the lookup of the parents kinship so it was retrieving the last value in the kinship matrix. This isn't an issue in theory because this kinship value will always be multiplied by the tau value of 0. An exception to this is when that incorrect kinship value is *occasionally* initialized with a nan. I switched from initializing the kinship array with `np.zeros` to `np.empty` shortly before switching to working on parallelism and I guess the test didn't fail in the interlude just by chance. The lesson learnt is that one should try initializing an array with nan values before using `np.empty`.

I've fixed the bug and added a test for it so this PR should be ready to go now. Of course we could actually add parallelism back in now although I'm happy to keep things single threaded until there's demand for it.

jeromekelleher:
Ooh, nasty one! Well spotted.

FWIW I don't think np.empty() is ever worth it - memset(buff, 0) is a pretty fast operation.

sgkit/stats/pedigree.py
```diff
@@ -165,3 +170,479 @@ def topological_argsort(parent: ArrayLike) -> ArrayLike:  # pragma: no cover
...
+           [0.25 , 0.25 , 0.5  , 0.375],
+           [0.375, 0.125, 0.375, 0.625]])
+
+    References
+    ----------
+    [1] - Matthew G. Hamilton, and Richard J. Kerr 2017.
+    "Computation of the inverse additive relationship matrix for autopolyploid
+    and multiple-ploidy populations." Theoretical and Applied Genetics 131: 851-860.
+    """
+    assert method in {"diploid", "Hamilton-Kerr"}
```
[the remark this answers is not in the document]
timothymillar: Good call, fixed.

timothymillar:
Thanks for reviewing @jeromekelleher!

PR #856, opened 2022-05-12
Fix mypy errors

tomwhite:
Fixes the latest set of build failures.


jeromekelleher (approved):
Happy to merge whenever you are

Looks like we have an unrelated failure coming through from Dask/scipy updates?

PR #862, opened 2022-05-31
A-matrix and inverse A-matrix

timothymillar:
Fixes #859 and #860 

This implements calculation of the A-matrix and its inverse. Calculation of the inverse follows Hamilton and Kerr for both the diploid and autopolyploid cases. I don't think a diploid specific implementation is necessary for the inverse as calculation of the (non-zero) values is O(N).

Todo:

- [x] Docstring examples
- [x] Tests (diploid case and half-founders)

~~I'm also wondering whether the `pedigree_` prefix on these functions is useful for clarity or unnecessarily verbose?~~ **Edit** prefix removed

@tomwhite are you happy with the changes made to [`define_variable_if_absent`](https://github.com/pystatgen/sgkit/commit/bc2ad704cd6bdd7cc6ed6ce674c0ca17533e68ba)? I think the use of kwargs is justified in this case.

PR #863, opened 2022-06-09
Remove mypy check in precommit

tomwhite:
As discussed in #862.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,9 +21,3 @@ repos:
     hooks:
       - id: flake8
         language_version: python3
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.931
-    hooks:
-      - id: mypy
-        args: ["--allow-redefinition", "--show-error-codes", "--ignore-missing-imports"]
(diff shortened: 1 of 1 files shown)


jeromekelleher (approved):
:tada:

ravwojdyla (approved):
> I'd be OK with that. It keeps breaking, and takes a while to fix each time. All for very little gain as far as I can tell. The simplest thing would be to remove the check from precommit.

This might be a case of an observer bias, because we have mypy in place we don't see that many typing related issues? To be fair IMO typing/mypy in python is annoying to say the least (especially in numerical libs), it was and certainly feels like an afterthought, nevertheless it's better than "nothing" and I do appreciate it more over time 🤔To play devil's advocate, I myself might be making this comment from a confirmation bias perspective (coming from mostly strongly typed background).

All-in-all I'm curious if we bring it back in the future or never look back :)

tomwhite:
Good points @ravwojdyla - I agree. Coming from a "strongly typed background" too, I had hoped that once we had added types to our code that we wouldn't need to change them. But it seems that with the numerical libraries still changing their type annotations, and mypy becoming more strict (or changing its behaviour perhaps - I'm not sure), it means that we get regular breakages. Maybe we can revisit when things settle down, but for the moment it seems like unnecessary overhead.

tomwhite:
I merged this to try to unblock the failures on main.

PR #867, opened 2022-06-20
Generic functions for cohort sums and means

### LINKED ISSUE #730: Generic functions for cohort means and sums

timothymillar:
Fixes #730

Adds four generic cohort reductions `cohort_sum`, `cohort_nansum`, `cohort_mean` and `cohort_nanmean`. Also updates `count_cohort_alleles` and `observed_heterozygosity` to use these functions. These functions aim to mimic standard numpy functions with regard to return dtypes (on Linux)  which results in `count_cohort_alleles` returning `uitn64` rather than `int32`. The changes also makes `count_cohort_alleles` more flexible in terms of chunking of the input `call_allele_count` array and, based on some simple benchmarks, there is a moderate performance improvement.

### DIFF
--- a/sgkit/tests/test_popgen.py
+++ b/sgkit/tests/test_popgen.py
@@ -516,6 +516,7 @@ def test_Garud_h__raise_on_no_windows():
         Garud_H(ds)
 
 
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
 @pytest.mark.parametrize("chunks", [((4,), (6,), (4,)), ((2, 2), (3, 3), (2, 2))])
 def test_observed_heterozygosity(chunks):
     ds = simulate_genotype_call_dataset(
@@ -581,6 +582,7 @@ def test_observed_heterozygosity(chunks):
     )
 
 
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
 @pytest.mark.parametrize("chunks", [((4,), (6,), (4,)), ((2, 2), (3, 3), (2, 2))])
 @pytest.mark.parametrize(
     "cohorts,expectation",
--- a/sgkit/stats/aggregation.py
+++ b/sgkit/stats/aggregation.py
@@ -8,7 +8,7 @@ from typing_extensions import Literal
 from xarray import Dataset
 
 from sgkit import variables
-from sgkit.stats.utils import assert_array_shape
+from sgkit.stats.utils import cohort_sum
 from sgkit.typing import ArrayLike
 from sgkit.utils import (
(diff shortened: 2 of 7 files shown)


jeromekelleher (approved):
LGTM, but the guvectorize stuff is over my head. 

Pinging @tomwhite for a look.

sgkit/tests/test_popgen.py
```diff
@@ -516,6 +516,7 @@ def test_Garud_h__raise_on_no_windows():
         Garud_H(ds)
 
 
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
```
tomwhite, line 519: What is this suppressing?

tomwhite (approved):
LGTM

sgkit/tests/test_popgen.py
timothymillar: This is suppressing a warning caused by dividing by zero (i.e., zero observations for a cohort because all observations are `nan`) which results in that `nan`s seen in the expected results. The previous implementation worked around this edge case to avoid the warning, however,  allowing the warning seems to be more consistent with what we do elsewhere. I'm happy reintroduce the workaround if it's preferred.

timothymillar:
Just a note on implementation. The reason that `da.map_blocks` works even when the input is chunked along the consumed/reduced axis is that dask concatenates chunks by the specified `drop_axis` prior to applying the mapped function (nice diagram in the [docs](https://docs.dask.org/en/latest/generated/dask.array.map_blocks.html)). This may not always be the most efficient approach but it's very convenient and we seem to generally discourage chunking in the `samples` dimension in the first place.

sgkit/tests/test_popgen.py
tomwhite: I think issuing the warning is the preferred way to do things. Thanks for the explanation!

tomwhite:
Might be worth adding the implementation note as a comment. Either way, feel free to add the auto-merge label when you are ready.

Issue #803, opened 2022-01-17
Run on Pyodide


tomwhite:
[Pyodide](https://github.com/pyodide/pyodide) uses WebAssembly to run Python in the browser. It has support for a lot of the PyData stack, so I wondered how easy it would be to get sgkit running on it. It would be a nice way to share demos and notebooks (see [JupyterLite](https://jupyterlite.readthedocs.io/en/latest/)). (This is work I did last year but didn't get round to sharing.)

The following libraries are not supported yet:
* Dask distributed. There is some discussion on https://github.com/dask/dask/issues/7764. The synchonous scheduler does work though, with a small workaround.
* Numba. Ideally numba decorators would be ignored, but for the demo below I just commented them out. (There is a problem with doing this for `guvectorize` since it generates code with a new signature, so anything that uses these functions won't work.)
* IO libraries. In principle these could be submitted to Pyodide as a [new package](https://pyodide.org/en/stable/development/new-packages.html).

I created a [branch](https://github.com/tomwhite/sgkit/tree/pyodide) with the above changes (and a few others), then built a wheel and uploaded to GCP in order to load it with [micropip](https://pyodide.org/en/stable/usage/api/micropip-api.html). Then using https://pyodide.org/en/latest/console.html, I managed to create an sgkit Dataset:

```
Welcome to the Pyodide terminal emulator 🐍
Python 3.9.5 (default, Jan 17 2022 04:07:25) on WebAssembly VM
Type "help", "copyright", "credits" or "license" for more information.
>>> import micropip
>>> import zarr # not sure why this is needed before installing sgkit
>>> import sklearn # needed since sgkit doesn't explicitly declare it as a dependency (need to fix)
>>> await micropip.install("https://storage.googleapis.com/tomwhite_test/sgkit-0.3.1.dev5%2Bg59736c0-py3-none-any.whl")
>>> # needed to import dask, see https://github.com/pyodide/pyodide/issues/1603
>>> import sys
sys.modules['_multiprocessing'] = object
>>> 
>>> import dask
>>> dask.config.set(scheduler='synchronous')
<dask.config.set object at 0x347de48>
>>> import sgkit as sg
/lib/python3.9/site-packages/pandas/compat/__init__.py:124: UserWarning: Could not import the lzma module. Your installed Python is incomplete. Attempting to use lzma com
pression will result in a RuntimeError.
  warnings.warn(msg)
[...]

PR #870, opened 2022-07-03
Toggle numba caching by environment variable

### LINKED ISSUE #371: Create a local version of guvectorize
That has `nopython=True` and `cache=True`, so they don't get inadvertently missed on new functions. See https://github.com/pystatgen/sgkit/pull/364#discussion_r515922535

### LINKED ISSUE #869: Disable numba caching via environment variable
Edit: related to #371

I've recently started experimenting with sgkit on a [SLURM cluster](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html) which is working well with the exception of methods using `guvectorize` with `cache=True`. Calling these functions results in a segmentation fault on the worker. This only seems to be an issue with `guvectorize` (not the `jit` or `vectorize` decorators) and there is no segmentation fault if I set `cache=False`.

There are a couple of open issues that may be related although neither quite match what I'm seeing (need to dig some more): 
- https://github.com/dask/distributed/issues/3450
- https://github.com/numba/numba/issues/4807

There is also an open issue for globally disabling numba caching which would provide a workaround although it might be stale:
- https://github.com/numba/numba/issues/4549

In the meantime, for the sake of debugging and workarounds, it'd be useful to be able to disable numba-caching in sgkit using an environment variable.

timothymillar:
Fixes #371 and #869 

This is just a suggestion and I'm not sure of the best way to test this in CI unless we create an additional runner?

### DIFF
--- a/sgkit/accelerate.py
+++ b/sgkit/accelerate.py
@@ -1,0 +1,31 @@
+import os
+from typing import Callable
+
+from numba import guvectorize, jit
+
+_DISABLE_CACHE = os.environ.get("SGKIT_DISABLE_NUMBA_CACHE", "0")
+
+try:
+    CACHE_NUMBA = {"0": True, "1": False}[_DISABLE_CACHE]
+except KeyError as e:  # pragma: no cover
+    raise KeyError(
+        "Environment variable 'SGKIT_DISABLE_NUMBA_CACHE' must be '0' or '1'"
+    ) from e
+
+
+DEFAULT_NUMBA_ARGS = {
+    "nopython": True,
+    "cache": CACHE_NUMBA,
+}
+
+
+def numba_jit(*args, **kwargs) -> Callable:  # pragma: no cover
+    kwargs_ = DEFAULT_NUMBA_ARGS.copy()
+    kwargs_.update(kwargs)
+    return jit(*args, **kwargs_)
+
+
+def numba_guvectorize(*args, **kwargs) -> Callable:  # pragma: no cover
+    kwargs_ = DEFAULT_NUMBA_ARGS.copy()
+    kwargs_.update(kwargs)
+    return guvectorize(*args, **kwargs_)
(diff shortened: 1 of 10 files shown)


jeromekelleher:
Totally agree we should do this, but I wonder if it'd be better if we just made our own local ``jit`` decorator? This would allow us to put in library-wide defaults for things like ``cache``, as well as give us more fine-grained control of when jitting is done.

tomwhite:
> Totally agree we should do this, but I wonder if it'd be better if we just made our own local `jit` decorator? This would allow us to put in library-wide defaults for things like `cache`, as well as give us more fine-grained control of when jitting is done.

Agree - see #371.

timothymillar:
Ah, thanks I hadn't noticed that issue. I can make a start on custom `jit`, `vectorise` and `guvectorise` decorators.

tomwhite:
Nice! I'm not sure how easy it would be to completely disable numba for `vectorize` and `guvectorize` since these wrappers change the type signature in a way that is non-trivial to do in pure python. We don't need that here though (but I'll note that it would be useful for #803) - just a way to enforce numba caching consistently.

timothymillar:
I've started with some simple wrappers for `jit` and `guvectorize`. I ran into this issue https://github.com/numba/numba/issues/8231 when wrapping `vectorize`, but then realized that the only use of it (mine) was completely unnecessary 🤷. As @tomwhite pointed out, we can't really disable numba when using `guvectorize` so I'm ignoring that for now. I'm also leaving out `cuda` for now as I don't have convenient means to test it.

jeromekelleher (approved):
LGTM!

timothymillar:
@tomwhite are you happy that this satisfies #371? Or would you prefer to also handle `@cuda` now?

tomwhite:
@timothymillar sorry I was away last week on vacation and have only just got back. I'm fine leaving cuda for now.

timothymillar:
No worries, thanks @tomwhite

Issue #882, opened 2022-08-03
Error using vcf_to_zarr with HLA contigs


d-laub:
HLA contigs have colons and dashes in their names and I believe this isn't addressed by the `vcf_to_zarr` implementation, specifically the `get_region_start` function:

https://github.com/pystatgen/sgkit/blob/d08feba59415fa502150ad1d052e5377cdb83a94/sgkit/io/vcf/vcf_reader.py#L94-L100

This causes a "too many values to unpack" exception since HLA contigs can be, for example, HLA-A*01:01:01:01. I'm not super familiar with the `sgkit` codebase but the fix may be as simple as changing the implementation of `get_region_start` to something like this:

```python
import re
...
def get_region_start(region: str) -> int:
    """Return the start position of the region string."""
    # check that region ends with start and end coordinates
    if not re.match("\d+-\d+$", region): 
        return 1
    contig, start_end = region.rsplit(":", 1)
    start, end = start_end.split("-")
    return int(start)
```

This change fixed the errors in my case and I submitted a PR #883.

tomwhite:
Fixed in #883

PR #883, opened 2022-08-03
Fix get_region_start to work with contig names that have colons and dashes.

d-laub:
Addresses issue #882.

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -15,6 +15,7 @@ from typing import (
     Tuple,
     Union,
 )
+import re
 
 import dask
 import fsspec
@@ -93,9 +94,9 @@ def region_filter(
 
 def get_region_start(region: str) -> int:
     """Return the start position of the region string."""
-    if ":" not in region:
+    if not re.match("\d+-\d+$", region):
         return 1
-    contig, start_end = region.split(":")
+    contig, start_end = region.rsplit(":", 1)
     start, end = start_end.split("-")
     return int(start)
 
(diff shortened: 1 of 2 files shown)


tomwhite:
Thanks for opening an issue and PR to fix it @d-laub!

The code looks good to me. Would you be able to add a short unit test of this function, perhaps in https://github.com/pystatgen/sgkit/blob/main/sgkit/tests/io/vcf/test_utils.py?

d-laub:
@tomwhite thanks for the suggestion. I gave it a shot and caught a bug. One thing I'll note is that this function will still fail if a contig name ends with <digits>-<digits>. Hopefully no one has or will name a contig in such a way. Otherwise, let know if there's anything else I can do.

tomwhite (commented):
Thanks for writing a test @d-laub. It looks good to me.

The build is failing, however, as we have pre-commit checks for import order etc. In this case, the `import re` line needs moving up, see https://github.com/pystatgen/sgkit/runs/7668388547?check_suite_focus=true#step:5:241.

You might want to  install pre-commit locally to catch these things before submitting them, see https://pystatgen.github.io/sgkit/latest/contributing.html#install and https://pystatgen.github.io/sgkit/latest/contributing.html#coding-standards

jeromekelleher (approved):

tomwhite:
It looks like the latest failures are when running tests against real VCF files, not pre-commit failures. It will need some digging to see if that is a problem introduced with this change, or whether it has uncovered a bug.

I can take a look but it probably won't be for a few days.

d-laub:
Yea I don't know how the ground truth was generated for the number of variants in those real VCF tests. My first thought was to compare to `bcftools stats` but like Tom I won't have time to dig into this for a few days.

tomwhite:
+1 to using `bcftools` for the ground truth here.

d-laub:
Turns out some regions end with ":\d+-" so I caught that case and now all the vcf tests passed locally. 🤞🏼

tomwhite (approved):
LGTM

Issue #609, opened 2021-06-15
Consider adding mean imputation function


eric-czech:
For association testing and PCA (at least), it may be useful to have a function that imputes dosages/allele counts.  With floating point values (i.e. from bgen), this can be very simple as a user, e.g. ```ds.call_genotype_probability.fillna(ds.call_genotype_probability.mean(dim="samples"))```.  With alternate allele counts having a sentinel integer, it is a little more complicated.  The best way to do this would be to expose a function like this publicly: https://github.com/pystatgen/sgkit/blob/20f4992bf1e4c1e09152ad930afd859cd012281d/sgkit/stats/pc_relate.py#L16-L22

I think a good signature for this would be:

```python
def mean_impute(ds: Dataset, variable: str, dim: str, merge: bool) -> Dataset
```

This is an unusual use case since the resulting variable should probably be something like ```f"{variable}_imputed"``` by default.  Setting ```merge=False``` and doing the rename manually would be one escape hatch for working with a different naming convention.

----

For reference, this is a workaround for any users that need this in the meantime:

```python
def mean_impute(ds, variable, dim):
    unmasked = ~ds[f"{variable}_mask"]
    return ds.assign(**{
        f"{variable}_imputed": ds[variable].where(
        unmasked, 
        ds[variable].where(unmasked).mean(dim=dim)
    )})
    
# To mean impute values for variants:
ds = mean_impute(ds, 'call_genotype_probability', 'samples')
```

PR #892, opened 2022-08-17
Add mean imputation function

tszfungc:
ref #609 

Add mean impute function for `call_dosage`, `call_genotype`, and `call_genotype_probability`

### DIFF
--- a/sgkit/stats/preprocessing.py
+++ b/sgkit/stats/preprocessing.py
@@ -1,4 +1,4 @@
-from typing import Hashable, Optional
+from typing import Hashable, Optional, Sequence
 
 import dask.array as da
 import numpy as np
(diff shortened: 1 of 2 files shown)


timothymillar:
Thanks for looking into this @tszfungc! I think this could be a great approach for imputing `call_dosage` and `call_genotype_probability`. However, I don't think it will produce the desired result for `call_genotype`. 

The values in `call_genotype` are (potentially unsorted) alleles whose order along the `ploidy` dimension doesn't have any particular meaning. So, as far as I can tell, the mean of those alleles can't really be used for anything.

tszfungc:
Thanks for the review @tomwhite @timothymillar. I agree that the allele order doesn't have a particular meaning. The order along `ploidy` should be ignored by computing the mean along `dim=['samples', 'ploidy']`, But this is also an unusual use to me.

sgkit/stats/preprocessing.py
```diff
@@ -185,3 +185,91 @@ def filter_partial_calls(
...
     return conditional_merge_datasets(ds, new_ds, merge)
+
+
+def mean_impute(
+    ds: Dataset,
+    variable: str,
+    dim: Union[Hashable, Sequence[Hashable]] = "samples",
+    merge: bool = True,
+) -> Dataset:
+    """Mean impute a masked variable
```
jeromekelleher, line 196: It would be helpful to give a more descriptive follow up sentence here, like say

> This replaces missing data for the specified variable with the mean of the non-missing values.

sgkit/variables.py
```diff
@@ -214,6 +214,15 @@ def _check_field(
     )
 )
 
+call_dosage_imputed, call_dosage_imputed_spec = SgkitVariables.register_variable(
```
jeromekelleher, line 217: I'm not sure we want to create a whole new bunch of variables here. Wouldn't it be simpler if we returned a copy of the original dataset in which all the missing data for the variable in question was replaced with the mean, and the mask was unset?

This would be more useful for downstream work, wouldn't it? We'd surely want to use the (say) imputed ``call_dosage`` in downstream analyses, and we wouldn't want to need to change variable names in order to do this.

jeromekelleher (commented):
Approach basically looks good to me, but I'm not convinced about the general approach of creating new ``_imputed`` variables. I would be simpler/better to just replace the missing data and reset the missingness mask in the returned dataset I think.

timothymillar:
@jeromekelleher the trade-off between returning new variables or replacing existing variables was previously discussed in https://github.com/pystatgen/sgkit/pull/308#issuecomment-705706571. I personally have a slight preference for replacing existing variables but there are some good points raised in that discussion. The primary concern seems to be that replacing existing variables is effectively a mutate operation, which goes against the general pattern of treating arrays as immutable.

jeromekelleher:
I see, thanks. Hmm, not much choice other than to create a bunch of new variables then.

PR #895, opened 2022-08-18
Replace additive_relationships with pedigree_relationship

### LINKED ISSUE #894: Replace `additive_relationships` method with a more generic `pedigree_relationship`
In hindsight, I dislike the way I implemented the `additive_relationships` method as it is too tightly coupled with `pedigree_kinship`. I think it'd be better to have a generic `pedigree_relationship` which does not automatically compute the kinship. This would allow combining the current `diploid` and `Hamilton-Kerr` options into a single `additive` option (the previous options are mostly just passed to the `pedigree_kinship` method). It would also allow the future addition of dominance methods in addition to additive methods.

timothymillar:
Fixes #894

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -74,7 +74,6 @@ Methods
 .. autosummary::
    :toctree: generated/
 
-   additive_relationships
    call_allele_frequencies
    cohort_allele_frequencies
    count_call_alleles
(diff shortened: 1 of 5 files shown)


jeromekelleher (approved):
I agree, this is nicer :+1:

tomwhite:
@timothymillar this should get auto-merged if you rebase, now #898 is in.

Issue #890, opened 2022-08-15
`test_max_str_len` failing due to change in upstream Dask


tomwhite:
See https://github.com/pystatgen/sgkit/runs/7831636851?check_suite_focus=true#step:5:1641

tomwhite:
Using git bisect I found the problematic change in Dask: https://github.com/dask/dask/pull/9268

I haven't looked into it, so not sure what the best fix is yet.

jeromekelleher:
The test itself is a bit complicated, and it looks like we're already special-casing the problematic empty array case. Maybe we should split it up into a three back-end specific tests, removing that parametrize? Hopefully that'll make it easier to follow/see what's going on.

PR #898, opened 2022-08-30
Pin dask to avoid test_max_str_len failure

tomwhite:


tomwhite:
I'm struggling to come up with a fix for #890, so let's pin Dask to unblock the other PRs and then we can figure out a better fix.

jeromekelleher (approved):
SGTM

PR #901, opened 2022-09-07
Replace inverse_additive_relationships with pedigree_inverse_relationship

### LINKED ISSUE #897: Replace `inverse_additive_relationships` with `pedigree_inverse_relationship`
Similar to #895, depends on #891

The inverse additive relationship matrix (ARM) is a sparse matrix which shouldn't require the full (dense) kinship matrix to be calculated in advance. The inverse ARM can be calculated from the same subset of parental kinships used to calculate the inbreeding coefficient in #891 and this should result in a similar speed up as seen in that PR. 
Currently `inverse_additive_relationships` returns the inverse ARM as a dense matrix, but ideally this would (optionally?) be returned as a sparse matrix. How do others feel about picking up [pydata/sparse](https://sparse.pydata.org/en/stable/) as a dependency?

timothymillar:
Fixes #897

This method uses the same approach as `pedigree_inbreeding` to calculate only a minimal subset of kinship values required for the inverse relationship matrix. Currently this method is still returning a dense matrix (as opposed to a sparse matrix), so the potential memory reduction isn't realized. However, it should be simple to generalize this code to a sparse encoding if/when required.

With some simple benchmarking using real pedigrees of approx 6500 samples (mentioned in #891) I'm seeing approximately a 3x speed up relative to the `pedigree_kinship` method which was a dependency of the old `inverse_additive_relationships` method. However, this speed up will depend on pedigree structure.

One of the changes in this PR is that the internal `inbreeding_Hamilton_Kerr` method now returns a `parent_kinship` array in addition to the `inbreeding` array. This is close to free and I don't see a noticeable slowdown in benchmarks.

I did consider having the public  `pedigree_inbreeding` method also return a `parent_kinship` array (i.e., add it to the schema), and then have the new `pedigree_inverse_relationship` consume both of these arrays. However, I chose to avoid this (at least for now) as we may want to add alternative methods of calculating inverse relationships that wouldn't consume those arrays.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -87,13 +87,13 @@ Methods
    hardy_weinberg_test
    identity_by_state
    individual_heterozygosity
-   inverse_additive_relationships
    ld_matrix
    ld_prune
    maximal_independent_set
    observed_heterozygosity
    pbs
    pedigree_inbreeding
+   pedigree_inverse_relationship
    pedigree_kinship
    pedigree_relationship
    pc_relate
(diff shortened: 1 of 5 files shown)


tomwhite (approved):
This looks good to me, but let's wait for @jeromekelleher to take a look too.

jeromekelleher (approved):
LGTM also

PR #903, opened 2022-09-14
VanRaden GRM

### LINKED ISSUE #874: VanRaden GRM
Implement the VanRaden genomic relationship matrix ([VanRaden 2008](https://www.sciencedirect.com/science/article/pii/S0022030208709901)). This is typically calculated from dosages for biallelic markers and has been generalized to autopolyploids (fixed ploidy).
One consideration is the specification of "base"/"ancestral"/"founder" allele frequencies. These are often estimated from the sample population but this assumes a population under HWE. I can't see a sensible default for estimating these so it's probably best to default to an array (a new variable) which must be pre-populated.

timothymillar:
Fixes #874 

I've tried to preemptively make this function suitable for additional estimators if/when required.

One point I'm unsure on is if we should have a default for `ancestral_frequency` based on the sample population. It's common practice to use the sample population allele frequency as an estimate of the ancestral allele frequency although this is often inappropriate. So the question is whether it's best to force an explicit choice by the user, or default to the (potentially inappropriate) most common choice?


jeromekelleher (approved):
Superb stuff @timothymillar :heart:

jeromekelleher:
> So the question is whether it's best to force an explicit choice by the user, or default to the (potentially inappropriate) most common choice?

I think it's best to force a choice in the first instance if there's any doubt. Otherwise you can get into nasty problems of backwards compabibility, if the default that you initially chose turns out to be not quite the thing you should have done.

tomwhite:
Is this ready to be merged or are you planning further changes @timothymillar?

timothymillar:
I think it's good to go, merging

PR #907, opened 2022-09-20
Allow use of Hamilton-Kerr methods with parents dim > 2

### LINKED ISSUE #906: Allow parents dimension > 2 for Hamilton-Kerr methods
Currently most of the pedigree methods for estimating kinship, inbreeding and etc. require that the parent dimension is of length two. I've run into this limitation when working with pedigree data structured with three parent columns - Mother, Father and Origin/Source to indicate clonal propagation. These data could be manually re-coded into two columns, but this is cumbersome to do in xarray (dimension resizing) and looses the clear distinction between parental types.

This limitation could be relaxed for the Hamilton-Kerr methods where the 'tau' parameter explicitly indicates contributions from parents. These methods will still be applicable so long as there are at most two contributing parents. Rather than generalizing the implementations of these methods an arbitrary parents dimension (which will likely incur a performance cost), a better option may be to "compress" the parent array (and associated arrays) down to two columns on the fly. This is a simple O(N) operation which would only apply if the parents dimension is not of length 2.

timothymillar:
Fixes  #906

### DIFF
--- a/sgkit/stats/pedigree.py
+++ b/sgkit/stats/pedigree.py
@@ -646,7 +646,7 @@ def pedigree_kinship(
         have a length of two.
     ValueError
         If the Hamilton-Kerr method is used and a sample has more than
-        two contributing parents as indicated by ``stat_Hamilton_Kerr_tau``.
+        two contributing parents.
     ValueError
         If the pedigree contains half-founders and allow_half_founders=False.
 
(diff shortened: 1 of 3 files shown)


timothymillar:
> It's not obvious to me where the specification of what's accepted in terms of the parent array is though

Good point, I've added some more documentation and examples which hopefully convey the details.

jeromekelleher (approved):

tomwhite:
Feel free to merge if this is ready @timothymillar

PR #833, opened 2022-04-01
Pin cyvcf2

tomwhite:
Fixes #830

PR #912, opened 2022-09-28
Remove cyvcf2 pin

### LINKED ISSUE #887: Wheels on macOS Python 3.10 failing
See https://github.com/pystatgen/sgkit/runs/7745453482?check_suite_focus=true

The problem is that cyvcf2 wheels for macOS Python 3.10 are only available from version [0.30.16](https://pypi.org/project/cyvcf2/0.30.16/#files), but that is built with NumPy 1.23 which is incompatible with Numba.

We should remove this from our build matrix for the time being.

tomwhite:
since https://github.com/brentp/cyvcf2/issues/248 is fixed and cyvcf2 0.30.17 has been released.

This fixes #887 and reverts #833.

I've tested that the wheels build successfully with these changes here: https://github.com/tomwhite/sgkit/actions/runs/3144967267/jobs/5111666983

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -46,10 +46,6 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
         python-version: ["3.7", "3.8", "3.9", "3.10"]
-        # see https://github.com/pystatgen/sgkit/issues/887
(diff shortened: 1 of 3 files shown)


jeromekelleher (approved):
LGTM

Issue #911, opened 2022-09-26
genotype_count_spec is inconsistent with VCF spec


timothymillar:
Just a minor inconsistency I noted given that we seem to follow the VCF spec in most cases. The `genotype_count_spec` states "the 3 columns contain heterozygous, homozygous reference, and homozygous alternate counts (in that order)". The VCF genotype ordering would typically be homozygous reference, heterozygous, homozygous alternate for a diploid biallelic variant as described for the "GL" field in the [spec](https://samtools.github.io/hts-specs/VCFv4.3.pdf).

timothymillar:
Just to add to that, If we were to change the ordering to match VCF ordering then this variable could easily be generalized to multi-allelic variants and autopolyploids.

tomwhite:
+1 to introducing a generalized ordering like GL.

It looks like the `genotype_count` variable is only used in `hardy_weinberg_test` - and it's optional. Does HWE generalize to multi-allelic/autopolyploids? If not, then we might leave HWE as it is, and rename its variable to `hwe_genotype_count_spec` or similar. If it does then we might want to consider changing it in HWE too, but may need to think about backwards compatibility (as it's optional it's possible it's not being used anywhere). I don't know why that order was used in HWE (@eric-czech might do).

timothymillar:
> Does HWE generalize to multi-allelic/autopolyploids?

Yes in general although I'm not familiar with implementation details (something on my long term todo list).

timothymillar:
@tomwhite I'm not sure how urgently you want to release 0.6.0 but I can try sort this tomorrow.

tomwhite:
Thanks Tim, that would be great - I'll try to release next week.

Issue #904, opened 2022-09-14
Include dimension names is variable specifications


timothymillar:
Related to #868 and #893

I'm wondering if we can start to check dimension names of variables without being too heavy handed by using warnings. This could potentially avoid manual comparison of array shapes and avoid issues like #868. It could also add make the variable documentation a bit more useful.

I'd suggest that we add dimension names as part of each variables specification. Outer dimensions could be optional, and a wildcard could be used when a dimension is not strictly defined. Dimension names could then be checked within the `validate_variables` function and a `DimensionWarning` thrown if there is a mismatch.

I also noticed that some dimension constants where initially created [here](https://github.com/pystatgen/sgkit/blob/main/sgkit/model.py#L11), but have mostly been specified using raw strings in other files.

Here's a rough idea of what the dimension checking code could look like:
```python
import warnings

class DimensionWarning(UserWarning):
    "Warning about dimension mismatches."
    pass

DIM_OPTION = None
DIM_WILDCARD = "*"
DIM_VARIANT = "variants"
DIM_WINDOWS = "windows"
DIM_CONTIG = "contigs"
DIM_SAMPLE = "samples"
DIM_COHORT = "cohorts"
DIM_PLOIDY = "ploidy"
DIM_ALLELE = "alleles"
DIM_GENOTYPE = "genotypes"
DIM_FILTER = "filters"

DIMS_POPULATION = {DIM_COHORT, DIM_SAMPLE}
DIMS_GENOMIC = {DIM_CONTIG, DIM_WINDOWS, DIM_VARIANT}

def dimension_match(dim, spec):
    if isinstance(spec, set):
        return (dim in spec) or (DIM_WILDCARD in spec)
    else:
        return (dim == spec) or (DIM_WILDCARD == spec)


def validate_dims(dims, spec):
    message = "Dimensions {} do not match {}".format(dims, spec)
    n_obs, n_exp = len(dims), len(spec)
    if  n_obs > n_exp:
        warnings.warn(message, DimensionWarning)
    diff = n_exp - n_obs
    for i in range(diff):
        if None not in spec[i]:
            warnings.warn(message, DimensionWarning)
    for dim, exp in zip(dims, spec[diff:]):
        if not dimension_match(dim, exp):
            warnings.warn(message, DimensionWarning)

# no warning
validate_dims(("windows", "samples", "ploidy"), (DIMS_GENOMIC, DIM_SAMPLE, DIM_PLOIDY))

# no warning due to wildcard
[...]

PR #913, opened 2022-09-29
Add dimension names to ArrayLikeSpec

timothymillar:
This is a proof of concept for #904.

It adds the option of named dimensions to `ArrayLikeSpec`. Individual dims can be optional, have multiple valid names, or be unnamed using a wildcard ("*"). 

If `dims` is specified without `ndim` then `ndim` will be calculated from `dims`. If they are both specified then they are checked for congruence. Dimension names are checked as part of variable validation and a `DimensionWarning` is given if the observed `dims` don't match the spec. A `DimensionWarning`  is also given if a spec is so vague that there is more than one possible mapping of array dims to spec dims (this is only possible with multiple optional dims and isn't an issue for any current variables).

I've possibly been a bit heavy handed with adding named dimensions to variables and some of these could be rolled back. I also identified a few slightly inconsistent variables that may be worth altering (possible breaking changes):

- `genotype_count` could have "genotypes" as second dimension but we should consider resolving  #911 first.
- `ld_prune_index_to_drop` has a unique dimension "variant_indexes" and holds integer indices for a subset of variants,
  I wonder if this could be replaced with a length "variants" array of booleans to avoid adding a new dimension?
- `variant_linreg_effect` doesn't appear to be used, can it be removed?

It would also be nice to automatically add the `ndim` and `dims` field to the `__doc__` string, is there a more elegant way to do this than just appending within the `__post_init__` method?

### DIFF
--- a/sgkit/variables.py
+++ b/sgkit/variables.py
@@ -21,10 +21,59 @@ class Spec:
 
 @dataclass(frozen=True, eq=False)
 class ArrayLikeSpec(Spec):
-    """ArrayLike type spec"""
+    """ArrayLike type spec.
+
+    Parameters
+    ----------
+    __doc__
+        Description of array.
+    kind
+        Array dtype kind following numpy conventions.
+        A set may be used to indicate multiple valid kinds.
+    ndim
+        Number of array dimensions.
+        A set may be used to indicate a variable number of dimensions.
+    dims
+        Tuple of expected dimension names.
+        A set may be used to indicate multiple valid names.
+        A set containing None may be used to indicate an optional dimension.
+        A wildcard ``"*"`` may be used to match any name.
+
+    Note
+    ----
+    If ndim is not specified and dims is specified then ndim will automatically
+    be calculated from the dims variable.
+
+    Raises
+    ------
+    ValueError
+        If conflicting ndim and dims are specified.
+    """
 
     kind: Union[None, str, Set[str]] = None
     ndim: Union[None, int, Set[int]] = None
+    dims: Union[None, Tuple[Union[Set[Union[None, str]], str]]] = None
+
+    def __post_init__(self):
+        if self.dims:
+            # calculate ndim from dims
+            maximum = len(self.dims)
+            optional = sum(
+                None in d if isinstance(d, set) else False for d in self.dims
+            )
+            if optional == 0:
+                ndim = maximum
+            else:
+                ndim = {maximum - i for i in range(optional + 1)}
+            if self.ndim:
+                # check correct
+                if ndim != self.ndim:
+                    raise ValueError(
+                        f"Specified ndim '{self.ndim}' does not match dims {self.dims}"
+                    )
+            else:
+                # use calculated ndim
+                object.__setattr__(self, "ndim", ndim)
 
 
 class SgkitVariables:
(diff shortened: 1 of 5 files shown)


timothymillar:
CI is failing on coverage due to the change made to `test_regenie_loco_regression`. 
@tomwhite I don't quite grok why there is a special case handled [here](https://github.com/pystatgen/sgkit/blob/main/sgkit/stats/association.py#L351) for a covariates array with an outer dimension of length 0? The documentation for this variable states "variables of shape (samples[, covariates])" which makes me think this should check for `covariates.shape[-1]`?

tomwhite:
> CI is failing on coverage due to the change made to `test_regenie_loco_regression`.
> @tomwhite I don't quite grok why there is a special case handled [here](https://github.com/pystatgen/sgkit/blob/main/sgkit/stats/association.py#L351) for a covariates array with an outer dimension of length 0? The documentation for this variable states "variables of shape (samples[, covariates])" which makes me think this should check for `covariates.shape[-1]`?

It looks like this was #509. Does the unit test introduced there still pass with your suggested change?

tomwhite:
> I also identified a few slightly inconsistent variables that may be worth altering (possible breaking changes):
> 
> * `genotype_count` could have "genotypes" as second dimension but we should consider resolving  [genotype_count_spec is inconsistent with VCF spec #911](https://github.com/pystatgen/sgkit/issues/911) first.

I think this would be a good change. You could do this as a follow on by omitting `dims` in this PR.

> * `ld_prune_index_to_drop` has a unique dimension "variant_indexes" and holds integer indices for a subset of variants,
>   I wonder if this could be replaced with a length "variants" array of booleans to avoid adding a new dimension?

Possibly, although that code path has had some performance work done on it, so changing it might have an unknown impact unless we measure it again. Another option would be not to name the dimension, and open an issue with a comment in the code.

> * `variant_linreg_effect` doesn't appear to be used, can it be removed?

Yes I think so.

sgkit/variables.py
```diff
@@ -671,7 +731,7 @@ def _check_field(
 stat_Garud_h1, stat_Garud_h1_spec = SgkitVariables.register_variable(
     ArrayLikeSpec(
         "stat_Garud_h1",
-        ndim={1, 2},
+        dims=("windows", "cohorts"),
```
tomwhite, line 734: It looks like the Garud stats are always windowed, so we should change their api doc to reflect that.

timothymillar:
> It looks like this was https://github.com/pystatgen/sgkit/issues/509. Does the unit test introduced there still pass with your suggested change?

That PR seems to relate to  `gwas_linear_regression` where I was referring to the code in `regenie_loco_regression`. The APIs of these methods are inconsistent in terms of the covariates parameter. The former expects a list of covariate variables where the latter expects a single covariates variable. They have very similar checks for `len(covariates) == 0` which actually do quite different things:

- When `gwas_linear_regression` checks `if len(covariates) == 0` it is checking how many "covariates" variables have been specified (i.e., the number of arrays which will be concatenated).
- When `regenie_loco_regression` checks `if len(covariates) == 0` it is checking the length of the first dimension of the single "covariates" array.

The check in `regenie_loco_regression` is odd to me because the description of the "covariates" array states that the first dimension should always have length=samples. Maybe the intent was to match the `gwas_linear_regression` API?

timothymillar:
Thanks @tomwhite, I've avoided the awkward variables for now.

tomwhite (approved):

timothymillar:
I'm thinking these changes should warrant a second reviewer or a bit more consensus?

jeromekelleher:
I wonder if @eric-czech or @ravwojdyla would like to take a quick look? I think they were heavily involved in the design of the variable naming/validation system.

ravwojdyla (approved):
This looks great, nice and clean code btw 👍

PR #927, opened 2022-10-06
Remove python 3.7

### LINKED ISSUE #802: Remove support for Python 3.7
[NumPy 1.22](https://numpy.org/devdocs/release/1.22.0-notes.html) no longer supports Python 3.7, so we should consider removing support for it in the next release of sgkit.

This would be consistent with https://numpy.org/neps/nep-0029-deprecation_policy.html, which says Python 3.7 can be dropped from Dec 26, 2021.

tomwhite:
Fixes #802

### DIFF
--- a/.cirun.yml
+++ b/.cirun.yml
@@ -17,7 +17,7 @@ runners:
     # Path of the relevant workflow file
     workflow: .github/workflows/build-gpu.yml
     # Number of runners to provision on every trigger on Actions job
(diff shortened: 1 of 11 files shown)


jeromekelleher (approved):
LGTM

I've opened https://github.com/conda-forge/sgkit-feedstock/issues/4 to track dropping 3.7 from the conda package also.

Issue #925, opened 2022-10-06
maximise lossless compression of vcf_to_zarr


elswob:
Moving and storing large VCF files is a problem. One option is to transform them into a more compact data format, which contains enough data to fully restore the VCF at a later date (see https://github.com/pystatgen/sgkit/issues/924). The default vcf_to_zarr reduces the size of the data significantly, however, when extra fields are included, e.g. `["INFO/*", "FORMAT/*"]` the resulting zarr structure can be larger than the original VCF.

tomwhite:
BTW there's also a summary on https://github.com/tomwhite/ga4gh-variant-comparison

jeromekelleher:
Have you got an example where the we don't do so well on compression @elswob? I'm guessing it's some particular INFO or FORMAT field that's not being dealt with well.

elswob:
Hi both, thanks for the comments. Have you tried genozip https://genozip.readthedocs.io/index.html ? Be interested to see how that compares.

I can confirm similar numbers using sgkit for the 1000 genomes data used in the comparison. The file I'm seeing strange values with is a [beagle imputation output VCF](https://faculty.washington.edu/browning/beagle/beagle_5.4_18Mar22.pdf) for around 1 million variants and 100 samples based on 1000 genomes data, around 40 MB. If I run it with the params [here](https://github.com/tomwhite/ga4gh-variant-comparison/blob/main/convert_to_sgzarr.py) I get a 138 MB Zarr directory.

elswob:
```
128M	call_DS
1.2M	call_genotype
144K	call_genotype_mask
140K	call_genotype_phased
 12K	sample_id
3.5M	variant_AF
920K	variant_DR2
 76K	variant_IMP
2.7M	variant_allele
 76K	variant_contig
 76K	variant_id
 76K	variant_id_mask
1.2M	variant_position
```

jeromekelleher:
There we go - I bet we're storing ``call_DS`` as a 32 or 64 bit float which isn't compressing very well.

tomwhite:
> I bet we're storing `call_DS` as a 32 or 64 bit float which isn't compressing very well.

It will be a 32-bit float if it's a VCF float.

[...]

PR #937, opened 2022-10-18
Improve default VCF compression

tomwhite:
This implements some of the findings in #925

* Adds Delta encoding (filter) for VCF POS fields
* Issue a warning if there are float FORMAT fields with no filters set
* Adds a section of documention to the VCF covering compression, and giving an example of how to truncate float fields to 2dp. 
* 0efd28a7826aac6b14f74bedd41ab880338eb8f2 is a fix for a bug where if you specify the filters for a dataset variable then the default sgkit compressor is ignored since the dictionaries weren't being merged correctly.

I considered always rounding FORMAT float fields to 2dp, but there are problems with NaN (can't use FixedScaleOffset since it doesn't preserve NaNs) and precision (which integer type to use for FixedScaleOffset depends on the range of the values).

Quantize and Bitround do better here (both support NaN), but they are not as compact as FixedScaleOffset. It's probably worth doing more investigation into them (another PR?), but so far in this PR I have just added a warning that suggests setting filters, which is a good start.

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -5,17 +5,7 @@ import warnings
 from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path
-from typing import (
-    Any,
-    Dict,
-    Hashable,
-    Iterator,
-    MutableMapping,
-    Optional,
-    Sequence,
-    Tuple,
-    Union,
-)
+from typing import Any, Dict, Iterator, MutableMapping, Optional, Sequence, Tuple, Union
 
 import dask
 import fsspec
@@ -23,7 +13,6 @@ import numpy as np
 import xarray as xr
 import zarr
 from cyvcf2 import VCF, Variant
-from numcodecs import PackBits
 
 from sgkit import variables
 from sgkit.io.utils import (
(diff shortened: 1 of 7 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -529,34 +531,33 @@ def vcf_to_zarr_sequential(
...
+                    # Issue warning for VCF FORMAT float fields with no filter
+                    if (
+                        var.startswith("call_")
+                        and ds[var].dtype == np.float32
+                        and (
+                            var not in merged_encoding
+                            or "filters" not in merged_encoding[var]
+                        )
+                    ):
+                        warnings.warn(
```
benjeffery, line 556: Might be nice to have an explicit test for the warning with `pytest.warns`?

benjeffery (approved):
Nice! One question about testing. 

A thought - you could add a generic `asv` benchmark for compression of some example VCFs:
https://asv.readthedocs.io/en/stable/writing_benchmarks.html#tracking-generic

jeromekelleher (approved):
LGTM

tomwhite:
> A thought - you could add a generic `asv` benchmark for compression ratio of some example VCFs:
> https://asv.readthedocs.io/en/stable/writing_benchmarks.html#tracking-generic

I opened #938 for this.

Issue #938, opened 2022-10-18
Add `asv` benchmark for compression ratio of some example VCFs


tomwhite:
Suggested by @benjeffery in https://github.com/pystatgen/sgkit/pull/937#pullrequestreview-1145718679

tomwhite:
@hammer notes that https://github.com/ionelmc/pytest-benchmark or https://github.com/conbench/conbench are alternatives to ASV.

hammer:
It definitely sounds like `pytest-benchmark` is the simpler starting point.

hammer:
Nvidia made a tool to transform `pytest-benchmark` output to a format expected by `asv` for visualization: https://github.com/rapidsai/asvdb

tomwhite:
It seems that `pytest-benchmark` doesn't support custom metrics that aren't timing-related, i.e. compression ratio in our case here.

https://github.com/coiled/coiled-runtime/issues/186 is worth a read.

tomwhite:
Fixed in #978

PR #941, opened 2022-10-25
VCF compression size test

tomwhite:
This uses data from #925 in a test that asserts that the Zarr size is less than 75% of the gzipped VCF. It could be made into a benchmark (`asv` or otherwise, see #938), but for the moment it was simplest to write it as a regular unit test.
 
When writing this I found a bug/limitation (f7d44a8ebd6bee7d3660c94a01d3116fd17ddc0d) where the Zarr chunk size in the samples dimension is set to 1000 (the default), and not limited to the number of samples. This would be potentially wasteful and slow for VCFs with just a few samples. @benjeffery I wonder if this change helps speed up your TB VCF import?


jeromekelleher:
Hmm, 17Mb is very big to add to the repo. Presumably this unit test takes a perceivable amount of time also - I wonder if this is the right format format for this benchmark?

tomwhite:
> Hmm, 17Mb is very big to add to the repo. Presumably this unit test takes a perceivable amount of time also - I wonder if this is the right format format for this benchmark?

You're right - maybe let's not merge this then, but use it as the basis for a benchmark when we've decided how to run them. (BTW we already have a 13MB test file in the repo, but good not to add too many of these.)

I'll move the chunk fix into its own PR.

tomwhite:
Closed in favour of #978

PR #942, opened 2022-10-25
Test for VCF lossless storage when using encoding

tomwhite:
Uses a snippet of data from #925 to check that using filters still results in lossless storage.


jeromekelleher (approved):
LGTM

PR #943, opened 2022-10-25
Ensure chunking is not excessive in samples dimension.

tomwhite:
From #941.

For VCFs with a small number of samples, a Zarr chunk size of 1000 was being used by default which was wasteful and slow. This change limits the chunk size to the number of samples.


jeromekelleher (approved):
LGTM

Issue #936, opened 2022-10-17
Optimise VCF import


tomwhite:
As a complement to #924 it would be good to measure and profile VCF import performance. What speed can we expect?

@benjeffery has already found some nice optimisations in #932 and #933.

We already have an [ASV instance](https://pystatgen.github.io/sgkit-benchmarks-asv/) that we could use to track changes.

jeromekelleher:
"Fast enough" is hard to quantify here all right. Within a factor of 10 of the speed that htslib can chew through the file would be great, but perhaps that's too hard a target.

Issue #924, opened 2022-10-06
VCF export


jeromekelleher:
VCF is the bottleneck that essentially all variant data goes through these days, and people will need to export data that they have worked with in sgkit to VCF so that it can be processed with other tools. Not having a solid way of exporting data to a widely used format is a significant missing piece of functionality.

I see three ways we can go about exporting to VCF:

1. Use htslib via a python library that supports writing VCF
2. Write the VCF text directly ourselves using numba
3. Use some numpy tricks to write the text by manipulating numpy arrays.

Both 2 and 3 have the disadvantage that we're writing out uncompressed text, and in practise the user would need to filter this through bgzip for the output to be useful. AFAIK there's no Python interface to bgzip (but, I haven't looked). I don't think this is a show-stopper, though.

I haven't tried using Python VCF writing libraries for a while, but previous experience has shown that they aren't great. I don't think cyvcf2 really supports it, and the only other one I know is PySAM (which I don't think we want another dependency on, and I doubt would scale in the way we want).

Option 2 could be good, but I'm not sure how adapt numba is at generating text. Presumably we could work with numpy int8 arrays which would do the equivalent things and we should be able to get reasonable performance from it.

Option 3 is what we've used in tskit, which gives quite respectable performance (100s of megabytes per second for large numbers of samples). See the code [here](https://github.com/tskit-dev/tskit/blob/a709954e9e4a941850ac9b0d431df4360978d52f/python/tskit/vcf.py#L256). The trick is to pre-allocate the arrays of text, manipulate the bits that change for each row using numpy operations, and then convert the buffer to a string just before outputting.

This works well for tskit, but it's a much more constrained space of VCFs that need to be dealt with (no GL cols, etc) so I'm not sure it would work here (or at least, it wouldn't get excessively complicated).

Any thoughts @tomwhite?

tomwhite:
cyvcf2 does support [writing VCF](https://brentp.github.io/cyvcf2/docstrings.html#cyvcf2.cyvcf2.Writer), including bgzip. 

There is actually already support for [converting sgkit datasets to VCF](https://github.com/pystatgen/sgkit/blob/6554e54d5808ba975d73d84c93976ba807076f3d/sgkit/tests/io/vcf/vcf_writer.py#L131-L201 ), which I wrote a while back for testing purposes. This is Option 1 in your categorisation. (It doesn't do compression, but that would be easy to add.) It works pretty well as the test covers lots of cases for round-tripping VCF.

[...]

PR #944, opened 2022-10-25
VCF benchmarks

tomwhite:
This is not ready for merging, but shows the current status of VCF read and write speed (#936 and #924).

```
sgkit/tests/io/vcf/test_vcf_benchmark.py::test_vcf_read_speed bytes read: 78683380
duration: 2.69 s
speed: 29.3 MB/s
PASSED
sgkit/tests/io/vcf/test_vcf_benchmark.py::test_vcf_write_speed [W::vcf_parse_format] Extreme FORMAT/AD value encountered and set to missing at 20:10000054
bytes written: 79602637
duration: 99.27 s
speed: 0.8 MB/s
PASSED
```

It should be possible to get both to ~100MB/s.

### DIFF
--- a/sgkit/tests/io/vcf/test_vcf_benchmark.py
+++ b/sgkit/tests/io/vcf/test_vcf_benchmark.py
@@ -4,6 +4,7 @@ import shutil
 import time
 
 from sgkit.io.vcf.vcf_reader import vcf_to_zarr
+from sgkit.io.vcf.vcf_writer import zarr_to_vcf
 from sgkit.tests.io.vcf.utils import path_for_test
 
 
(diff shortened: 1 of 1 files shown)


tomwhite:
Re-running with the new code from #953, I get:

```
sgkit/tests/io/vcf/test_vcf_benchmark.py::test_vcf_write_speed bytes written: 81811189
duration: 0.84 s
speed: 97.7 MB/s
PASSED
```

This could be merged given that it only takes a few seconds to run now. It could be added to benchmark tests later (#938).

jeromekelleher:
Do we want to put this in as a unit test? Maybe simpler to use have a ``vcf_benchmarks.py`` file in the root or something that you can run if you like? I don't really see the advantage of running this under pytest, since we also need benchmarks that definitely wouldn't run in seconds.

tomwhite:
Yes, that's probably the best thing to do.

tomwhite:
I've opened #976 as a replacement for this.

tomwhite:
Closed in favour of #976

PR #945, opened 2022-11-02
Add `window_by_genome`

### LINKED ISSUE #893: Optional genomic-scale for popgen stats
See discussion in #888

Most methods within `popgen.py` are automatically applied over windows and default to variants if the dataset has not been windowed. It would be useful to optionally apply these methods over the whole genome and potentially per contig. Additionally, some methods such as `identity_by_state` and `Weir_Goudet_beta` are currently applied to the whole genome, but it would also be useful to apply them within windows/contigs.

An option to allow more user control, whist keeping sensible defaults, would be to introduce a `by` parameter which can be used to specify the genomic scale (e.g., one of `{"genome", "windows", "variants"}`). The specified scale would determine the genomic-dimension and shape of the resulting values.

tomwhite:
Fixes #893

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -118,6 +118,7 @@ Utilities
    infer_variant_ploidy
    parent_indices
    simulate_genotype_call_dataset
+   window_by_genome
    window_by_position
    window_by_variant
 
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -214,7 +214,9 @@ are defined then the values will typically be computed for each variant. It is t
 important to define windows *before* computing statistics on a dataset.
 
 Windows are intervals that span the ``variants`` dimension in a dataset, and they are defined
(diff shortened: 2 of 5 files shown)


tomwhite:
> How do you handle release changelogs in sgkit? Usually with a PR like this I'd expect an update to a changelog - or do you create it retrospectively on release?

The release page has a link to changes (see e.g. https://github.com/pystatgen/sgkit/releases/tag/0.5.0), but that was before GitHub had the ability to generate release notes. I'd be interested to hear suggestions on how to improve this.

benjeffery:
I've never found automated changelogs to be great for release notes as you get a lot of technical changes that are not relevant or visible to end users. Over at tskit we have a changelog file that gets updated (if needed) with each PR. Our release automation then pushes that to the github release description, and the changelog is compiled into the docs.

timothymillar (approved):
LGTM!

PR #952, opened 2022-11-15
Disable `count_a1=True` in `read_plink`

### LINKED ISSUE #947: potential issue in how `ds.variant_allele` is populated in `read_plink`
Hi all! I'd greatly appreciate some clarification if the following test case shows intended behavior, or if there might be an issue with how alleles in `ds.variant_allele` are associated with their corresponding count data when reading data from plink:

We'll set up two SNPs, one A/G SNP with minor allele A at MAF=0.3, and one T/C SNP with minor allele T at MAF=0.4.
`test.ped`:
```
ind0 ind0 0 0 0 0 A A T T
ind1 ind1 0 0 0 0 A A T T
ind2 ind2 0 0 0 0 A A T T
ind3 ind3 0 0 0 0 G G T T
ind4 ind4 0 0 0 0 G G C C
ind5 ind5 0 0 0 0 G G C C
ind6 ind6 0 0 0 0 G G C C
ind7 ind7 0 0 0 0 G G C C
ind8 ind8 0 0 0 0 G G C C
ind9 ind9 0 0 0 0 G G C C
```

`test.map` (dummy data needed for conversion):
```
1 1_10 0 10
1 1_20 0 20
```
 and covert it to plink binary format:
``` bash
plink --file test --make-bed --out test
```

running this through `read_plink` with default parameters shows that the alleles are incorrectly associated with the count data:
``` python
import sgkit as sg
from sgkit.io.plink import read_plink

ds = read_plink(path="test")
ds = sg.call_allele_frequencies(ds)

print(ds.call_allele_frequency.mean(dim="samples").values)
print(ds.variant_allele.values)
```
```
[[0.7 0.3]
 [0.6 0.4]]
[[b'A' b'G']
 [b'T' b'C']]
```
i.e. A and T are incorrectly identified as the major, rather than the minor alleles for our two SNPs:
``` python
print(ds.isel(alleles=0).call_allele_frequency.mean(dim="samples").values)
print(ds.isel(alleles=0).variant_allele.values)
```
```
[0.7 0.6]
[b'A' b'T']
```

However, `read_plink` has the option `count_a1` which defaults to `True`. Setting it to `False` correctly associates the alleles with their counts:
``` python
import sgkit as sg
from sgkit.io.plink import read_plink

ds = read_plink(path="test", count_a1=False)
ds = sg.call_allele_frequencies(ds)

print(ds.call_allele_frequency.mean(dim="samples").values)
print(ds.variant_allele.values)
```
```
[[0.3 0.7]
 [0.4 0.6]]
[[b'A' b'G']
 [b'T' b'C']]
```

My issue with this is that the doc string for `count_a1` says:
```
        Whether or not allele counts should be for A1 or A2,
        by default True. Typically A1 is the minor allele
        and should be counted instead of A2. This is not enforced
        by PLINK though and it is up to the data generating process
        to ensure that A1 is in fact an alternate/minor/effect
        allele.
```
Inspecting the relevant `.bim` file shows that A1 is in fact the minor allele:
`test.bim`:
```
1	1_10	0	10	A	G
1	1_20	0	20	T	C
```
So I think we should be in the regime that's supposed to be covered by the default parameter, but isn't.

tomwhite:
Fixes #947


jeromekelleher (approved):
LGTM

jeromekelleher:
Should probably document this in CHANGELOG so we don't forget?

jeromekelleher:
Would you mind taking a quick look please here to see if you agree with the approach @clwgg?

clwgg:
I think the changes look great! My only comment would be on the help message of the `count_a1` argument, where it says that:
```
Typically A1 is the minor allele and should be counted instead of A2.
```
I think this may lead to confusion, as the default (and now only) option is setting it to `False`, but the help message appears to state that it "typically" should be `True`.

tomwhite:
Thanks @clwgg - I missed that! I've updated the wording in the latest commit. Does that look OK or is there more explanation that would help in the doc?

clwgg:
I think that's fine! :+1:

PR #962, opened 2022-11-22
Optionally initialise pedigree kinship estimates with known founder kinships

### LINKED ISSUE #961: Initialize pedigree kinship estimation with marker based estimates among founders
Following [Goudet et al. (2018)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6220858/) it'd be useful to have the option of initializing pedigree based kinship estimates with marker estimates among the pedigree founders. This is a simple way of mitigating cryptic relatedness among founders when marker data is available for founders, but not for other samples.

timothymillar:
Fixes #961

I'm still looking into options for comparing with external data/software but I think `test_pedigree_kinship__projection` is a reasonably effective approach.

I don't like requiring two additional arguments but I think this API strikes the balance between flexible and safe. Suggestions to improve on it are welcome!


sgkit/stats/pedigree.py
```diff
@@ -203,6 +203,23 @@ def _raise_on_half_founder(
             raise ValueError("Pedigree contains half-founders")
 
 
+@numba_jit
+def _identify_founders(
+    parent: ArrayLike, tau: ArrayLike = None
```
jeromekelleher, line 208: I wonder if doing the control flow like this confuses numba jitting, as it's going to compile two versions here (one for tau=None and one for tau 2D array) I'd imagine.

Since it's such a simple function, perhaps it would be as well to do two versions, taking the control flow out of the tight inner loop?

sgkit/stats/pedigree.py
```diff
@@ -293,6 +310,94 @@ def kinship_diploid(
...
 
 
+@numba_jit
+def project_kinship_diploid(
+    parent: ArrayLike,
+    founder_kinship: ArrayLike,
+    founder_indices: ArrayLike,
+    allow_half_founders: bool = False,
+) -> ArrayLike:  # pragma: no cover
+    """Project founder kinships down a pedigree assuming all individuals
```
jeromekelleher, line 320: "Down" can confuse people as people visualise trees/pedigrees in opposite directions (phylogenetics people tend to think of "down" as backwards in time, putting the root at the bottom).

Can you rephrase in terms of ancestral/descendant language?

jeromekelleher (approved):
LGTM, couple of small points.

sgkit/stats/pedigree.py
```diff
@@ -203,6 +203,23 @@ def _raise_on_half_founder(
             raise ValueError("Pedigree contains half-founders")
 
 
+@numba_jit
+def _identify_founders(
+    parent: ArrayLike, tau: ArrayLike = None
+) -> ArrayLike:  # pragma: no cover
```
tomwhite, line 209: BTW you should be able to omit the no cover pragma now that we are testing coverage of numba-jitted functions (#964).

sgkit/stats/pedigree.py
timothymillar: Fair call, it could even be inlined completely but it seems likely that the logic will be reused at some point

timothymillar:
Looks like I have some coverage gaps ☹️

sgkit/stats/pedigree.py
timothymillar: 🥳... ☹️

tomwhite: Thanks for fixing!

jeromekelleher:
Are we good to go here @timothymillar?

sgkit/stats/pedigree.py
```diff
@@ -294,17 +290,113 @@ def kinship_diploid(
...
+    Any non-founders included in the founder_kinship and founder_indices arrays
+    will be treated as founders by ignoring their parentage.
+
+    Returns
+    -------
+    kinship
+        A square matrix of kinship estimates with self-kinship values on
+        the diagonal.
+
+    Raises
```
benjeffery, line 341: I'm not sure how exact the docs need to be here, but this list is missing the AssertionErrors that can be raised by 354 and 355. (Same for the other function below)

benjeffery (approved):
LGTM! One question.

sgkit/stats/pedigree.py
timothymillar: Good question, those assertions should never be raised via the public API (`pedigree_kinship`) because the dimensions are checked here: https://github.com/pystatgen/sgkit/blob/b966a1a76b486d5aa5ddc162d9b631d36c210e4e/sgkit/stats/pedigree.py#L1037. That check itself is also not strictly necessary as the dimensions are checked automatically against the `da.gufunc` signature here https://github.com/pystatgen/sgkit/blob/b966a1a76b486d5aa5ddc162d9b631d36c210e4e/sgkit/stats/pedigree.py#L1057 but this results in an error that may be unclear to the user (`ValueError: Dimension 'f' with different lengths in arrays`) hence the custom error message (or maybe I'm being overly cautious?). The assertions are really just there as a safety net for future development.

timothymillar:
@jeromekelleher good to go if @benjeffery is happy with my explanation for the use of assertions

sgkit/stats/pedigree.py
benjeffery: Ah, nice! Thanks for the explanation. I think it is good to catch more obscure error messages as you've done here.

benjeffery:
> @jeromekelleher good to go if @benjeffery is happy with my explanation for the use of assertions

Yep, I'm more than happy!

PR #968, opened 2022-11-28
Add `assert_vcfs_close` for VCF files

tomwhite:
From https://github.com/pystatgen/sgkit/pull/953#issuecomment-1317184549:

> Trying to get exact float matching is a fool's game in my experience (rendering binary floats in base 10 is fundamentally lossy), if we're equal to within a small epsilon that's all that matters.

This change adds a `assert_allclose` test function (like the one in NumPy and Xarray) to test that two VCF files are equal - to within a small epsilon for float fields.

With this test the "all fields" generated VCF passes the unit test using the new numba VCF writer from #953.


benjeffery (approved):
LGTM, nice!

sgkit/tests/io/vcf/utils.py
```diff
@@ -1,12 +1,103 @@
...
     """Return a test data path whose type is determined by `is_path`.
 
-    If `isPath` is True, return a `Path`, otherwise return a `str`.
+    If `is_path` is True, return a `Path`, otherwise return a `str`.
     """
     path: PathType = shared_datadir / file
     return path if is_path else str(path)
+
+
+def assert_allclose(f1, f2, *, rtol=1e-05, atol=1e-03):
```
jeromekelleher, line 19: Is this going to clash with the numpy.testing.assert_allclose though? I can imagine both functions being used in the same file and the ``from module import function`` style makes it hard to know where particular functions come from. How about ``assert_vcfs_close``?

jeromekelleher (approved):
LGTM - suggested name change though?

sgkit/tests/io/vcf/utils.py
tomwhite: I wondered about this - I agree this makes things less ambiguous and clash-prone.

Issue #790, opened 2022-01-05
Can't make equally spaced overlapping windows


timothymillar:
I can't seem to do this with `window_by_position`. 

I looks like the  `window_start_position` argument implicitly expects the specified positions match the contigs of the `variant_contig` array. The contig bounds are calculated [here](https://github.com/pystatgen/sgkit/blob/main/sgkit/window.py#L229) from the `variant_contig` array. The bounds of a given contig (start and stop) are the indices corresponding to the first and last variants on the contig. These bounds are then used [here](https://github.com/pystatgen/sgkit/blob/main/sgkit/window.py#L291) to slice the provided `window_start_positions` and therefore assume that `window_start_positions` match the contigs of `variant_contig`. This assumption is true if  you use `window_start_positions=variant_position` but unlikely to be true for a custom array.

There doesn't seem to be a way to make overlapping windows without using the  `window_start_position` argument.

tomwhite:
You are right @timothymillar - in the current implementation `window_start_position` has to be of size `variants`. This could be made more flexible though and probably should be to accommodate your use case. (Another generalisation that would be useful would be to have `window_stop_positions` to make it possible to define any number of variable-sized windows.)

Is this something you could look at implementing? I have less time this year on sgkit, so can't promise how soon I could get to it, but would be happy to review.

timothymillar:
Thanks @tomwhite. I'm currently quite busy with my thesis but I can have a look when I have some time. It's fairly straightforward to generate overlapping windows on an an ad-hoc basis anyway.

Yes, window_stop_positions would be useful to generate windows from BED/GFF files.

tomwhite:
Thinking about this more, this case can easily be supported by added a `step` argument (like the one in `window_by_variant`). See #793. How does this look @timothymillar?

timothymillar:
That looks great, thanks @tomwhite !

Issue #692, opened 2021-09-30
Implement gene set association analysis


eric-czech:
We have methods for associating individual genetic variants and phenotypes (https://github.com/pystatgen/sgkit/pull/16, https://github.com/pystatgen/sgkit/pull/66), but inferring associations with gene sets and pathways would also be very useful.

In principle, these methods are fairly simple and nearly always appeal to a multivariate regression model of the individual SNPs for a gene set against a phenotype (or approximations of this based on gene-level or summary statistic data).  Correlation between the SNPs and the samples, rare occurrences of variants and phenotypes, gene length and pathway size are all complicating factors though that led to a fairly long history of ever-improving methods to correct for these things.

Methods in this space support computing gene set statistics from individual phenotypes, variant-level sumstats, gene-level sumstats or some combination of those.  For the sumstats cases, some reference LD panel is usually required.  

Here are some notes from @hammer on approaches to this problem:

**Old Methods**
- [Genetic analysis of biological pathway data through genomic randomization](https://pubmed.ncbi.nlm.nih.gov/21279722/) (2011): PARIS
- [Optimal unified approach for rare-variant association testing with application to small-sample case-control whole-exome sequencing studies](https://pubmed.ncbi.nlm.nih.gov/22863193/) (2012): SKAT-O
- [MAGMA: generalized gene-set analysis of GWAS data](https://pubmed.ncbi.nlm.nih.gov/25885710/) (2015)
- [Gene-set association tests for next-generation sequencing data](https://pubmed.ncbi.nlm.nih.gov/27587681/) (2016): QTest

**New Methods**
- [A Bayesian hierarchical variable selection prior for pathway-based GWAS using summary statistics](https://pubmed.ncbi.nlm.nih.gov/31777110/) (2020)
- [Explaining the Genetic Causality for Complex Phenotype via Deep Association Kernel Learning](https://pubmed.ncbi.nlm.nih.gov/33205126/) (2020): DAK
- [A powerful subset-based gene-set analysis method identifies novel associations and improves interpretation in UK Biobank](https://www.biorxiv.org/content/10.1101/799791v2) (2020): GAUSS
- [Estimation of non-null SNP effect size distributions enables the detection of enriched genes underlying complex traits](https://pubmed.ncbi.nlm.nih.gov/32542026/) (2020): gene-ε
[...]

PR #974, opened 2022-12-05
`window_by_interval`

tomwhite:
Adds a function to create windows from arbitrary intervals, which may come from interval lists or BED files, for example. It's left to the user to load from these files (though there's an example for BED in the tests), mainly to avoid having to handle genomic coordinate conventions in sgkit itself (#434).

This provides the useful minimum needed for gene-ε (#692), where some of the examples have gene lists that I would like to convert to sgkit windows.

@timothymillar you may be interested in this, since it implements a case you brought up in [#790](https://github.com/pystatgen/sgkit/issues/790#issuecomment-1006298165).


timothymillar (approved):
LGTM, I think leaving the IO to users is the best approach for simple tabular formats.

Just out of curiosity, was there a particular reason that contig names are stored in an attribute of the dataset rather than in an array (e.g., `contig_id`)?

jeromekelleher (approved):
LGTM

benjeffery (approved):
LGTM!

tomwhite:
> Just out of curiosity, was there a particular reason that contig names are stored in an attribute of the dataset rather than in an array (e.g., `contig_id`)?

I looked back and found that contig names were introduced in #10, and there doesn't seem to have been any particular reason that they are stored as an attribute. Sample IDs are stored as a variable, for example. I guess we could change this, but might need to think about backward compatibility.

timothymillar:
> I guess we could change this, but might need to think about backward compatibility.

I don't think it's a big issue, just struck me as a little inconsistent.

PR #976, opened 2022-12-12
Add asv benchmarks for VCF

tomwhite:
This is an [`asv`](https://asv.readthedocs.io/) version of #944.

We discussed moving away from `asv` in #938, but alternatives like `pytest-benchmark` don't allow metrics that aren't time based. Given that we already publish `asv` benchmarks at https://pystatgen.github.io/sgkit-benchmarks-asv/, I had a go at re-purposing the VCF speed benchmarks in #944 as `asv` benchmarks. This wasn't too hard, and it's possible to run them locally if needed using `asv run` from the `benchmarks` directory.

### DIFF
--- a/benchmarks/benchmarks_vcf.py
+++ b/benchmarks/benchmarks_vcf.py
@@ -1,0 +1,78 @@
+"""Benchmark suite for VCF module."""
+import gzip
+import os
+import shutil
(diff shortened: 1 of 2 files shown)


jeromekelleher (approved):
LGTM

PR #978, opened 2022-12-14
Add asv benchmarks for VCF compression size

tomwhite:
This is an `asv` version of #941, which it supersedes.

I've reduced the size of the test file to a more manageable size of 5.5MB (500k lines of vcf).


jeromekelleher (approved):
LGTM

PR #983, opened 2023-01-03
Process input in chunks in VCF writer

### LINKED ISSUE #965: VCF writer should process input in chunks
The implementation in #953 loads the entire dataset into memory using `dataset.load()`. Instead it should read a chunk at a time so it's possible to stream through a large zarr file/dataset. 

Part of #924

tomwhite:
Fixes #965

### DIFF
--- a/sgkit/io/vcf/vcf_writer.py
+++ b/sgkit/io/vcf/vcf_writer.py
@@ -38,9 +38,6 @@ def dataset_to_vcf(
     containing the VCF header. VCF files converted to Zarr using :func:`vcf_to_zarr`
     will contain this attribute, but datasets loaded from other sources will not.
 
-    The current implementation loads the entire dataset into memory, so it should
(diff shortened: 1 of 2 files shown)


sgkit/io/vcf/vcf_writer.py
```diff
@@ -49,220 +46,227 @@ def dataset_to_vcf(
...
+            dataset_chunk_to_vcf(
+                ds, header_info_fields, header_format_fields, contigs, filters, output
+            )
 
-        # format fields
 
-        # these can have different sizes for different fields, so store in sequences
-        format_values = []
-        format_bufs = []
+def dataset_chunk_to_vcf(
```
tomwhite, line 72: This looks like a bigger change than it actually is, as it's basically just an indentation change.

benjeffery (approved):
Nice, LGTM!

PR #988, opened 2023-01-06
Add auto_rechunk to save_dataset

### LINKED ISSUE #981: Add rechunking to `save_dataset`
I'm creating a workflow where I need to subset the variants in a dataset and write it back to zarr. However:
```python
ds = sg.simulate_genotype_call_dataset(50000,10)
sg.save_dataset('test.zarr')
ds = sg.load_dataset('test.zarr')
selection = xr.ones_like(ds['variant_position'], dtype=bool)
selection[1000:1100] = False
ds = ds.sel(variants=selection)
sg.save_dataset('test.zarr')
```
Results in: `ValueError: Final chunk of Zarr array must be the same size or smaller than the first.` 
Or with more chunks: `ValueError: Zarr requires uniform chunk sizes except for final chunk.`

I could rechunk all the arrays with a variants dimension - this doesn't make for an easy time for the average user though.
Should `save_dataset` auto rechunk in these situations?

benjeffery:
Fixes #981

### DIFF
--- a/sgkit/io/dataset.py
+++ b/sgkit/io/dataset.py
@@ -13,6 +13,7 @@ def save_dataset(
     ds: Dataset,
     store: Union[PathType, MutableMapping[str, bytes]],
     storage_options: Optional[Dict[str, str]] = None,
+    auto_rechunk: Optional[bool] = None,
     **kwargs: Any
 ) -> None:
     """Save a dataset to Zarr storage.
@@ -28,6 +29,9 @@ def save_dataset(
         Zarr store or path to directory in file system to save to.
     storage_options:
         Any additional parameters for the storage backend (see ``fsspec.open``).
+    auto_rechunk:
+        If True, automatically rechunk the dataset to uniform chunks before saving,
+        if necessary. This is required for Zarr, but can be expensive. Defaults to False.
     kwargs
         Additional arguments to pass to :meth:`xarray.Dataset.to_zarr`.
     """
@@ -36,6 +40,8 @@ def save_dataset(
         store = fsspec.get_mapper(store, **storage_options)
     elif isinstance(store, Path):
         store = str(store)
+    if auto_rechunk is None:
+        auto_rechunk = False
     for v in ds:
         # Workaround for https://github.com/pydata/xarray/issues/4380
         ds[v].encoding.pop("chunks", None)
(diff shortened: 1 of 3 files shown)


sgkit/io/dataset.py
```diff
@@ -13,6 +13,7 @@ def save_dataset(
     ds: Dataset,
     store: Union[PathType, MutableMapping[str, bytes]],
     storage_options: Optional[Dict[str, str]] = None,
+    auto_rechunk: bool = False,
```
jeromekelleher, line 16: Safer to make the default ``None`` here in case we want to do something clever in future?

sgkit/io/dataset.py
```diff
@@ -48,7 +52,34 @@ def save_dataset(
...
+            # intended chunk size.
+            ds = ds.chunk(
+                chunks={dim: max(chunks) for dim, chunks in ds.chunks.items()}
+            )
+
+    # Catch unequal chunking errors to provide a more helpful error message
+    try:
+        ds.to_zarr(store, **kwargs)
+    except ValueError as e:
+        if "Zarr requires uniform chunk sizes" in str(
```
jeromekelleher, line 74: This feels a bit brittle - what happens when upstream changes the error messages slightly?

jeromekelleher (commented):
LGTM, but not familiar enough with problems to comment on solution.

sgkit/io/dataset.py
benjeffery: Yes, I agree, but I think its the best of two bad options. The other option is to copy across the detection code from xarray - but then we can't tell if they have changed the logic. At least this way our tests should fail if xarray changes the message. If they do so, and we haven't updated the end user will see the xarray message as before, which isn't awful, just inconvenient.

sgkit/io/dataset.py
benjeffery: I also thought this - but checking the rest of the API `bool` options aren't following this pattern in sgkit.

sgkit/tests/io/test_dataset.py
```diff
@@ -46,3 +46,63 @@ def test_save_and_load_dataset__mutable_mapping():
...
+    with pytest.raises(
+        ValueError, match="Zarr requires uniform chunk sizes. Use the `auto_rechunk`"
+    ):
+        save_dataset(ds, {}, auto_rechunk=False)
+
+    store = {}
+    save_dataset(ds, store, auto_rechunk=True)
+    assert_identical(ds, load_dataset(store))
+
+    # A equal chunked ds retains its original chunking
```
tomwhite, line 101: Nit: "An"

tomwhite (approved):
LGTM (after Jerome's comments have been addressed)

sgkit/tests/io/test_dataset.py
benjeffery: Fixed

sgkit/io/dataset.py
benjeffery: Changed.

benjeffery:
All fixed up, should be good to go.

PR #992, opened 2023-01-10
Use pydata_sphinx_theme and add version switcher

### LINKED ISSUE #669: Version our docs
It will add complexity to our doc build and push but I think we're at the point now where we should support searching different versions of the docs, including the latest version of `main`.

tomwhite:
- [x] Fixes #669 

To use the version switcher, we need to use the [PyData Sphinx Theme](https://pydata-sphinx-theme.readthedocs.io/en/latest/index.html). This is very similar to the Sphinx Book Theme we are using now, so it's not too hard to change it.

I checked that our search plugin still works - it does, and in fact it is more usable as it uses the whole screen:

<details>

Screenshot 2023-01-10 at 15 03 06

</details>

We also get a dark theme option:

<details>

Screenshot 2023-01-10 at 15 02 30

</details>

And finally the version switcher in action:

Screenshot 2023-01-10 at 15 02 08

We should be able to merge this now and it will just work as we already have the document versions deployed (just no way of switching between them). See https://github.com/pystatgen/sgkit/tree/gh-pages


jeromekelleher:
Wow, that looks super pretty, go for it!

benjeffery (approved):
LGTM! Nice, reminds me I need to get the switcher working over at tskit.

PR #995, opened 2023-01-16
Deduplicate dosage variables

### LINKED ISSUE #875: Is there a diference between variables `dosage` and `call_dosage`
Are these duplicate variables? `call_dosage` seems to be more explicit with ndim=2 and the associated `call_dosage_mask` but isn't currently being used. Variable `dosage` doesn't specify dimensions and is used for LD pruning.

timothymillar:
Replaces variable `dosage` with `call_dosage`.

- [x] Fixes #875
- [x] Tests added

This is a bit more of a breaking change than originally anticipated. The main user facing change is the default argument in the `ld.py` methods. I have also had to generalize the `call_dosage` spec to allow ints and uints. I wonder if there's an elegant way to deprecate a variable from the spec?

[`test_regenie_loco_regression__raise_on_non_2D`](https://github.com/pystatgen/sgkit/blob/main/sgkit/tests/test_association.py#L505) is failing because the incorrect dimensionality is now caught early by variable validation. I could just remove the test but the comment  `# we also cover Q.rechunk(X.chunksize) in this test` suggests that a new test would be required. However, I'm not quite confident I understand the code well enough. Could @pentschev or @eric-czech advise?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -159,7 +159,6 @@ By convention, variable names are singular in sgkit. For example, ``genotype_cou
     variables.cohort_allele_count_spec
     variables.cohort_allele_frequency_spec
     variables.covariates_spec
-    variables.dosage_spec
     variables.genotype_count_spec
     variables.interval_contig_name_spec
     variables.interval_start_spec
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -46,6 +46,12 @@ Breaking changes
   defaulted to ``True`` but now defaults to ``False``. Furthermore, ``True``
   is no longer supported since it is not clear how it should behave.
   (:user:`tomwhite`, :pr:`952`, :issue:`947`)
+- The ``dosage`` variable specification has been removed and all references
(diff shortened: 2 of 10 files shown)


tomwhite:
> I wonder if there's an elegant way to deprecate a variable from the spec?

Not sure, but please be sure to add an entry in the change log under "Breaking changes".

timothymillar:
AFAICT [this test](https://github.com/pystatgen/sgkit/blob/main/sgkit/tests/test_association.py#L277) already covers the same code path as the test which was failing. I've updated the failing test to cover only the `Q.rechunk(X.chunksize)` case.

timothymillar:
Looks like the test failure was unrelated (`test_vcf_to_zarr `). Otherwise I think this is good to go but would prefer to get some feedback from one of the original authors.

tomwhite:
> Looks like the test failure was unrelated (`test_vcf_to_zarr `)

This should be fixed in latest main branch (22bb25541dc8ee34f8c6526e632b0e08aca24ec4).

tomwhite:
I think this can be merged now @timothymillar

PR #996, opened 2023-01-16
Document SGKIT_DISABLE_NUMBA_CACHE variable #869

### LINKED ISSUE #869: Disable numba caching via environment variable

timothymillar:
- [x] Fixes #869

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -339,3 +339,24 @@ By passing keyword arguments to ``visualize`` we can see the order tasks will ru
     :align: center
 
 Task order number is shown in circular boxes, colored from red to yellow.
+
+Disabling the Numba cache
+-------------------------
+
+Internally, sgkit uses the `Numba JIT compiler <https://numba.pydata.org/>`_ to accelerate some methods.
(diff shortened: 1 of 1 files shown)


tomwhite (approved):
+1

jeromekelleher (approved):

PR #1001, opened 2023-01-18
Optionally return relationship matrices from pedigree kinship methods

### LINKED ISSUE #994: Generic approach to converting between kinship and relationship estimates
The kinship and (additive) relationship matrices are closely related. This is simply `kinship * ploidy = additive_relationship` for samples of a single ploidy (a bit more complicated in mixed ploidy datasets). In sgkit there are multiple methods for estimating kinship or relationship matrices, but no standard way to convert between them.

Kinship estimators:
- `pedigree_kinship`
- `pca_relate`
- `Weir-Goudet_beta`

Relationship estimators:
- `genomic_relationship`

We also have a method `pedigree_relationship` which converts a kinship matrix to a relationship matrix. This is actually a generic conversion, but it targets the pedigree specific variables within the spec. In retrospect, I think that the `pedigree_relationship` method (in its current form) may have been a mistake as it implies that there should be separate conversion methods for each case. 

We could instead:

- Have generic `kinship_as_relationship` and `relationship_as_kinship` methods. Annoyingly these would have to return generic `stat_relationship` and `stat_kinship` variables.
- Leave it to the user to perform these simple conversions (with suitable documentation).

With regard to the `pedigree_kinship` and `pedigree_relationship` methods, I think it'd be better to combine them into a single method  which (optionally?) returns both the relationship and kinship matrices. My guess is that the relationship matrix will be used more commonly, but the kinship matrix is calculated as an intermediate and we want to avoid kinship -> relationship -> kinship. 
It's a bit awkward calling `pedigree_kinship` and then `pedigree_relationship` in their current state with non-diploid datasets (converting tau to ploidy).

Thoughts?

timothymillar:
- [x] Fixes #994
- [x] Tests added

This makes the changes discussed in #994
- removes `pedigree_relationship` method
- adds `return_relationship` option to `pedigree_kinship`

I have also replaced `pedigree_inverse_relationship` with `pedigree_inverse_kinship` for the sake of a consistent API:
- replaces  `pedigree_inverse_relationship` with `pedigree_inverse_kinship`
- adds `return_relationship` option to `pedigree_inverse_kinship`
- adds `stat_pedigree_inverse_kinship` variable to spec
- relevant tests migrated

My apologies for proposing last-minute API changes. I think this is fairly clean and consistent but I'm happy to hear any objections.

**Edit:** Just noting that the changes in [0b94ccf](https://github.com/pystatgen/sgkit/pull/1001/commits/0b94ccf6412d46cfb729293648da6edbb6817abc) do result in a slight performance loss when calculating the inverse relationship matrix. This is a result of factoring out some steps into the dask task graph (more function calls etc.). Testing on pedigrees from 1K to 10K samples; this seems to fairly consistently be around 10ms.


jeromekelleher (approved):
LGTM

tomwhite (approved):

PR #1002, opened 2023-01-19
Replace `genotype_count` with `variant_genotype_count`

### LINKED ISSUE #911: genotype_count_spec is inconsistent with VCF spec

timothymillar:
- [x] Fixes #911
- [x] Tests added

This introduces a new method `count_variant_genotypes` to count multi-allelic genotypes of any fixed ploidy. The genotype counts are reported in variable `variant_genotype_count` and ordered following the VCF specification. 

The `variant_genotype_count` variable is used in method `hardy_weinberg_test` instead of `genotype_count`. The `genotype_count` variable has been removed from the spec as this was its only use. I am a little bit concerned that any custom generation of the (het, hom1, hom2) ordering (specified for `genotype_count`) will still run against the new version and produce an incorrect result. However, I'd imagine that the incorrect result would be obviously in most situations?


sgkit/stats/aggregation.py
```diff
@@ -285,6 +285,143 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+    for d in range(1, k + 1):
+        gcd = _greatest_common_denominator(r, d)
+        r //= gcd
+        r *= n
+        r //= d // gcd
+        n -= 1
+    return r
+
+
+@numba_jit
```
jeromekelleher, line 312: It would be good to have a handful of standalone tests for these functions to check some boundary cases against some other implementation (or just known results).

sgkit/stats/aggregation.py
```diff
@@ -285,6 +285,143 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+)
+def _count_genotypes(
+    genotypes: ArrayLike, _: ArrayLike, out: ArrayLike
+) -> ArrayLike:  # pragma: no cover
+    # Warning: genotype alleles must be sorted in ascending order!
+    n, k = genotypes.shape
+    out[:] = 0
+    for i in range(n):
+        index = 0
+        compleat = True
```
jeromekelleher, line 335: Typo, ``complete``?

sgkit/stats/aggregation.py
```diff
@@ -285,6 +285,143 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+        for j in range(k):
```
jeromekelleher, line 336: I wonder if this exit early is worth it - we usually expect k to be fairly small and you could special_case `_comb_with_replacement` for a=0 to exit quickly.

Often when compiling code like this the less logic you have the better, even if it seemingly means doing a bit more work 

Would this remove the ordering requirement?

It probably won't make any measurable difference mind, feel free to ignore!

sgkit/stats/aggregation.py
```diff
@@ -285,6 +285,143 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+            out[index] += 1
+
+
+def count_variant_genotypes(
+    ds: Dataset,
+    *,
+    call_genotype: Hashable = variables.call_genotype,
+    merge: bool = True,
+) -> Dataset:
+    """Count the number of calls for each possible genotype at each variant.
```
jeromekelleher, line 352: Can you spell out what you mean by "genotype" here? It means different things to different people, and helpful to be precise about it somewhere.

sgkit/stats/aggregation.py
```diff
@@ -285,6 +285,143 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+        gcd = _greatest_common_denominator(r, d)
+        r //= gcd
+        r *= n
+        r //= d // gcd
+        n -= 1
+    return r
+
+
+@numba_jit
+def _comb_with_replacement(n: int, k: int) -> int:
```
jeromekelleher, line 313: If you want to get into optimisation here, I suspect a small lookup table covering the most common values of n and k here, with a fallback to the general implementation would be very fast.

jeromekelleher (commented):
Looks great, a few minor comments.

Feel free to ignore perf suggestions.

sgkit/stats/aggregation.py
timothymillar: Hope it's clearer now

timothymillar:
Thanks for the suggestions @jeromekelleher. With respect to performance, I've gone with an optimized path of the biallelic case. This sees over an order of magnitude speed up relative to the previous version. Also got a small improvement for multi-allelic case.

Some simple numbers for 100k variants, 500 samples:

| Ploidy | N-Alleles |  new (4b9e7eb) | old (0d6e852) | `count_genotypes` |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 2  | 2  | 163 ms | 1.87 s | 4.01 s  |
| 4  | 2  | 250 ms | 4.68 s | NA |
| 2  | 3  | 2.37 s | 2.76 s | NA |
| 4  | 3  | 5.41 s | 5.89 s | NA |

Note the `count_genotypes` function is only counting equivalent things for the diploid biallelic case.

sgkit/stats/aggregation.py
timothymillar: Have split into two functions to avoid this, slight perf improvement. The ordering requirement is unavoidable AFAICT , it's described that way in the VCF spec.

sgkit/stats/aggregation.py
timothymillar: I couldn't quite get it working efficiently, maybe how I'm storing the table? Have specialized for the biallelic case instead which is a massive improvement (see below).

sgkit/stats/aggregation.py
```diff
@@ -285,6 +286,181 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+        See :ref:`dataset_merge` for more details.
+
+    Returns
+    -------
+    A dataset containing :data:`sgkit.variables.variant_genotype_count_spec`
+    of genotype counts with shape (variants, genotypes).
+
+    Warnings
+    --------
+    This method is not suitable for mixed-ploidy genotype calls. If it is
```
benjeffery, line 408: Has there been discussion of whether this should be a hard error? I can imagine cases where a user hasn't realised this.

benjeffery (approved):
LGTM, one question about erroring out on mixed ploidy.

sgkit/stats/aggregation.py
eric-czech: Couldn't the lookup just be a global array with say values of `n` in the rows and values of `k` in the columns with values equal to whatever `_comb` outputs there?  I'm curious what happened when you tried.

Given that `_comb` is O(k log(k)) it's probably not a big deal but I could see that possibly making a meaningful difference with high ploidy data.

sgkit/stats/aggregation.py
timothymillar: Good call

> Does it make sense to raise an error from the numba functions if you hit a call that is < -1?

In the `call_genotype` spec we state "... or -2 to indicate a non allele in mixed ploidy datasets" which is a little ambiguous with respect to the non-mixed case. But probably best raise on < -1 for safety.

docs/changelog.rst
```diff
@@ -46,6 +49,11 @@ Breaking changes
   defaulted to ``True`` but now defaults to ``False``. Furthermore, ``True``
   is no longer supported since it is not clear how it should behave.
   (:user:`tomwhite`, :pr:`952`, :issue:`947`)
+- The ``genotype_count`` variable has been removed in favour of
+  :data:`sgkit.variables.variant_genotype_count_spec` which follows VCF ordering.
```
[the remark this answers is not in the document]
timothymillar: Also added to spec description

sgkit/stats/aggregation.py
```diff
@@ -285,6 +286,181 @@ def call_rate(ds: Dataset, dim: Dimension, call_genotype_mask: Hashable) -> Data
...
+    merge
+        If True (the default), merge the input dataset and the computed
+        output variables into a single dataset, otherwise return only
+        the computed output variables.
+        See :ref:`dataset_merge` for more details.
+
+    Returns
+    -------
+    A dataset containing :data:`sgkit.variables.variant_genotype_count_spec`
+    of genotype counts with shape (variants, genotypes).
```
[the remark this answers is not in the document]
timothymillar: Will document with examples in the spec description

sgkit/stats/aggregation.py
timothymillar: Thanks, now seeing 3-5x speedup for multi-allelic cases tested above. Don't know what I was doing last time 🤷.

timothymillar:
Thanks for the feedback @benjeffery and @eric-czech.

> add a coordinate to the genotypes count array

I like this idea although I'm not sure that we are relying on coordinates elsewhere as they seem to be largely optional? There's also the question of how to create generic coordinates for this dimension. The first thing that comes to mind is VCF style genotype strings? For example `ds['genotype_id'] = "genotypes", ['0/0', '0/1', '1/1', '0/2', '1/2', '2/2']` for the triallelic case.

eric-czech:
> I like this idea although I'm not sure that we are relying on coordinates elsewhere as they seem to be largely optional?

That's true, but not intentional.  As far as I recall we just haven't had a great use case for it like this and FWIW, I was using coordinates quite a bit for metadata on human samples in the past with no big issues.

> The first thing that comes to mind is VCF style genotype strings?

Sounds reasonable to me!

timothymillar:
I guess there should be an option to not set the coordinates just in case someone wants to avoid their custom coords being clobbered. So API along the lines of:
```python
def count_variant_genotypes(
    ds: Dataset,
    *,
    call_genotype: Hashable = variables.call_genotype,
    assign_coords: bool = True,  # match xarray wording
    merge: bool = True,
)
```
which returns a dataset with `variant_genotype_count` (variants, genotypes) and `genotype_id` (genotypes) where `genotype_id` has been set as coords for the `genotypes` dimension if `assign_coords=True`.

timothymillar:
After rebasing, test failure seems unrelated (VCF `test_format_field `)

sgkit/stats/aggregation.py
jeromekelleher: Great!

jeromekelleher (approved):
LGTM - auto-merge whenever you're happy @timothymillar!

eric-czech:
> which returns a dataset with variant_genotype_count (variants, genotypes) and genotype_id (genotypes) where genotype_id has been set as coords for the genotypes dimension if assign_coords=True.

Awesome, sounds good to me! 👍

timothymillar:
Thanks for the reviews all. I've tidied up the commit history an implemented @eric-czech's suggestion to pass genotype counts by coords in `hardy_weinberg_test`. Quite a bit of functionality was added in [4147a54](https://github.com/pystatgen/sgkit/pull/1002/commits/4147a5473dd7e003fb1f23cb67f1f75b6f525730) to generate the coords so it'd be great if at least one person could review the two most recent commits. Otherwise good to go (reviewers feel free to auto-merge).

sgkit/stats/aggregation.py
```diff
@@ -477,9 +502,187 @@ def count_variant_genotypes(
     new_ds = create_dataset(
         {variables.variant_genotype_count: (("variants", "genotypes"), C)}
     )
+    if assign_coords is True and merge is False:
```
tomwhite, line 505: Could be simplified to `if assign_coords and not merge`.

What happens if `merge` is true? Are the coordinates assigned in that case?

tomwhite (approved):
This looks great!

One minor comment.

jeromekelleher (approved):
LGTM - happy to merge whenever

sgkit/stats/aggregation.py
timothymillar: Fixed.

Yes, the coords will already exist in `ds` from the initial `define_variable_if_absent` of `genotype_coords`. Any coords in `ds` are kept when it's merged with `new_ds` (checked in `test_genotype_coords`).

Issue #1010, opened 2023-01-30
Write family information in `write_plink`


tomwhite:
If the dataset has `sample_family_id`, `sample_paternal_id`, `sample_maternal_id` fields (e.g. from `read_plink`), then we can use those to write family information in `write_plink`. (See https://www.cog-genomics.org/plink/1.9/formats#fam)

Otherwise we should set FID to "0" (missing) and IID to `sample_id`. The father and mother IDs should either be set to missing, or set from the [`parent_id`](https://pystatgen.github.io/sgkit/latest/generated/sgkit.variables.parent_id_spec.html) variable if it is present.

(Or can we not assume anything about paternal/materal ordering in the `parent_id` variable? Thoughts @timothymillar, @jeromekelleher?)

jeromekelleher:
This is tricky, I don't think we thought much about interoperability with Plink when doing the pedigree encoding @timothymillar ?

timothymillar:
> Or can we not assume anything about paternal/materal ordering in the `parent_id` variable

In general no, but the intent was that these can be specified using coordinate when appropriate. So, you could use the `parent_id` array on the condition that appropriate coords are set.

PR #1019, opened 2023-02-09
Use `sample_*` variables if present to populate plink family fields when writing

### LINKED ISSUE #1017: Allow user to specify coordinate as phenotype in `write_plink`
Issue on top of PR https://github.com/pystatgen/sgkit/pull/1003
 
The idea is to give the user an option to specify a custom coordinate to populate the `pheno` property of `to_bed` with. This would make it much easier to write out to plink and directly feed into plink-based GWAS tools.

In terms of spec, one would want to check (at least) that the given coordinate is of dimension `samples` and of a numerical dtype.

tomwhite:
- [x] Fixes #1017 and a part of #1010
- [x] Tests added

The idea here is that you can use well-known variables, like `sample_phenotype` to set fields in the plink fam file. This should address #1017.

This isn't finished yet as it needs some tests. Also, I noticed that currently we only support case/control phenotypes (coded as 2 and 1, respectively) and not more general quantative trait values encoded as a float. This would need fixing too - I'm not sure the best way of doing that yet.


jeromekelleher:
@timothymillar would you mind taking a look through this please? I know very little about plink.

timothymillar:
> I know very little about plink

That makes two of us 😄.

sgkit/io/plink/plink_reader.py
```diff
@@ -106,8 +106,10 @@ def coerce_code(v: dd.Series, codes: List[int]) -> dd.Series:
         v = v.where(v.isin(codes), np.nan)
         return v.fillna(-1).astype("int8")
 
-    df["paternal_id"] = df["paternal_id"].where(df["paternal_id"] != "0", None)
-    df["maternal_id"] = df["maternal_id"].where(df["maternal_id"] != "0", None)
+    # replace fam "0" with sgkit missing values (".")
```
timothymillar, line 109: I think the missing value needs to be documented in the docstring of `read_plink`.

sgkit/io/plink/plink_writer.py
```diff
@@ -20,6 +31,20 @@ def write_plink(
 ) -> None:
```
timothymillar, line 31: Should this have a `missing: Hashable ='.' ` parameter? To specify the missing parent value and default to the new value from `read_plink`.

sgkit/io/plink/plink_writer.py
```diff
@@ -78,11 +103,12 @@ def write_plink(
...
         "allele_2": ds.variant_allele.values[:, 1],
-        "iid": ds.sample_id,
+        "iid": ds.sample_id.values,  # may be overridden by sample_member_id below (if present)
     }
 
-    if "variant_id" in ds:
-        properties["sid"] = ds.variant_id
+    for var, prop in FAM_VARIABLE_TO_BED_READER.items():
+        if var in ds:
+            properties[prop] = ds[var].values
```
timothymillar, line 111: I might be missing something but does the new/optional missing value `'.'` need to be transformed back to `0`?

sgkit/io/plink/plink_writer.py
tomwhite: I'm not sure this is needed. We don't parameterise the missing value in VCF for example, so I think it's ok to have the sgkit default of "." used here too.

sgkit/io/plink/plink_writer.py
tomwhite: Done

tomwhite:
Thanks for the review @timothymillar. I've updated the PR with changes.

sgkit/io/plink/plink_writer.py
timothymillar: Ah, I see it's handled by `values = np.where(values == STR_MISSING, "0", values)`. 
> We don't parameterise the missing value in VCF

Good point, but "." is the missing value in the VCF spec where as "0" is the missing value in a plink fam file. Although, a "0" will stay as a "0" so I guess both cases are covered anyway 🤷

> the sgkit default of "."

We should note that in the main documentation (not needed in this PR)

timothymillar (approved):
LGTM!

PR #1021, opened 2023-02-13
Generate VCF header when writing, if no header is explicitly supplied

### LINKED ISSUE #1020: Generate VCF header when writing VCF
Currently [`write_vcf`](https://pystatgen.github.io/sgkit/0.6.0/generated/sgkit.io.vcf.write_vcf.html) expects a `vcf_header` dataset attribute, which it uses for the header of the VCF file it produces.

One problem with this is that while it works for round tripping VCF -> Zarr -> VCF with no changes, in practice users want to modify the dataset in some way (typically by adding fields), which means that the header for writing will be different to the one for reading. So the current behaviour is unhelpful since it won't include the new fields. Users have to create the header themselves and then set it as a dataset attribute.

Instead, I propose that by default we generate the VCF header, while allowing users to override the default by explicitly specifying a header. This is a breaking change, but I think it is a better user experience.

tomwhite:
- [x] Fixes #1020
- [x] Tests added

Not ready to merge (but happy to discuss) since it doesn't have full coverage yet.

### DIFF
--- a/sgkit/io/vcf/vcf_writer.py
+++ b/sgkit/io/vcf/vcf_writer.py
@@ -1,12 +1,15 @@
+import io
 import re
 from contextlib import ExitStack
 from pathlib import Path
-from typing import MutableMapping, TextIO, Union
+from typing import MutableMapping, Optional, TextIO, Union
 
 import numpy as np
 from xarray import Dataset
 
 from sgkit import load_dataset
+from sgkit.io.utils import FLOAT32_MISSING
+from sgkit.io.vcf.vcf_reader import RESERVED_VARIABLE_NAMES
 from sgkit.io.vcf.vcf_writer_utils import (
     byte_buf_to_str,
     create_mask,
@@ -26,16 +29,73 @@ from sgkit.io.vcf.vcf_writer_utils import (
 from sgkit.model import get_contigs, get_filters
 from sgkit.typing import PathType
 
+# references to the VCF spec are for https://samtools.github.io/hts-specs/VCFv4.3.pdf
+
+# [Table 1: Reserved INFO keys]
+RESERVED_INFO_KEY_DESCRIPTIONS = {
+    "AA": "Ancestral allele",
+    "AC": "Allele count in genotypes, for each ALT allele, in the same order as listed",
+    "AD": "Total read depth for each allele",
+    "ADF": "Read depth for each allele on the forward strand",
+    "ADR": "Read depth for each allele on the reverse strand",
+    "AF": "Allele frequency for each ALT allele in the same order as listed",
+    "AN": "Total number of alleles in called genotypes",
+    "BQ": "RMS base quality",
+    "CIGAR": "Cigar string describing how to align an alternate allele to the reference allele",
+    "DB": "dbSNP membership",
+    "DP": "Combined depth across samples",
+    "END": "End position on CHROM",
+    "H2": "HapMap2 membership",
+    "H3": "HapMap3 membership",
+    "MQ": "RMS mapping quality",
+    "MQ0": "Number of MAPQ == 0 reads",
+    "NS": "Number of samples with data",
+    "SB": "Strand bias",
+    "SOMATIC": "Somatic mutation",
+    "VALIDATED": "Validated by follow-up experiment",
+    "1000G": "1000 Genomes membership",
+}
+
+# [Table 2: Reserved genotype keys]
+RESERVED_FORMAT_KEY_DESCRIPTIONS = {
+    "AD": "Read depth for each allele",
+    "ADF": "Read depth for each allele on the forward strand",
+    "ADR": "Read depth for each allele on the reverse strand",
+    "DP": "Read depth",
+    "EC": "Expected alternate allele counts",
+    "FT": 'Filter indicating if this genotype was "called"',
+    "GL": "Genotype likelihoods",
+    "GP": "Genotype posterior probabilities",
+    "GQ": "Conditional genotype quality",
+    "GT": "Genotype",
+    "HQ": "Haplotype quality",
+    "MQ": "RMS mapping quality",
+    "PL": "Phred-scaled genotype likelihoods rounded to the closest integer",
+    "PP": "Phred-scaled genotype posterior probabilities rounded to the closest integer",
+    "PQ": "Phasing quality",
+    "PS": "Phase set",
+}
+
 
 def write_vcf(
-    input: Dataset,
-    output: Union[PathType, TextIO],
+    input: Dataset, output: Union[PathType, TextIO], *, vcf_header: Optional[str] = None
 ) -> None:
     """Convert a dataset to a VCF file.
 
-    The VCF fields included in the output are those in the ``vcf_header``
-    attribute of the dataset. There is currently no way to change the fields
-    that are included apart from by manually updating this attribute.
+    The VCF header to use is dictated by the ``vcf_header`` parameter, and the
+    ``vcf_header`` attribute on the input dataset. The ``vcf_header`` parameter
+    takes precedence and will be used if specified. Any variables in the dataset
+    that are not in the header will not be included in the output.
+
+    Otherwise, if the ``vcf_header`` attribute is present (because the dataset
+    derives from one imported via :func:`vcf_to_zarr`) it will be used to
+    generate the new VCF header. Any variables in the dataset that are not
+    specified in the original header will have corresponding header lines
+    added, and any variables not in the dataset that are in the original
+    header will have the corresponding header line omitted.
+
+    In the case of no ``vcf_header`` parameter or attribute, a VCF header will
+    be generated, and will include all variables in the dataset.
 
     Float fields are written with up to 3 decimal places of precision.
     Exponent/scientific notation is *not* supported, so values less than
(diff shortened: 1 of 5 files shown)


sgkit/io/vcf/vcf_writer.py
```diff
@@ -25,16 +28,59 @@
 )
 from sgkit.typing import PathType
 
+# [Table 1: Reserved INFO keys]
```
jeromekelleher, line 31: Good to note what version of VCF this corresponds to, it'll make it easier to stay up to date over time.

sgkit/io/vcf/vcf_writer.py
```diff
@@ -60,36 +106,37 @@ def write_vcf(
...
+        The VCF header to use (including the line starting with ``#CHROM``). If omitted, then
+        a header will be generated from the dataset fields.
     """
 
     with ExitStack() as stack:
         if isinstance(output, str) or isinstance(output, Path):
             output = stack.enter_context(open(output, mode="w"))
 
-        header_str = input.attrs["vcf_header"]
+        if vcf_header is None:
```
jeromekelleher, line 124: This feels like a breaking change - shouldn't we at least warn the user if there is a ``vcf_header`` attribute? Would it not make sense to use the ``vcf_header`` attribute if we already have one? I.e., parse the header (if it exists) and check whether all the fields match up, and generate any new ones as necessary?

jeromekelleher (commented):
Looks good, but I think we need to update any existing headers rather than autogenerate unfortunately. All sorts of stuff will be stored in these headers and autogenerating a minimal header for the data will lose that information.

sgkit/io/vcf/vcf_writer.py
tomwhite: Yes, it is a breaking change (mentioned on #1020), and a warning is probably a good idea.

VCF header management is tricky though, and merging two headers even more so. What would we do if the version of the VCF header in the attribute is older than the one we are using to write? VCF 4.3 had lots of changes in things like header name validation - basically making things stricter - so it wouldn't be very safe to use a line from the original header since it might not be valid in the later version. (To take one example, PEDIGREE fields now require an ID.)

jeromekelleher: I agree it's very tricky. Maybe I'm misunderstanding though - what happens in the case where we do something like

- Read in a VCF to dataset
- Compute something which changes the dataset (say, phase the data, just for example)
- Export back out to VCF

Do we lose all header information beyond the fields and types, like comments and provenance information from the original?

I agree parsing out the header and trying to understand every bit of it would be very difficult, but you could do something simpler like assuming that if header information exists for a given field, it's correct.

tomwhite: To give an example, here is a diff of the output header (generated by the code in this PR) with the input header for the sample.vcf in this repo. (The lines are sorted to make it easier to compare - normally they are not in this order of course.)

<details>

```
E             + ##ALT=<ID=CNV,Description="Copy number variable region">
E             + ##ALT=<ID=DEL:ME:ALU,Description="Deletion of ALU element">
E               ##FILTER=<ID=PASS,Description="All filters passed">
E             - ##FILTER=<ID=q10,Description="">
E             + ##FILTER=<ID=q10,Description="Quality below 10">
E             ?                               ++++++++++++++++
E             - ##FILTER=<ID=s50,Description="">
E             + ##FILTER=<ID=s50,Description="Less than 50% of samples have data">
E               ##FORMAT=<ID=DP,Number=1,Type=Integer,Description="Read Depth">
E               ##FORMAT=<ID=GQ,Number=1,Type=Integer,Description="Genotype Quality">
E               ##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
E               ##FORMAT=<ID=HQ,Number=2,Type=Integer,Description="Haplotype Quality">
E               ##INFO=<ID=AA,Number=1,Type=String,Description="Ancestral Allele">
E             - ##INFO=<ID=AC,Number=2,Type=Integer,Description="Allele count in genotypes, for each ALT allele, in the same order as listed">
E             ?                      ^
E             + ##INFO=<ID=AC,Number=.,Type=Integer,Description="Allele count in genotypes, for each ALT allele, in the same order as listed">
E             ?                      ^
E             - ##INFO=<ID=AF,Number=2,Type=Float,Description="Allele Frequency">
E             ?                      ^
E             + ##INFO=<ID=AF,Number=.,Type=Float,Description="Allele Frequency">
E             ?                      ^
E               ##INFO=<ID=AN,Number=1,Type=Integer,Description="Total number of alleles in called genotypes">
E               ##INFO=<ID=DB,Number=0,Type=Flag,Description="dbSNP membership, build 129">
E               ##INFO=<ID=DP,Number=1,Type=Integer,Description="Total Depth">
E               ##INFO=<ID=H2,Number=0,Type=Flag,Description="HapMap2 membership">
E               ##INFO=<ID=NS,Number=1,Type=Integer,Description="Number of Samples With Data">
E               ##contig=<ID=19>
E               ##contig=<ID=20>
E               ##contig=<ID=X>
E             + ##fileDate=20090805
E             - ##fileformat=VCFv4.3
E             ?                    ^
E             + ##fileformat=VCFv4.0
E             ?                    ^
E             - ##source=sgkit-0.6.1.dev2+gcc728043
E             + ##phasing=partial
E             + ##reference=1000GenomesPilot-NCBI36
```

</details>

Summary of differences:
* `fileformat` version numbers
* `source`
* Extra unstructured (key/value) fields in the original:
  * `fileDate`
  * `phasing`
  * `reference`
* Extra structured fields in the original:
  * `ALT`
* Missing description in `FILTER`
* More specific `Number` (e.g. `.` in original, `2` in output)

I've run this on other test files in the repo, and this difference crops up too:
* Missing `assembly` field in `contig` (related to #464)

tomwhite: > I agree parsing out the header and trying to understand every bit of it would be very difficult, but you could do something simpler like assuming that if header information exists for a given field, it's correct.

Good idea - I'm investigating this approach...

tomwhite:
OK, I've updated to follow the approach suggested by @jeromekelleher, which is to use existing VCF header lines if they are available - and only generate header lines if they are not available. See new API doc for details.

jeromekelleher (commented):
I've not gone through the gory details, but LGTM. Perhaps @benjeffery could take a closer look?

benjeffery:
@tomwhite I see you're still pushing here, let me know when I should review.

tomwhite:
@benjeffery It's ready for a review now (I was just fixing conflicts). Thanks!

sgkit/io/vcf/vcf_writer.py
```diff
@@ -26,16 +29,73 @@
...
+    ``vcf_header`` attribute on the input dataset. The ``vcf_header`` parameter
+    takes precedence and will be used if specified. Any variables in the dataset
+    that are not in the header will not be included in the output.
+
+    Otherwise, if the ``vcf_header`` attribute is present (because the dataset
+    derives from one imported via :func:`vcf_to_zarr`) it will be used to
+    generate the new VCF header. Any variables in the dataset that are not
+    specified in the original header will have corresponding header lines
+    added, and any variables not in the dataset that are in the original
+    header will have the corresponding header line omitted.
```
benjeffery, line 95: I found this hard to parse - how about:

```suggestion
    The VCF header to use is dictated by either the ``vcf_header`` parameter or the
    ``vcf_header`` attribute on the input dataset. 

    If specified, the ``vcf_header`` parameter will be used, and any variables in the dataset
    that are not in this header will not be included in the output.

    If the ``vcf_header`` parameter is left as the default (`None`) and a ``vcf_header`` 
    attribute is present in the dataset (such as one created by :func:`vcf_to_zarr`), 
    it will be used to generate the new VCF header. In this case, any variables in the
    dataset that are not specified in this header will have corresponding header lines
    added, and any lines in the header without a corresponding variable in the dataset
    will be omitted.
```

sgkit/tests/io/vcf/test_vcf_writer.py
```diff
@@ -129,17 +134,153 @@ def test_write_vcf__drop_fields(shared_datadir, tmp_path):
...
+
+    # check dropped fields are not present in VCF
+    v = VCF(output)
+    assert "##INFO=<ID=NS" not in v.raw_header
+    assert "##FORMAT=<ID=HQ" not in v.raw_header
+    count = 0
+    for variant in v:
+        assert "NS" not in dict(variant.INFO).keys()
+        assert "HQ" not in variant.FORMAT
+        assert variant.genotypes is not None
```
benjeffery, line 167: Should this also assert that H3 and GL are retained?

benjeffery (approved):
Nice! Just a couple of comments.
Should there also be tests for what happens in edge cases? Empty string header arguments, that kind of thing.

tomwhite:
> Should there also be tests for what happens in edge cases? Empty string header arguments, that kind of thing.

If a header is passed in by the user then it will be used as-is, so an empty string header will result in a headerless VCF. I'm not sure that's something we want to support or encourage though - perhaps we leave it as undefined at this point?

sgkit/tests/io/vcf/test_vcf_writer.py
tomwhite: Done

benjeffery:
Great, this looks ready to merge now.

jeromekelleher (approved):

PR #1028, opened 2023-02-20
Improve method `identity_by_state`

### LINKED ISSUE #1026: Performance issue with `identity_by_state` on samples dimension
I noticed the `Weir_Goudet_beta` method scales poorly with the number of samples. After a bit of profiling this seems to be caused by `da.einsum` (also similar result for `np.einsum`) in the `identity_by_state` method.

Using 1000 variants with 2 alleles and varying the number of samples:

perf

Where blue uses `da.einsum`:
```python
da.einsum("ixj,iyj->xy", af, af).compute()
```

and an alternative in orange:
```python
sum(m.T @ m for m in af.transpose(2, 0, 1)).compute()
```

Interestingly, if I using the alternative implementation on a numpy array I get poor performance unless I explicitly copy the transposed array  (`af.transpose(2, 0, 1).copy()`). I assume this must be due to numpy passing by reference with fancy indexing which results in a slow matmul.

I also checked the performance of the alternative version when increasing the number of alleles as there's efectively a python loop over this dimension. Using 1000 variants with 100 samples and varying allele number:

perf2

Which seems fine and is actually slightly better for the biallelic case.

### LINKED ISSUE #1027: Add skipna option to method `identity_by_state`
The `identity_by_state` method currently skips nan allele frequencies. This is a good default, but it should be optional (there is some performance overhead).

timothymillar:
- [x] Fixes #1026 
- [x] Fixes #1027 
- [x] Tests added

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,8 +23,14 @@ New Features
 .. Deprecations
 .. ~~~~~~~~~~~~
 
-.. Improvements
-.. ~~~~~~~~~~~~
+Improvements
+~~~~~~~~~~~~
+
+- Improve performance scaling of method `sgkit.identity_by_state`
+  with number of samples.
+  (:user:`timothymillar`, :pr:`1028`, :issue:`1026`)
+- Add ``skipna`` option to method `sgkit.identity_by_state`.
+  (:user:`timothymillar`, :pr:`1028`, :issue:`1027`)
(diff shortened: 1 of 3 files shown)


tomwhite (approved):
LGTM

jeromekelleher (approved):

PR #1034, opened 2023-02-28
Split the concat_zarrs step to avoid very large dask task counts

benjeffery:
When converting very large VCFs I'm running out of RAM when constructing the task graph for the `concat_zarrs` step. This PR breaks the task graph into the independent steps.

### DIFF
--- a/sgkit/io/vcfzarr_reader.py
+++ b/sgkit/io/vcfzarr_reader.py
@@ -340,7 +340,6 @@ def concat_zarrs_optimized(
     # NOTE: that this uses _to_zarr function defined here that is needed to avoid
     # race conditions between writing the array contents and its metadata
     # see https://github.com/pystatgen/sgkit/pull/486
-    delayed = []  # do all the rechunking operations in one computation
     for var in vars_to_rechunk:
         dtype = None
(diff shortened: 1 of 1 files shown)


tomwhite:
Thanks for this! Does this change fix your problem?

Do you think we should keep the old behaviour as a default for smaller datasets or doesn't it make much difference?

benjeffery:
With this change the conversion of this dataset was successful!
Screenshot from 2023-02-28 10-42-57
Thanks for the amazing work on the vcf conversion code @tomwhite, it has worked very well.

benjeffery:
> Do you think we should keep the old behaviour as a default for smaller datasets or doesn't it make much difference?

I'll run a quick comparison on a smaller VCF.

benjeffery:
I've done a test on chr 22 of the 1k genomes data, there is no difference in runtime.

tomwhite (approved):
+1

Pending fix to pre-commit.

benjeffery:
BTW - I tried bigger blocks to reduce the number of dask tasks needed, but could only go 2.5x bigger than the default before I hit a RAM limit in the first vcf parsing step, so it is a tradeo-off between these two steps.

benjeffery:
@mergify rebase

PR #1039, opened 2023-02-28
Defer numba guvectorize compilation for aggregate functions

### LINKED ISSUE #939: Run `numba` compilation as a post-install step?
Currently, this is the import performance of `sgkit` on a fresh install:

```
(testenv) benj@benj-X580VD:~$ time python -c "import sgkit"

real	0m16.715s
user	0m16.149s
sys	0m0.987s
(testenv) benj@benj-X580VD:~$ time python -c "import sgkit"

real	0m2.184s
user	0m2.414s
sys	0m0.914s
```

It might be possible to move the 14s of `numba` compilation to a post-install step, hopefully on `pip` and `conda`.

tomwhite:
- [x] Fixes #939

This shows the approach described in https://github.com/pystatgen/sgkit/issues/939#issuecomment-1315086288 for the aggregation functions.

The import goes from 13s to 10s on my machine using

```shell
export SGKIT_DISABLE_NUMBA_CACHE=1
time python -c "import sgkit"
```

If we did this for all the guvectorize usages we can get to around 3s.

This is simply an internal reorganisation so shouldn't affect user APIs. It is quite a disruptive change though, so I wanted to get any thoughts before I do it for the whole codebase.


timothymillar:
Thanks for looking at this. When a single gufunc is imported from `aggregation_numba_fns.py`, are all the gufuncs in that file compiled? If so, would it make more sense to have a separate file per gufunc to make compilation even more modular (ignoring the mess that creates)? But this general approach seems fine to me.

tomwhite:
Thanks Tim.

> When a single gufunc is imported from `aggregation_numba_fns.py`, are all the gufuncs in that file compiled? If so, would it make more sense to have a separate file per gufunc to make compilation even more modular (ignoring the mess that creates)?

Yes, all the gufuncs in that file are compiled, but I think that's OK. The goal is to avoid the 13 second pause when a user first types `import sgkit`. There will be a delay when they first run a function, but it will be less than 13 seconds, and it will be more forgivable since the function is doing something useful. The granularity of a compile unit per module rather than per function should therefore be a reasonable compromise.

tomwhite:
Now applied across the codebase. The import time is now 1.4s on my machine.

benjeffery (approved):
LGTM! This is great, thanks @tomwhite

jeromekelleher (approved):
Great idea, and totally agree with the pattern.

PR #1044, opened 2023-03-02
Add read_chunk_length

### LINKED ISSUE #1042: `vcf_to_zarr_sequential` high memory usage with large sample counts.
The current VCF I'm working on has 29,924,147 sites and 78,195 samples. With the default chunk sizes (10,000x1,000) this results in 236,447 genotype chunks. Any dask task graphs run on this become very slow as most of the time is spent waiting for the scheduler. The genotype chunks are also very small on disk at around 20KB - this isn't ideal on a cluster filesystem such as NFS where reading fewer, larger chunks better (at least for tasks where you read most of the genotypes, which is mostly what I need).

Ideally, I would have chunks of around (100,000x3,000) I have managed to raise the chunk size to (15,000x3,000), but can't go above this due to `vcf_to_zarr_sequential` using too much RAM. This is because it allocates `chunk_length*n_samples*(1+n_ploidy)` bytes which for my ideal chunks is `23.4GB`. The max I can get per-core before my job is terminated is 16GB.

It seems a solution would be to add an argument to `vcf_to_zarr_sequential` that sets a `read_chunk_length` that is separate from `chunk_length`. Then the max RAM is easily controlled as `max(read_chunk_length*n_samples*(1+n_ploidy), chunk_length*chunk_width*n_ploidy)`. The cost here is that when appending the read chunk to the `zarr` file you need to recode the zarr chunks `chunk_length/read_chunk_length` times, but I think that will be a small factor compared to the VCF parsing.

It looks like this change is simple to implement I will do some perf checks and raise a PR.

EDIT: If the VCF is read in ~20MB chunks (the default) then with a high number of samples `chunk_length` number of variants is almost never reached (generally around 15,000) so the memory is unused any way and the number of zarr re-writes low.

benjeffery:
- [x] Fixes #1042 
- [x] Tests added

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,11 @@ New Features
 - Add :func:`sgkit.convert_call_to_index` method.
   (:user:`timothymillar`, :pr:`1050`, :issue:`1048`)
 
+- Add ``read_chunk_length`` option to :func:`sgkit.io.vcf.vcf_to_zarr` and
+  :func:`sgkit.io.vcf.vcf_to_zarrs` functions. These are useful to reduce memory usage
+  with large sample counts or large ``chunk_lengths``.
+  (:user:`benjeffery`, :pr:`1044`, :issue:`1042`)
+
 - Add ``retain_temp_files`` to :func:`sgkit.io.vcf.vcf_to_zarr` function.
   (:user:`benjeffery`, :pr:`1046`, :issue:`1036`)
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -26,11 +26,20 @@ New Features
 
 - Add ``read_chunk_length`` option to :func:`sgkit.io.vcf.vcf_to_zarr` and
   :func:`sgkit.io.vcf.vcf_to_zarrs` functions. These are useful to reduce memory usage
-  with large sample counts or large ``chunk_lengths``.
+  with large sample counts or a large ``chunk_length``.
   (:user:`benjeffery`, :pr:`1044`, :issue:`1042`)
 
 - Add ``retain_temp_files`` to :func:`sgkit.io.vcf.vcf_to_zarr` function.
   (:user:`benjeffery`, :pr:`1046`, :issue:`1036`)
+
+Bug fixes
+~~~~~~~~~
+
+- ``vcf_to_zarrs`` was outputting zarr files with ``temp_chunk_length`` sized chunks. This has now been fixed
+  so that the chunks are ``chunk_length`` sized.
+  (:user:`benjeffery`, :pr:`1044`,)
+
+
 
 .. Breaking changes
 .. ~~~~~~~~~~~~~~~~
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -416,7 +416,10 @@ def vcf_to_zarr_sequential(
     fields: Optional[Sequence[str]] = None,
     exclude_fields: Optional[Sequence[str]] = None,
     field_defs: Optional[Dict[str, Dict[str, Any]]] = None,
+    read_chunk_length: Optional[int] = None,
 ) -> None:
+    if read_chunk_length is None:
+        read_chunk_length = chunk_length
     with open_vcf(input) as vcf:
         sample_id = np.array(vcf.samples, dtype="O")
         n_allele = max_alt_alleles + 1
@@ -436,7 +439,7 @@ def vcf_to_zarr_sequential(
         # Remember max lengths of variable-length strings
         max_alt_alleles_seen = 0
 
-        # Iterate through variants in batches of chunk_length
+        # Iterate through variants in batches of read_chunk_length
 
         if region is None:
             variants = vcf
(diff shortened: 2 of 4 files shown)


docs/changelog.rst
```diff
@@ -17,6 +17,11 @@ New Features
 - Add ``phased`` option to :func:`simulate_genotype_call_dataset` function.
   (:user:`tomwhite`, :pr:`1022`, :issue:`973`)
 
+- Add ``read_chunk_length`` option to :func:`sgkit.io.vcf.vcf_to_zarr` and
+  :func:`sgkit.io.vcf.vcf_to_zarrs` functions. These are useful to reduce memory usage
+  with large sample counts or large ``chunk_lengths``.
```
tomwhite, line 22: Perhaps reword this as the parameter is `chunk_length`, not `chunk_lengths`.

tomwhite (approved):
A few typos to fix but otherwise looks good.

Would be good to add something to https://pystatgen.github.io/sgkit/latest/vcf.html#chunk-sizes

benjeffery:
@tomwhite I've just realised there is some conflict here with `temp_chunk_length` that needs thinking about. I couldn't use that to fix this particular issue as having small temp chunks makes the `concat_zarrs` task graph too large. Maybe `temp_chunk_length` could be deprecated in favour of `read_chunk_length`?

tomwhite:
> @tomwhite I've just realised there is some conflict here with `temp_chunk_length` that needs thinking about. I couldn't use that to fix this particular issue as having small temp chunks makes the `concat_zarrs` task graph too large. Maybe `temp_chunk_length` could be deprecated in favour of `read_chunk_length`?

Are these two parameters the same?

I think there may be a problem with `temp_chunk_length` anyway. If specified, it correctly writes intermediate Zarr files with smaller chunk lengths (e.g. in the test `test_vcf_to_zarr__parallel_temp_chunk_length` its chunks are of size 2500). However, the `chunk_length` is ignored in `concat_zarrs`, which seems wrong. (In the `test_vcf_to_zarr__parallel_temp_chunk_length` the final chunks are still 2500 variants long, not 5000.)

Is the intent of `read_chunk_length` the same? Regardless, I think `test_vcf_to_zarr__large_vcf` should have some assertions about chunk sizes to check that specifying `read_chunk_length` has the intended effect.

benjeffery:
> > @tomwhite I've just realised there is some conflict here with `temp_chunk_length` that needs thinking about. I couldn't use that to fix this particular issue as having small temp chunks makes the `concat_zarrs` task graph too large. Maybe `temp_chunk_length` could be deprecated in favour of `read_chunk_length`?
> 
> Are these two parameters the same?

Not quite - there are three different chunk sizes here!

1. The final chunk size after `concat_zarrs` currently `chunk_length`.
2. The intermediate on-disk chunk size of the parts before `concat_zarrs` currently `temp_chunk_length`.
3. The size of the read buffer in-memory when parsing the VCF currently `read_chunk_length`

I've changed my mind on the deprecation as I need to optimise all three when parsing VCFs:

- Set `read_chunk_length` low enough so that my dask workers don't die due to excessive memory use.
- Set `temp_chunk_length` higher so that the dask task graph for `concat_zarrs` isn't intractable due to scheduler memory usage, but not too large to avoid excessive re-writes of the temporary zarr chunks. 
- Set `chunk_length` higher again so that eventual computations on the genotype array have the right balance of dask scheduler performance and random access.

> I think there may be a problem with `temp_chunk_length` anyway. If specified, it correctly writes intermediate Zarr files with smaller chunk lengths (e.g. in the test `test_vcf_to_zarr__parallel_temp_chunk_length` its chunks are of size 2500). However, the `chunk_length` is ignored in `concat_zarrs`, which seems wrong. (In the `test_vcf_to_zarr__parallel_temp_chunk_length` the final chunks are still 2500 variants long, not 5000.)

I'll look into this!

> Is the intent of `read_chunk_length` the same? Regardless, I think `test_vcf_to_zarr__large_vcf` should have some assertions about chunk sizes to check that specifying `read_chunk_length` has the intended effect.

I'll have to think a bit about this - you can't detect the `read_chunk_length` in the temp chunks, or the final chunks.

tomwhite:
Looks like this still needs updating. @benjeffery have you got this working yet on a cluster?

tomwhite:
Looks like this still needs updating. @benjeffery have you got this working yet on a cluster?

docs/changelog.rst
```diff
@@ -24,9 +24,23 @@ New Features
...
+  (:user:`benjeffery`, :pr:`1044`, :issue:`1042`)
+
 - Add ``retain_temp_files`` to :func:`sgkit.io.vcf.vcf_to_zarr` function.
   (:user:`benjeffery`, :pr:`1046`, :issue:`1036`)
 
+Bug fixes
+~~~~~~~~~
+
+- ``vcf_to_zarrs`` was outputting zarr files with ``temp_chunk_length`` sized chunks. This has now been fixed
+  so that the chunks are ``chunk_length`` sized``
```
tomwhite, line 39: Spurious double backtick at end of line.

tomwhite (approved):
This looks good. There are a couple of minor doc fixes that I mentioned in my last review that still need addressing. Otherwise ready to merge.

benjeffery:
Fixed up the doc issues! Thanks for the reminder.

PR #1046, opened 2023-03-02
Add retain_temp_files

### LINKED ISSUE #1036: When converting large VCFs ~50% of wall clock time is spent deleting temporary files
After the `concat_zarrs` step the temporary zarr files have to be deleted. As the VCF to convert gets larger the fraction of time taken by this step of the process rises - becoming more than half the time above a ~5GB compressed VCF, at least on the `NFS` system I'm using. This step should also use dask to parallelize the deletion of the files.

benjeffery:
- [x] Fixes #1036
- [x] Tests added


sgkit/io/vcf/utils.py
```diff
@@ -144,7 +145,9 @@ def temporary_directory(
         The directory may be specified as any fsspec URL.
     storage_options : Optional[Dict[str, str]], optional
         Any additional parameters for the storage backend (see `fsspec.open`).
-
+    retain_files : Optional[bool], optional
```
jeromekelleher, line 148: param name doesn't match here

benjeffery: Fixed

benjeffery:
Comments fixed - should be good to go.

sgkit/io/vcf/utils.py
```diff
@@ -129,6 +129,7 @@ def temporary_directory(
     prefix: Optional[str] = None,
     dir: Optional[PathType] = None,
     storage_options: Optional[Dict[str, str]] = None,
+    retain_temp_files: bool = None,
```
tomwhite, line 132: Should be `Optional[bool]`

tomwhite (approved):
LGTM

PR #1053, opened 2023-03-20
Hybrid relationship methods

### LINKED ISSUE #993: Hybrid relationship matrices
Implement hybrid relationship matrix caluclation (H-matrix) (and its inverse) following [Martini 2018](https://gsejournal.biomedcentral.com/articles/10.1186/s12711-018-0386-x). This method combines pedigree and genomic relationship estimates. It can be used to "project" genomic estimates for a subset of samples onto a (full) pedigree based matrix. This is the matrix used in the single-step GBLUP #824

timothymillar:
- [x] Fixes #993 
- [x] Tests added
- [x] Commit history squashed
- [x] Warnings/Raises added to docstrings
- [x] Examples added to docstrings

Opening PR to track progress, needs more testing and examples.


timothymillar:
This is ready for review now, but I would like a second opinion on API design. The API is slightly awkward in an xarray dataset because the GRM `(genotyped_samples * genotyped_samples)` can be a subset of the ARM `(samples * samples)`. Currently this requires an index for each sample in the GRM which specifies its position in the ARM (`genomic_sample_index`). An alternative would be to join these matrices via coordinates (sample identifiers) so long as all coordinates of the GRM are found in the ARM. We could make this optional so that `genomic_sample_index` is only required if coordinates have not been assigned.

sgkit/stats/grm.py
```diff
@@ -147,3 +148,522 @@ def genomic_relationship(
...
+        # H = da.linalg.inv(tau * da.linalg.inv(G))
+    else:
+        H = da.linalg.inv(tau * da.linalg.inv(G) + (1 - omega) * Ainv)
+    return H
+
+
+def _hybrid_sub_matrix_Martini(A, G, Gidx, tau=1, omega=1, A22inv=None):
+    n = len(A)
+    # indices of non G-matrix
+    bools = np.ones(n, bool)
```
benjeffery, line 176: There are a few points in this function where `np` methods are used - I'm not familiar enough with the method to know if it is the right thing to do to bring computation out of dask at those points, but a comment explaining why it's ok (for example where it is known the arrays will be small at that point) would be good.

sgkit/stats/grm.py
```diff
@@ -147,3 +148,522 @@ def genomic_relationship(
...
+    "A relationship matrix including full pedigree and genomic information."
+    Journal of Dairy Science 92 (9): 4656-4663.
+
+    [3] - I. Aguilar, I. Misztal, D. L. Johnson, A. Legarra, S. Tsuruta and T. J. Lawlor 2010.
+    "A unified approach to utilize phenotypic, full pedigree, and genomic information for genetic
+    evaluation of Holstein final score."
+    Journal of Dairy Science 93 (2): 743-752.
+
+    """
+    estimator = estimator or "Martini"
```
benjeffery, line 364: Best to be explicit here so that arguments such as `False` raise the right error.
```suggestion
    if estimator is None:
         estimator = "Martini"
```

sgkit/stats/grm.py
```diff
@@ -147,3 +148,522 @@ def genomic_relationship(
...
+    [2] - A. Legarra, I. Aguilar and I. Misztal 2009.
+    "A relationship matrix including full pedigree and genomic information."
+    Journal of Dairy Science 92 (9): 4656-4663.
+
+    [3] - I. Aguilar, I. Misztal, D. L. Johnson, A. Legarra, S. Tsuruta and T. J. Lawlor 2010.
+    "A unified approach to utilize phenotypic, full pedigree, and genomic information for genetic
+    evaluation of Holstein final score."
+    Journal of Dairy Science 93 (2): 743-752.
+    """
+    estimator = estimator or "Martini"
```
benjeffery, line 624:
```suggestion
    if estimator is None
        estimator = "Martini"
```

sgkit/tests/test_grm.py
```diff
@@ -173,3 +174,745 @@ def test_genomic_relationship__raise_on_ancestral_frequency_missing():
...
+@pytest.mark.parametrize("samples_chunks", [None, 10])
+@pytest.mark.parametrize("genotypes_chunks", [None, 2])
+@pytest.mark.parametrize("grm_position", ["default", "first", "last", "mixed"])
+def test_hybrid_relationship__Legarra(grm_position, genotypes_chunks, samples_chunks):
+    # simulate example from Legarra et al 2009
+    ds = load_Legarra2009_example()
+    ds = sg.pedigree_kinship(ds, return_relationship=True)
+    if grm_position == "first":
+        # sort so GRM samples are first
+        idx = [
```
benjeffery, line 405: If you want the auto-formating to keep these on one line, remove the final comma.

benjeffery (approved):
Very nice!
On your API question - would arrays of the same shape but with NaNs for the rows/columns that are missing work? I assume then you wouldn't need to index in to them?

timothymillar:
> would arrays of the same shape but with NaNs for the rows/columns that are missing work? I assume then you wouldn't need to index in to them?

I did consider this approach, but as far as I can tell you would still need to index into the arrays to extract the relevant sub matrices. `da.linalg.inv` is a bit of a pain point here because it needs a matrix of known size and can't handle NaN values. Using a padded G-matrix also increases memory use, although that's a minor concern as it as at most the same size as the A-matrix.

I guess we could use a NaN-padded G-matrix in combination with a boolean array `genomic_samples` to indicate the non-NaN samples within the G-matrix. This would only require the `samples[_0,_1]` dimensions. We could then have an option `genomic_samples=None` which results in the automatic computation of the `genomic_samples` array by early evaluation of the G-matrix (with appropriate performance warnings in the docs).

sgkit/stats/grm.py
timothymillar: Good call, these arrays are used to slice the A-matrix into sub-matrices. They need to be evaluated eagerly so that the sub-matrices have known size. They are all 1-d arrays with maximum size of n_samples

sgkit/stats/grm.py
```diff
@@ -147,3 +148,522 @@ def genomic_relationship(
...
+            Ainv = da.linalg.inv(A)
+        H = _hybrid_full_matrix_Martini(Ainv=Ainv, G=G, tau=tau, omega=omega)
+    else:
+        # Following Martini with G-matrix as a subset of A-matrix
+        if genomic_sample_index is None:
+            raise ValueError(
+                "The genomic and pedigree matrices must share dimensions"
+                " if genomic_sample_index is not defined"
+            )
+        Gidx = ds[genomic_sample_index].values
```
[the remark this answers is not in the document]
timothymillar: Same as above

sgkit/tests/test_grm.py
timothymillar: Thanks!

timothymillar:
@benjeffery I'm trying an API using NaN-padded matrices on this [branch](https://github.com/timothymillar/sgkit/blob/993-hybrid-relationships-api2/sgkit/stats/grm.py#L324). It seems like a big improvement so will push back here when it's ready. It does still require an array to indicate genomic samples. Thanks for the suggestion!

benjeffery:
Great! Your plan in the message above sounds good.

timothymillar:
Could you please have a look at the revised API @benjeffery? This uses NaN-padded matrices as discussed. I decided on requiring explicit specification of submatrices (rather than automatically inferring them) because it seems like the least surprising option. Furthermore, I can't imagine a situation in which identifying genomic samples via NaN values is a good practice. This puts a little extra burden on the user for the sake of clarity.

sgkit/stats/grm.py
```diff
@@ -147,3 +149,655 @@ def genomic_relationship(
...
+
+
+def _hybrid_full_matrix_Martini(
+    Ainv: ArrayLike, G: ArrayLike, tau: float = 1, omega: float = 1
+):
+    # calculate sub-matrices of H
+    if (tau == 1) and (omega == 1):
+        H = G
+    elif omega == 1:
+        H = 1 / tau * G
```
jeromekelleher, line 277: Maybe comment here on why we're not using the Dask version (or delete the Dask line)?

sgkit/stats/grm.py
```diff
@@ -147,3 +149,655 @@ def genomic_relationship(
...
+    if A22inv is None:
+        A22inv = da.linalg.inv(A22)
+    else:
+        A22inv = A22inv[is_genomic, :][:, is_genomic]
+    if (tau == 1) and (omega == 1):
+        H22 = G22
+    elif omega == 1:
+        H22 = 1 / tau * G22
+    else:
+        H22 = da.linalg.inv(tau * da.linalg.inv(G22) + (1 - omega) * A22inv)
```
jeromekelleher, line 310: This line looks like it could involve a lot of computation (an inverse of another inverse). Worth splitting across two lines, just for the sake of providing accurate feedback on line-numbers when doing profiling?

jeromekelleher (approved):
This is amazing stuff @timothymillar, I'm not really able to review this in any detail but 
 LGTM, with a few very superficial comments

benjeffery (approved):
LGTM! I agree that specifying the samples to avoid detecting by NaNs is a good idea. Too brittle otherwise.

sgkit/stats/grm.py
timothymillar: The commented code was meant to be deleted. The lines are equivalent, but I originally had the less efficient version.

PR #1054, opened 2023-03-22
Store contig names in a `contig_id` array

### LINKED ISSUE #1035: Store contig names in a `contig_id` array
Currently these are stored in a dataset attribute, but it would be more consistent to store them as an array of dimension "contigs".

Storing contig names as an attribute can get messy when using a fragmented genome with thousands of scaffolds because, by default, the dataset attrs are printed in the repl representation.

Furthermore, storing them as an array would simplify some operations. E.g., in the GWAS tutorial

```python
ds["variant_contig_name"] = (["variants"], [ds.contigs[contig] for contig in ds["variant_contig"].values])
```

Could be simplified to

```python
ds["variant_contig_name"] = ds.contig_id[ds.variant_contig]
```

But these are minor issues, so I understand if others would rather avoid a breaking change.

tomwhite:
- [x] Fixes #1035
- [x] Tests added

### DIFF
--- a/sgkit/model.py
+++ b/sgkit/model.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Dict, Hashable, List, Optional
 
 import numpy as np
@@ -8,6 +9,7 @@ import sgkit as sg
 from .typing import ArrayLike
 from .utils import create_dataset
 
+DIM_CONTIG = "contigs"
 DIM_VARIANT = "variants"
 DIM_SAMPLE = "samples"
 DIM_PLOIDY = "ploidy"
@@ -61,6 +63,7 @@ def create_genotype_call_dataset(
     The dataset of genotype calls.
     """
     data_vars: Dict[Hashable, Any] = {
+        "contig_id": ([DIM_CONTIG], np.array(variant_contig_names, dtype=str)),
         "variant_contig": ([DIM_VARIANT], variant_contig),
         "variant_position": ([DIM_VARIANT], variant_position),
         "variant_allele": ([DIM_VARIANT, DIM_ALLELE], variant_allele),
@@ -139,6 +142,7 @@ def create_genotype_dosage_dataset(
 
     """
     data_vars: Dict[Hashable, Any] = {
+        "contig_id": ([DIM_CONTIG], np.array(variant_contig_names, dtype=str)),
         "variant_contig": ([DIM_VARIANT], variant_contig),
         "variant_position": ([DIM_VARIANT], variant_position),
         "variant_allele": ([DIM_VARIANT, DIM_ALLELE], variant_allele),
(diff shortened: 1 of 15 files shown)


sgkit/model.py
```diff
@@ -156,5 +159,24 @@ def create_genotype_dosage_dataset(
...
+    attrs: Dict[Hashable, Any] = {
+        "contigs": variant_contig_names,
+        "source": f"sgkit-{sg.__version__}",
+    }
     return create_dataset(data_vars=data_vars, attrs=attrs)
+
+
+def num_contigs(ds: xr.Dataset) -> ArrayLike:
+    """Return the number of contigs in a dataset."""
+    if "contigs" in ds.attrs:
```
timothymillar, line 171: Could we add a deprecation warning here? It will only be triggered by internal code but would still communicate the change to users.

tomwhite: Goode idea - I've added a warning in the latest update

sgkit/tests/io/vcf/test_vcf_reader.py
```diff
@@ -1528,5 +1531,21 @@ def test_vcf_to_zarr__retain_files(shared_datadir, tmp_path, retain_temp_files):
...
+
+    vcf_to_zarr(path, output, chunk_length=5, chunk_width=2)
+    ds = xr.open_zarr(output)
+
+    # drop new contig_id and filter_id variables
+    ds = ds.drop_vars(["contig_id", "filter_id"])
+
+    # check that contigs and filters can still be retrieved
+    assert num_contigs(ds) == 3
+    assert_array_equal(get_contigs(ds), np.array(["19", "20", "X"], dtype="S"))
```
benjeffery, line 1550: Could use `pytest.warns` here to check the warning.

benjeffery (approved):
LGTM, one thought on testing.

Have we thought about `load_dataset` inserting the new arrays as copies from the attribute? That would mean future code using `ds['contigs']` would work on old datasets.

tomwhite:
> Have we thought about `load_dataset` inserting the new arrays as copies from the attribute? That would mean future code using `ds['contigs']` would work on old datasets.

Something like this might work. I'm not sure what the best way to go though, since I'd like to eventually remove the attributes completely, to avoid duplication and long string representations.

sgkit/tests/io/vcf/test_vcf_reader.py
tomwhite: Fixed

timothymillar (approved):
LGTM!

benjeffery:
> > Have we thought about `load_dataset` inserting the new arrays as copies from the attribute? That would mean future code using `ds['contigs']` would work on old datasets.
> 
> Something like this might work. I'm not sure what the best way to go though, since I'd like to eventually remove the attributes completely, to avoid duplication and long string representations.

The code would only do the copy of the attributes to an array if it saw the attributes. And certainly not the other way around!

jeromekelleher (approved):
LGTM

PR #1055, opened 2023-03-29
Add extra comment about filters to write_vcf

benjeffery:
Hit this while writing a VCF from a simulated dataset, let me know if it is incorrect about filters coming from `vcf_to_zarr`!


tomwhite:
This shouldn't be needed once #1021 is in, since it takes account of the case when there are no filters in the dataset.

benjeffery:
> This shouldn't be needed once #1021 is in, since it takes account of the case when there are no filters in the dataset.

Ah, nice. Will close.

PR #1056, opened 2023-03-29
Add variant id to simulate dataset

benjeffery:
Came across this while trying to save a VCF of a simulated callset - seems like it should be in the simulated set?

### DIFF
--- a/sgkit/testing.py
+++ b/sgkit/testing.py
@@ -18,6 +18,7 @@ def simulate_genotype_call_dataset(
     seed: Optional[int] = 0,
     missing_pct: Optional[float] = None,
     phased: Optional[bool] = None,
+    additional_variant_fields: Optional[dict] = None,
 ) -> Dataset:
     """Simulate genotype calls and variant/sample data.
 
@@ -50,6 +51,9 @@ def simulate_genotype_call_dataset(
         The percentage of missing calls, must be within [0.0, 1.0], optional
     phased
         Whether genotypes are phased, default is unphased, optional
+    additional_variant_fields
+        Additional variant fields to add to the dataset as a dictionary of
+        {field_name: field_dtype}, optional
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


tomwhite:
This problem should be fixed by #1021.

benjeffery:
Should we also add `variant_quality` etc?

tomwhite:
Variant IDs and filters are not required variables, which is why the simulated dataset code doesn't set them. We could add them, but perhaps make them optional? Quality values aren't in bgen or plink datasets, only VCF, so making quality optional in the simulated dataset would make sense.

benjeffery:
Yeah, a set of optional additions to the simulation seems to make sense, will change this PR to that.

benjeffery:
Hows this? It allows arbitrary additional fields.

sgkit/tests/test_testing.py
```diff
@@ -29,3 +29,26 @@ def test_simulate_genotype_call_dataset__phased(tmp_path):
...
+        phased=True,
+        additional_variant_fields={
+            "variant_id": np.str,
+            "variant_filter": np.bool,
+            "variant_quality": np.int8,
+            "variant_yummyness": np.float32,
+        },
+    )
+    assert "variant_id" in ds
+    print(ds["variant_id"].dtype)
```
tomwhite, line 47: This print should be removed

tomwhite (approved):
LGTM. I think the error case needs testing to meet the coverage target.

sgkit/tests/test_testing.py
benjeffery: Whoops, fixed!

benjeffery:
All fixed up and good to go.

benjeffery:
@tomwhite I have a `vcf_to_zarr` failure here which isn't related to my changes I think.

sgkit/testing.py
```diff
@@ -96,3 +101,18 @@ def simulate_genotype_call_dataset(
...
+            if field_dtype in (np.float32, np.float64):
+                field = rs.rand(n_variant).astype(field_dtype)
+            elif field_dtype in (np.int8, np.int16, np.int32, np.int64):
+                field = rs.randint(0, 100, n_variant, dtype=field_dtype)
+            elif field_dtype is np.bool:
+                field = rs.rand(n_variant) > 0.5
+            elif field_dtype is np.str:
+                field = np.arange(n_variant).astype("S")
+            else:
+                raise ValueError(f"Unrecognized dtype {field_dtype}")
```
tomwhite, line 116: This line needs a test to meet the 100% coverage requirement

tomwhite (approved):

PR #1063, opened 2023-04-16
Use Codecov GitHub action since the PyPi package has been removed

tomwhite:
The build is breaking due to this (e.g. https://github.com/pystatgen/sgkit/actions/runs/4710629116).

See https://about.codecov.io/blog/message-regarding-the-pypi-package/ for context.


jeromekelleher (approved):
Lgtm 

Looks like a straightforward fix to duplicate on tskit repos @benjeffery ?

PR #1069, opened 2023-04-19
Fix handling of VCFs with no samples 

### LINKED ISSUE #1068: VCF with zero sample columns results in divide by zero error
As discovered in #1067 running `vcf_to_zarr` with no sample columns in the VCF results in 
```
  File "/home/benj/projects/sgkit/env/lib/python3.10/site-packages/zarr/indexing.py", line 160, in ceildiv
    return math.ceil(a / b)
ZeroDivisionError: division by zero
```

A simple fix is to disable chunking if a zero-length array is detected in `concat_zarrs`. I'll write up a PR.

benjeffery:
- [x] Fixes #1068
- [X] Tests added
- [X] User visible changes (including notable bug fixes) are documented in `changelog.rst`


tomwhite (approved):
Thanks for fixing this @benjeffery! It looks good to me.

The `test_vcf_to_zarr__no_genotypes` test is also testing a VCF file with no samples (`no_genotypes.vcf`), but not the parallel code path, which yours does and which is where the error lies. Is there any point in combining them in some way or do you think they are OK as they are?

benjeffery:
Good spot, i had a search for an existing test but didn't spot that one.
I think it's worth keeping both, I couldn't just add a parallel path to the existing test as the file didn't have an index and worth keeping some in plaintext. It's also not a bad thing to have one more "found in the wild" VCF to test on.

Issue #463, opened 2021-02-10
Replicating Hail's GWAS Tutorial


hammer:
I decided to work on https://github.com/pystatgen/sgkit/issues/88 by trying to implement the [Hail GWAS Tutorial](https://hail.is/docs/0.2/tutorials/01-genome-wide-association-study.html) with `sgkit`.

I'll update this issue with my experiences. I barely know what I'm doing, so this should be fun.

hammer:
Preamble
```
import xarray as xr
import sgkit as sg
```

[Importing data from VCF](https://hail.is/docs/0.2/tutorials/01-genome-wide-association-study.html#Importing-data-from-VCF)
```
sg.io.vcf.vcf_to_zarr("https://storage.googleapis.com/hail-tutorial/1kg.vcf.bgz", "1kg.zarr")
ds = xr.open_zarr("1kg.zarr")
```

[Getting to know our data](https://hail.is/docs/0.2/tutorials/01-genome-wide-association-study.html#Getting-to-know-our-data)

It's not so easy for us to mimic `mt.rows().select().show(5)` . Perhaps we should consider making a data frame out of the `variant_` data arrays?

Here's an ugly thing
```
from itertools import starmap

var_head = ds.isel(variants=slice(None,5))
locus = list(starmap(lambda chr, pos: str(chr) + ':' + str(pos), list(zip(var_head.variant_contig.values, var_head.variant_position.values))))
list(zip(locus, var_head.variant_allele.values))
```

We don't seem to be picking up contig metadata, and we confusingly zero index our contigs so that `set(ds.variant_contig.values)` runs from 0 to 22.

tomwhite:
> Note that sometimes I see an error when using `sg.io.vcf.vcf_to_zarr` that `sg.io.vcf` doesn't exist, but `from sgkit.io.vcf import vcf_to_zarr` always works.

The `vcf_to_zarr` function is not imported in the top-level `sgkit` namespace (i.e. by `sgkit/__init__.py`), so `sg.io.vcf.vcf_to_zarr` will give an error after `import sgkit as sg`. However, it will work if you have run `from sgkit.io.vcf import vcf_to_zarr` too, which is confusing.

For the time being, I would suggest you use `from sgkit.io.vcf import vcf_to_zarr` - until we consolidate the "extra" packages into one.

[...]

PR #1072, opened 2023-04-24
Add tutorial on relationship matrices

### LINKED ISSUE #956: Example notebook exercising polyploidy and pedigree work

timothymillar:
- [x] Fixes #956 (doesn't cover polyploidy)

It's still a bit rough, but I think the key parts are there. This covers working with pedigree data, calculating relationship matrices from pedigree and molecular data, and combining them into a hybrid relationship matrix. It also showcases the flexibility of having sgkit based on xarray.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,12 @@ jobs:
       if: failure()
       with:
         name: gwas_tutorial
-        path: /home/runner/work/sgkit/sgkit/docs/_build/html/reports/gwas_tutorial.log
+        path: /home/runner/work/sgkit/sgkit/docs/_build/html/reports/examples/gwas_tutorial.err.log
+    - uses: actions/upload-artifact@v2
+      if: failure()
+      with:
+        name: relatedness_tutorial
+        path: /home/runner/work/sgkit/sgkit/docs/_build/html/reports/examples/relatedness_tutorial.err.log
     - name: Test with pytest
       run: |
         pytest -v
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -59,6 +59,12 @@ extensions = [
     *[p.stem for p in (HERE / "extensions").glob("*.py")],
 ]
 
+# Enable math notation in mystnb
+myst_enable_extensions = [
+    "amsmath",
+    "dollarmath",
+]
+
 extlinks = {
     "issue": ("https://github.com/pystatgen/sgkit/issues/%s", "GH %s"),
     "pr": ("https://github.com/pystatgen/sgkit/pull/%s", "PR %s"),
--- a/docs/examples/index.rst
+++ b/docs/examples/index.rst
@@ -10,3 +10,4 @@ Example notebooks showing how to use sgkit.
    :hidden:
 
    gwas_tutorial
+   relatedness_tutorial
--- a/validation/gwas/method/pc_relate/README.md
+++ b/validation/gwas/method/pc_relate/README.md
@@ -5,13 +5,11 @@ HapMap data.
 This code is scheduled as part of the Github Actions CI.
 
 To run manually, you need to first download the test data
-from `gs://sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip`,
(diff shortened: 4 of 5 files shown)


tomwhite (commented):
This looks like a great addition.

I read through the text and found a few minor typos:

> It is also common to use a 0 to indicate an unknown parent pedigree files.

Change to "an unknown parent in pedigree files".

> to ensure they are note mistaken

"not"

> dimensions named "samples" and "parent"

"parents"

> Using out pedigree dataset

"our"

timothymillar:
Thanks!

timothymillar:
I think this is getting close now, but CI is failing due to a timeout, possibly the data download? `/home/runner/work/sgkit/sgkit/docs/examples/relatedness_tutorial.ipynb: WARNING: Executing notebook failed: CellTimeoutError [mystnb.exec]`.

I'm also having some issue with building locally as the equations in markdown cells are not being formatted correctly e.g.:
Capture

timothymillar:
@tomwhite I was pulling the data from PMC supplementary material (not ideal). I can cut the data down to a pair of files as 1.1MB and 0.3MB. Should we upload these to the same bucket as the GWAS example?

hammer:
Maybe? Let me check!

hammer:
I do not see that bucket on my personal or Hammer Lab account. I may have created it with my Related Sciences account but I can no longer access that. @eric-czech do you happen to see that bucket on any RS-owned accounts?

eric-czech:
> did you create the gs://sgkit-gwas-tutorial bucket for https://github.com/pystatgen/sgkit/issues/463?

Might have been me too but either way, I'd rather use the project and bucket we have specifically for sgkit.  I added you all (@tomwhite, @jeromekelleher, @hammer, @timothymillar) as storage object admins for https://console.cloud.google.com/storage/browser/sgkit-data.  There is a `validation` directory in there used by this [workflow](https://github.com/pystatgen/sgkit/blob/a140316e0f493da25ce9891b9277ea4689321dc0/.github/workflows/validation.yml#L22) so I think a `tutorial` directory at the top level would make sense.  @hammer and @tomwhite I also added you as owners on the GCP project so I believe you should be able to add/remove people from GCS IAM roles in the future.

@timothymillar can you upload this to `gs://sgkit-data/tutorial/<whatever_you_like>`? Let me know if you get some kind of permissions error.

timothymillar:
Thanks @eric-czech, I was able to create the `tutorial` dir through the GUI but failing to upload data with a generic "Error".

eric-czech:
> Edit: Ahh, it's because the whole bucket is set up as "requester pays" which I think means it won't work for tutorials anyway

Ok, I disabled that.  Try again?  Everybody I mentioned above should have full control over the bucket now.

timothymillar:
Thanks, I've uploaded the data now. The only remaining issue is that the bucket grants read access to "allAuthenticatedUsers" rather than "allUsers":

> One or more bucket-level permissions grant access to everyone on the internet (allUsers) or anyone signed into a Google account (allAuthenticatedUsers).

This means a user can't just grab the data with the URL only. I should be able to make this change but thought I should check with whoever is paying the bills! This would also apply to the validation data which is a bit more substantial. This also outdates some of the [documentation](https://github.com/pystatgen/sgkit/blob/main/validation/gwas/method/pc_relate/run.sh#L14)

eric-czech:
> The only remaining issue is that the bucket grants read access to "allAuthenticatedUsers" rather than "allUsers"

Ah, didn't realize that.  It's "allUsers" now so there should be no more restrictions.

> This also outdates some of the [documentation](https://github.com/pystatgen/sgkit/blob/main/validation/gwas/method/pc_relate/run.sh#L14)

Good catch!  That last line could be removed and it would be better if that path to the file was `https://storage.googleapis.com/sgkit-data/validation/hapmap_JPT_CHB_r23a_filtered.zip` instead.

timothymillar:
I think this is ready to go unless there are any suggestions? I imagine it'll get updated as relevant features are added.

tomwhite (approved):

Issue #1059, opened 2023-04-05
sgkit.io.vcf.vcf_reader.vcf_to_zarr() fails to convert VCFs with INFO/CSQ and other unbounded annotations


tnguyengel:
`sgkit.io.vcf.vcf_reader.vcf_to_zarr()`   fails to convert VCFs with INFO/CSQ annotations with error:

`ValueError: INFO field 'CSQ' is defined as Number '.', which is not supported.`

Presumably, the method will also fail for any VCFs containing annotations with unbounded size.  INFO/CSQ contains variant effect predictions from [VEP](https://www.ensembl.org/info/docs/tools/vep/index.html).   There can be multiple predictions for each allele, one for every transcript that an allele overlaps.  Each prediction is separated by a comma.  The number of predictions per allele is not known in  advance, and so the INFO/CSQ field is defined with unbounded size in the header, or "Number=.":

For example:

`##INFO=<ID=CSQ,Number=.,Type=String,Description="Consequence annotations from Ensembl VEP. Format: Allele|Consequence|...>`

It would be very useful to be able to filter a zarr for variants that are deemed clinically relevant according to annotation, such as loss of function variants.

Do you suggest any workarounds in the meantime?

timothymillar:
Hi @tnguyengel, I'm not familiar INFO/CSQ annotations but you can specify a maximum bound for a VCF field using the `field_defs` parameter.

```python
from sgkit.io.vcf import vcf_to_zarr
vcf_to_zarr(
    "example.vcf.gz",
    "example.zarr",
    fields=["INFO/CSQ "],
    field_defs={"INFO/CSQ ": {"Number": 100}},  # or whatever a reasonable maximum is.
)
```

Essentially, the `vcf_to_zarr` function needs to know the maximum size of each field so that it can allocate space. If the field is unbounded ('.'), then you will need to manually provide an upper bound.

tnguyengel:
Thanks!  That sounds like a quick workaround!  

From brief googling, a human gene has ~4 transcripts on average, but can have up to ~1000 transcripts.  If this is correct, in order to ensure that the INFO/CSQ field is not truncated in the zarr, I would need to set:

`field_defs={"INFO/CSQ ": {"Number": 1000}}`

[...]

Issue #634, opened 2021-07-19
Investigate ragged arrays to represent alleles


tomwhite:
Alleles are a challenge to represent efficiently in fixed-length arrays. There are a couple of problems:
1. the number of alleles is not known until the whole VCF file has been processed
2. there can be a very wide variation in the number of alt alleles (most variants will have one, but a few could have thousands

Both these problems could be solved by using ragged arrays.

Zarr has support for [ragged arrays](https://zarr.readthedocs.io/en/stable/tutorial.html#ragged-arrays), but these don't currently work with variable length strings (needed for alleles), and they don't fit the Xarray data model, which assumes fixed sized dimensions. There is a good discussion of the problem in https://github.com/pydata/xarray/issues/4285, in the context of Awkward Array.

jeromekelleher:
Tricky one. I guess an option we should consider also is a home-grown ragged array encoding for alleles (although it would be pretty icky to expose users to this). Could we (in principle) do an encoding ourselves and expose the ``alleles`` dataset variable as a numpy object array (or similar)?

tomwhite:
> Could we (in principle) do an encoding ourselves and expose the `alleles` dataset variable as a numpy object array (or similar)?

Yes, I think it would be worth trying this out to see if it's possible, and how it works in Xarray.

jeromekelleher:
FWIW we've been doing this in tskit like [this](https://tskit.dev/tskit/docs/stable/data-model.html#encoding-ragged-columns). I think the underlying encoding (a data and offset array) is basically the same as what Arrow uses, and it works well in practise.

ravwojdyla:
FYI @tomwhite. Following today's "standup", "open sourcing" some of our internal comments about awkward array, these are likely outdated (they were made about a year ago) and might not fully apply in the sgkit context. It might be still useful tho.

From @eric-czech:

There aren't any docs for [awkward-1.0](https://github.com/scikit-hep/awkward-1.0) yet, just empty stubs at https://awkward-array.org, so I looked at the original project instead.  Here are a few questions I wanted to answer:

- How do we parallelize with awkward?

It looks like dask.Bag (or even just delayed) wrapping awkward would make the most sense.  I was hoping initially that it would be possible for awkward to act on dask arrays, but it just converts everything to numpy afaict:

```python
import pyarrow, awkward
import dask.array as da
arrow_buffer = pyarrow.ipc.open_file(open("tests/samples/exoplanets.arrow", "rb")).get_batch(0)
[...]

PR #1083, opened 2023-05-04
Add experimental `vcf_to_parquet` function

tomwhite:
- [ ] Fixes #xxxx
- [x] Tests added
- [ ] New functions are listed in `api.rst`

Inspired by the discussion in #1059, I've written some code to export the fixed fields of a VCF file (including INFO fields) to a collection of Parquet files. The idea is that you can run filtering-style queries on this variant metadata, and use that to subset/mask the variants in a Zarr dataset. I explored this a bit in https://github.com/pystatgen/sgkit/issues/634#issuecomment-1523285115.

This is not to be merged, just posting for interest.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ zarr >= 2.10.0, != 2.11.0, != 2.11.1, != 2.11.2
 fsspec != 2021.6.*
 scikit-learn
 pandas
+pyarrow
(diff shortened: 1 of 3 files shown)


tomwhite:
Closing. In https://github.com/pystatgen/sgkit/discussions/1091 we decided that for VEP use case it would be best to recommend users export to JSON and use that to construct filters on the Zarr data.

PR #1087, opened 2023-05-10
Add pedigree_sel method

### LINKED ISSUE #1033: Subsetting a pedigree dataset
While working on #956 I've found it quite annoying to take a subset of pedigree dataset (containing the `sample_id`, `parent_id` and maybe `parent` arrays). The pain points are:

- replacing any parent ids that have been removed from the dataset with the missing value '.'
- removing/recalculating the now invalidated `parent` array
- remembering to apply the same select statement to the `samples_0` and `samples_1` dimensions for relationship matrices

The basics of this are shown here:
```python
ds  # dataset with pedigree arrays and relationship matrices
keep  # array of samples I want to keep

# select using xarray method
ds2 = ds.sel(dict(
    samples=keep,
    samples_0=keep,
    samples_1=keep,
))

# replace removed parents with missing symbol
remaining = set(keep)
ds2.parent_id.data = [[s if s in remaining else '.' for s in row] for row in ds2.parent_id.values]

# remove the parent array with outdated indices
ds2.drop_var("parent")
```

I have also been intending to add a method to subset pedigree data based on structure. I.e., ask questions like:

- Select sample "S521" and all of its ancestors
- Select samples "S3", "S7" and "S9" and all descendants within x generations

An API covering all of the above could look something like:
```python
def pedigree_sel(
    ds,
    *,
    samples: Arraylike,
    parent: Hashable = variables.parent,  # computed if absent
    ancestor_depth: int = 0,
    descendant_depth: int = 0, 
    drop_parent_array: bool = True,  # drop invalidated parent array by default
    update_parent_id: bool = True,  # replace any removed parent_ids with "."
    sel_samples_0: bool = True,  # apply the select to samples_0 dimension if present
    sel_samples_1: bool = True,  # apply the select to samples_1 dimension if present
    merge: bool = True,
)
```

timothymillar:
- [x] Fixes #1033
- [x] Tests added

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -137,6 +137,7 @@ Utilities
    infer_variant_ploidy
    invert_relationship_matrix
    parent_indices
+   pedigree_sel
    simulate_genotype_call_dataset
    window_by_genome
    window_by_interval
--- a/sgkit/__init__.py
+++ b/sgkit/__init__.py
@@ -45,6 +45,7 @@ from .stats.pedigree import (
     pedigree_inbreeding,
     pedigree_inverse_kinship,
     pedigree_kinship,
+    pedigree_sel,
 )
 from .stats.popgen import (
     Fst,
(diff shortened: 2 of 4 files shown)


jeromekelleher (approved):
Wow, this is cool, nice :+1:

PR #1093, opened 2023-05-29
Pedigree contribution

### LINKED ISSUE #963: Expected founder/ancestor contributions from pedigree structure
Expected founder contribution is the proportion of alleles within an individual/cohort which are expected to to have been derived from a given founder based upon pedigree structure. This is a relatively strait-forward metric to estimate and is quite a useful for understanding the structure of complex pedigrees.

I think we could could generalize the idea slightly to produce a samples * samples matrix `M` where `M[i, j]` is the expected contribution of sample `i` to sample `j`. This would be a non-symmetrical matrix containing expected contributions between all pairs of samples (diagonals = 1). Founder contributions to a specific individual/cohort can then be found by selecting founders along the first axis and advance individuals/cohorts along the second axis e.g.: `ds.pedigree_contribution.loc[<founders>, <advanced>]`.

timothymillar:
- [x] Fixes #963
- [x] Tests added

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -112,6 +112,7 @@ Methods
    maximal_independent_set
    observed_heterozygosity
    pbs
+   pedigree_contribution
    pedigree_inbreeding
    pedigree_inverse_kinship
    pedigree_kinship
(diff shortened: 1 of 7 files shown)


timothymillar:
I'm not 100% about the chunking API, but I think it's better to require explicit chunks from the user in this instance.

tomwhite:
> I'm not 100% about the chunking API, but I think it's better to require explicit chunks from the user in this instance.

Can you say a bit more about this? It looks like the default is not to chunk (which is fine), but the other option is to chunk in either of the samples dimensions `samples_0` or `samples_1` (but not both). Presumably you would do the latter if you had a lot of samples, but it's not clear to me whether there is a reason to prefer `samples_0` or `samples_1`.

Overall LGTM.

tomwhite:
Thanks for the explanation @timothymillar! This sounds like a good approach to me.

tomwhite (approved):

Issue #1109, opened 2023-08-14
Windows build is failing


tomwhite:
From https://github.com/pystatgen/sgkit/actions/runs/5850622797/job/15860289990#step:5:4585

```
============================= test session starts =============================
platform win32 -- Python 3.8.17, pytest-7.4.0, pluggy-1.2.0 -- C:\Miniconda3\envs\test\python.exe
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('D:\\a\\sgkit\\sgkit\\.hypothesis\\examples')
rootdir: D:\a\sgkit\sgkit
configfile: setup.cfg
plugins: hypothesis-6.82.4, cov-4.1.0, datadir-1.4.1, mock-3.11.1
collecting ... collected [10](https://github.com/pystatgen/sgkit/actions/runs/5633732790/job/15262949106#step:5:11) items / [12](https://github.com/pystatgen/sgkit/actions/runs/5633732790/job/15262949106#step:5:13)7 errors

=================================== ERRORS ====================================
_____________________ ERROR collecting sgkit/__init__.py ______________________
C:\Miniconda3\envs\test\lib\site-packages\_pytest\runner.py:341: in from_call
    result: Optional[TResult] = func()
C:\Miniconda3\envs\test\lib\site-packages\_pytest\runner.py:372: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
C:\Miniconda3\envs\test\lib\site-packages\_pytest\doctest.py:547: in collect
    module = import_path(
C:\Miniconda3\envs\test\lib\site-packages\_pytest\pathlib.py:565: in import_path
    importlib.import_module(module_name)
C:\Miniconda3\envs\test\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:10[14](https://github.com/pystatgen/sgkit/actions/runs/5633732790/job/15262949106#step:5:15): in _gcd_import
    ???
<frozen importlib._bootstrap>:991: in _find_and_load
    ???
<frozen importlib._bootstrap>:975: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:671: in _load_unlocked
    ???
<frozen importlib._bootstrap_external>:843: in exec_module
    ???
[...]

PR #1111, opened 2023-08-14
Update minimum Dask dependency versions

tomwhite:
- [ ] Fixes #xxxx
- [ ] Tests added
- [ ] New functions are listed in `api.rst`

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -28,10 +28,8 @@ jobs:
     - name: Install dependencies
       # activate conda
       shell: bash -l {0}
-      # conda can't install all dev tools, so we need to split it between conda and pip
       run: |
-        conda install --file requirements.txt msprime
-        pip install -r requirements-dev.txt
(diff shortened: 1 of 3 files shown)


tomwhite:
Restricting to `pandas < 2` fixes the problem on Windows, but I'm not sure what the underlying problem is. It would be nice to know what that is to see if it could be fixed in a less restrictive way.

tomwhite:
I haven't been able to diagnose the underlying problem, so I tried to only add the `pandas < 2` restriction to Windows with the following in requirements.txt:

```
pandas; platform_system != "Windows"
pandas < 2; platform_system == "Windows"
```

This is valid for Pip, but not Conda, which we use for (part of) the Windows installation, since msprime wheels are not available on Windows.

I'm not sure how to make progress with this.

tomwhite:
BTW this PR is to fix #1109

tomwhite:
Closed in favour of #1122

PR #1119, opened 2023-08-30
Improve performance of `variant_stats`

### LINKED ISSUE #1116: Compute variables in variant_stats concurrently
Doing downstream calculations with variables returned by ``variant_stats`` is quite slow (see here: https://github.com/pystatgen/sgkit-publication/issues/34). It seems likely that this is mostly due to requiring us go through the whole variant matrix to compute it for each stat. While this is fine for a single variable, it adds up quickly if you want to look at the number of hets, homs, allele counts etc.

The variables are defined as Dask arrays here using numpy operations:

https://github.com/pystatgen/sgkit/blob/cc048581be1b44be7208634107877f3512f29823/sgkit/stats/aggregation.py#L435

It would be much more efficient to compute all these values at once, in a single pass through the genotype matrix.  Would it make sense to add an option to this (and I guess ``sample_stat``) to do this? We could call it either ``eager`` or ``lazy`` --- since things are lazy by default, I guess it makes sense to call the parameter ``eager`` to make it stand out?

In terms of implementation, I guess this would need to be a numba gufunc? The code should be straightforward.

timothymillar:
- [x] Fixes #1116
- [x] Tests added
- [x] Update the relate `sample_stats` function
- [x] Update Tajimas_D to use faster  `count_variant_alleles`

This updates `variant_stats` to use a pair of gufuncs to calculate variant counts and hom/het genotype counts respectively.
This seems to be the sweet spot for performance, It may be possible to use a single gufunc but I can't seem to maintain the same performance (probably due to the complexity of the inner loop). The two gufunc soultion is also nice for modularity. I also looked at using [`da.apply_gufunc`](https://docs.dask.org/en/stable/generated/dask.array.gufunc.apply_gufunc.html) to have multiple return values but the performance seems to be worse than `map_blocks` with a single return value.

This also changes the `count_variant_alleles` function to give the option of counting alleles from the `call_genotype` array directly rather than via the `call_allele_count` array. This introduces the `using` argument to specify the array to use. Feedback on this approach/argument name would be appreciated as this pattern could be used elsewhere. I have kept the old behavior as the default for now.

I seem to be getting around an 18-22x speedup compared to the current master branch. This is variable and seems to increase with numbers of samples which makes sense given the changes to `count_variant_alleles` (no longer producing intermediate counts for each sample). It'd be good if someone else can verify this as it's better than I expected.

Relative speedup was calculated using biallelic diploid data with 1% missing allele calls. The variants array was split into 8 chunks in all cases to ensure CPU saturation on an 8 core machine.

Relative speedup:

|  Threads |   100k * 1k |  20k * 5k |   10k * 10k |
|----------:|------------:|-----------:|------------:|
|         1 |        21.1 |       22   |        23.4 |
|         2 |        19.5 |       19.7 |        19.8 |
|         4 |        18.6 |       18.9 |        19.1 |
|         6 |        19.2 |       19.8 |        20.5 |
|         8 |        17.8 |       18.6 |        18.7 |

The first value in each column name is the number of variants and the second is the number samples.

Note: I'm using `dask.config.set(pool=ThreadPool(<number>))` to control thread number.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,6 +19,8 @@ New Features
   (:user:`timothymillar`, :pr:`1100`, :issue:`1062`)
 - Add :func:`display_pedigree` function.
   (:user:`timothymillar`, :pr:`1104`, :issue:`1097`)
+- Add option to count variant alleles directly from call genotypes in function :func:`count_variant_alleles`.
+  (:user:`timothymillar`, :pr:`1119`, :issue:`1116`)
 
 .. Breaking changes
 .. ~~~~~~~~~~~~~~~~
@@ -26,8 +28,11 @@ New Features
 .. Deprecations
 .. ~~~~~~~~~~~~
 
-.. Improvements
-.. ~~~~~~~~~~~~
+Improvements
+~~~~~~~~~~~~
+
+- Improve performance of :func:`variant_stats` and :func:`sample_stats` functions.
+  (:user:`timothymillar`, :pr:`1119`, :issue:`1116`)
 
 .. Bug fixes
 .. ~~~~~~~~~
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -203,7 +203,7 @@ shows how it can be used in the context of doing something simple like counting 
 
     # Now the result is correct -- only the third sample is heterozygous so the count should be 1.
     # This how many sgkit functions handle missing data internally:
-    sg.variant_stats(ds).variant_n_het.item(0)
+    sg.variant_stats(ds).variant_n_het.values.item(0)
 
 Windowing
 ---------
@@ -320,8 +320,8 @@ Xarray and Pandas operations in a single pipeline:
     # for windows of size 20 variants
     (
         ds
-        # Add call rate and other statistics
-        .pipe(sg.variant_stats)
+        # Add and compute call rate and other statistics
+        .pipe(sg.variant_stats).compute()
         # Apply filter to include variants present across > 80% of samples
         .pipe(lambda ds: ds.sel(variants=ds.variant_call_rate > .8))
         # Create windows of size 20 variants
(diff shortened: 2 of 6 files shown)


jeromekelleher:
Amazing, thanks @timothymillar! I'll try it out on my benchmark tomorrow and report back.

timothymillar:
I've updated the `sample_stats` function to use the same gufunction. This gives a bit of a speed up but not as much as seen for `variant_stats`.

jeromekelleher:
I've tried running this by installing to my local setup @timothymillar, but I'm consistently getting a segfault on running the new code. I've tried clearing out the numba cache, but that doesn't seem to help. Any ideas what's going on?

jeromekelleher:
This is a massive improvement, thanks @timothymillar! We're looking at more than 10X speedup in my benchmark.

timothymillar:
@jeromekelleher my guess is that you've run into https://github.com/numba/numba/issues/4807 which was the main motivation for #869. Do you get the same issues with other gufuncs when cached in parallel?

In practice, I've only run into this issue using gufuncs in parallel (via dask or mp) on our cluster, so I wondered if the NFS may increase the likelihood. I can also get segfaults on my local machine using the self-contained POC in that numba issue. I don't know why I only run into it using gufuncs, but based on that issue others have also had similar issue with `@njit(parallel=True)`.

I've been meaning to follow up on it for some time but wanted to have a well parameterized example before adding to that thread.

jeromekelleher:
Can you update this please @timothymillar, windows stuff should be fixed now.

timothymillar:
Note that I had to make some minor adjustments to the "getting started" docs because `variant_stats` now always returns a dask array, even if the input is a numpy array. I think this is in line with what we have done elsewhere.

jeromekelleher (approved):
LGTM!

tomwhite (approved):
LGTM

Issue #280, opened 2020-09-24
Apply to NumFOCUS by October 15


hammer:
A reminder to myself from a previous [discussion](https://discourse.pystatgen.org/t/what-should-we-call-our-new-genetics-toolkit/52/11?u=hammer) in June.

I think we may want a logo too #30

hammer:
Well I just spent about 30 minutes filling out their online application and then when I finished was sent to a page saying "page not found" so now I have no idea if the application went in. I'll check with them to be sure.

hammer:
Ah just got the confirmation email, we're good.

PR #1129, opened 2023-09-20
Fix memory leak (see cyvcf2 #280)

benjeffery:
Calling`vcf.close` leaks memory, the good news is that we don't have to! `vcf.__dealloc__` calls it anyway.

See https://github.com/brentp/cyvcf2/pull/280

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -100,10 +100,11 @@ class MaxAltAllelesExceededWarning(UserWarning):
 def open_vcf(path: PathType) -> Iterator[VCF]:
     """A context manager for opening a VCF file."""
     vcf = VCF(path)
-    try:
-        yield vcf
-    finally:
-        vcf.close()
+    yield vcf
+    # This line is commented out as it causes a memory leak in cyvcf2
+    # See https://github.com/brentp/cyvcf2/pull/280
+    # We should reinstate this after a new release of cyvcf2
+    # vcf.close()
 
 
 def region_filter(


jeromekelleher:
Fantastic, great job tracking this down @benjeffery!

I guess we can just pin on cyvcf>=0.30.23 (since Brent has released it) rather than removing the ``close`` here?

benjeffery:
> I guess we can just pin on cyvcf>=0.30.23 (since Brent has released it) rather than removing the `close` here?

My understanding is that only the conda build is working, so that release won't be on PyPI yet. The close isn't necessary as dealloc will be called when the context manager closes. I've tested that using this style of opening and closing a VCF doesn't result in memory growth.

benjeffery:
Oops I fat fingered the close button

tomwhite:
Great sleuthing @benjeffery!

jeromekelleher:
If we're going to merge this rather than wait for a bugfix update from cyvcf then we definitely need some comments explaining why we're not doing the standard thing of calling "close".

I can also easily see a case where we start leaking memory in the future because cyvcf changes semantics. 

What's your take @tomwhite?

tomwhite:
This was a really nasty bug because usually memory leaks happen because the calling code neglects to call close. I've never seen a case where calling close _blocks_ the clean up of resources!

> I can also easily see a case where we start leaking memory in the future because cyvcf changes semantics.

I don't think cyvcf2 is going to change semantics. It was a bug that has now been fixed. I think it's unlikely there will be a regression that causes the same bug.

I would merge this (with a comment) and then revert it in the future when cyvcf2 has a release (wheel builds are currently not working for non-obvious reasons).

benjeffery:
I've pushed a new commit which has an explanation of the situation.

benjeffery:
As cyvcf2 has released the fix, I don't think it is worth merging this one.

PR #1132, opened 2023-09-21
Skip parts of a VCF that have already been done

benjeffery:
An attempt to fix https://github.com/pystatgen/sgkit-publication/issues/35 by having the VCF chunk parse be skipped if dask is running it more than once due to dying workers.

### DIFF
--- a/sgkit/io/vcf/vcf_reader.py
+++ b/sgkit/io/vcf/vcf_reader.py
@@ -429,7 +429,10 @@ def vcf_to_zarr_sequential(
     exclude_fields: Optional[Sequence[str]] = None,
     field_defs: Optional[Dict[str, Dict[str, Any]]] = None,
     read_chunk_length: Optional[int] = None,
+    check_if_done: bool = False,
 ) -> None:
+    if check_if_done and "_sgkit_parse_done" in output:
+        return
     if read_chunk_length is None:
         read_chunk_length = chunk_length
     with open_vcf(input) as vcf:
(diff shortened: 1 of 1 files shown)


sgkit/io/vcf/vcf_reader.py
```diff
@@ -429,7 +429,10 @@ def vcf_to_zarr_sequential(
     exclude_fields: Optional[Sequence[str]] = None,
     field_defs: Optional[Dict[str, Dict[str, Any]]] = None,
     read_chunk_length: Optional[int] = None,
+    check_if_done: bool = False,
 ) -> None:
+    if check_if_done and "_sgkit_parse_done" in output:
```
tomwhite, line 434: This would fail if `output` was a path. Saving the flag in the mutable mapping seems a bit odd - I would expect it to be in the Zarr attributes.

sgkit/io/vcf/vcf_reader.py
```diff
@@ -792,6 +796,10 @@ def vcf_to_zarrs(
                 fields=fields,
                 exclude_fields=exclude_fields,
                 field_defs=field_defs,
+                # If a worker dies, dask will redo the tasks on that worker,
+                # which is unnecessary, with this flag the parse will be skipped
+                # if it has already been done.
+                check_if_done=True,
```
tomwhite, line 802: Should this parameter be propagated all the way back to the public API so that the user can set it?

benjeffery:
Closing in favour of #1138

PR #1139, opened 2023-10-12
Add gene-drop simulation method

### LINKED ISSUE #1107: Gene-drop estimation of genotype probabilities
A gene-drop simulation treats a pedigree as a stochastic model from which alleles/genotypes are drawn at random following (simplified) Mendelian inheritance. These simulations typically don't model linkage among loci and are used to approximate expected genotype distributions per locus.

I already have a rough implementation in numba, however there's multiple ways in which the results could be reported. The most general way to store the results of a simulation is as a call-genotype array per replicate, but this can result in a massive array (replicates * variants * samples * ploidy). It's much more efficient to summaries the replicates on the fly and there are a few options here. An obvious choice is to store the distribution of possible genotypes for each call (variants * samples * genotypes). This captures most of the information about individual calls. We could also look at reporting the probability of the observed call genotype which would be simpler (variants * samples).

timothymillar:
- [x] Fixes #1107
- [x] Tests added

### DIFF
--- a/sgkit/stats/genedrop.py
+++ b/sgkit/stats/genedrop.py
@@ -1,0 +1,345 @@
+from typing import Hashable, Optional
+
+import dask.array as da
+import numpy as np
+import xarray as xr
+from typing_extensions import Literal
+from xarray import Dataset
+
+from sgkit import variables
+from sgkit.accelerate import numba_guvectorize, numba_jit
+from sgkit.typing import ArrayLike
+from sgkit.utils import (
+    conditional_merge_datasets,
+    create_dataset,
+    define_variable_if_absent,
+)
+
+from .pedigree import (
+    _compress_hamilton_kerr_parameters,
+    parent_indices,
+    topological_argsort,
+)
+
+EST_DIPLOID = "diploid"
+EST_HAMILTON_KERR = "Hamilton-Kerr"
+
+
+@numba_guvectorize(  # type: ignore
+    [
+        "void(int8[:,:], int64[:,:], uint32, int8[:,:])",
+        "void(int16[:,:], int64[:,:], uint32, int16[:,:])",
+        "void(int32[:,:], int64[:,:], uint32, int32[:,:])",
+        "void(int64[:,:], int64[:,:], uint32, int64[:,:])",
+    ],
+    "(n,k),(n,p),()->(n,k)",
+)
+def genedrop_diploid(
+    genotypes: ArrayLike,
+    parent: ArrayLike,
+    seed: ArrayLike,
+    out: ArrayLike,
+) -> None:  # pragma: no cover
+    n_sample, n_parent = parent.shape
+    _, ploidy = genotypes.shape
+    if n_parent != 2:
+        raise ValueError("The parents dimension must be length 2")
+    if ploidy != 2:
+        raise ValueError("Genotypes are not diploid")
+    order = topological_argsort(parent)
+    np.random.seed(seed)
+    for i in range(n_sample):
+        t = order[i]
+        unknown_parent = 0
+        for j in range(n_parent):
+            p = parent[t, j]
+            if p < 0:
+                # founder
+                unknown_parent += 1
+            else:
+                idx = np.random.randint(2)
+                out[t, j] = out[p, idx]
+        if unknown_parent == 1:
+            raise ValueError("Pedigree contains half-founders")
+        elif unknown_parent == 2:
+            # copy founder
+            out[t, 0] = genotypes[t, 0]
+            out[t, 1] = genotypes[t, 1]
+
+
+@numba_jit(nogil=True)
+def _random_gamete_Hamilton_Kerr(
+    genotype: ArrayLike, ploidy: int, tau: int, lambda_: float
+) -> ArrayLike:
+    if ploidy < len(genotype):
+        # remove fill values
+        genotype = genotype[genotype > -2]
+    if ploidy != len(genotype):
+        raise ValueError("Genotype ploidy does not match number of alleles")
+    if tau > ploidy:
+        # TODO: this can be an option for encoding somatic genome duplication (with suitable lambda)
+        raise NotImplementedError("Gamete tau cannot exceed parental ploidy")
+    gamete = np.random.choice(genotype, tau, replace=False)
+    if lambda_ > 0.0:
+        if tau == 2:
+            if np.random.rand() <= lambda_:
+                gamete[1] = gamete[0]
+        elif tau != 2:
+            raise NotImplementedError("Non-zero lambda is only implemented for tau = 2")
+    return gamete
+
+
+@numba_guvectorize(  # type: ignore
+    [
+        "void(int8[:,:], int64[:,:], uint64[:,:], float64[:,:], uint32, int8[:,:])",
+        "void(int16[:,:], int64[:,:], uint64[:,:], float64[:,:], uint32, int16[:,:])",
+        "void(int32[:,:], int64[:,:], uint64[:,:], float64[:,:], uint32, int32[:,:])",
+        "void(int64[:,:], int64[:,:], uint64[:,:], float64[:,:], uint32, int64[:,:])",
+    ],
+    "(n,k),(n,p),(n,p),(n,p),()->(n,k)",
+)
+def genedrop_Hamilton_Kerr(
+    genotypes: ArrayLike,
+    parent: ArrayLike,
+    tau: ArrayLike,
+    lambda_: ArrayLike,
+    seed: int,
+    out: ArrayLike,
+) -> None:  # pragma: no cover
+    if parent.shape[1] != 2:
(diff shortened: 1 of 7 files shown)


timothymillar:
I've tried to be consistent with dask stochastic functions, hence the use of a random seed at call time rather than compute time. This ensures that spereate evaluations of the same dask object will produce a consistent result. Using a random seed when none is specified also ensures that the dask object has a unique ID which is esential when evaluating multiple gene-drop simulations in a single compute call.

timothymillar:
The original issue "estimation of genotype probabilities" can be achieved like this:

```python
import xarray as xr
import sgkit as sg
from sgkit.tests.test_pedigree import random_parent_matrix

ds = sg.simulate_genotype_call_dataset(n_variant=10, n_sample=8)
ds["parent"] = ["samples", "parents"], random_parent_matrix(8, 3)
ds = sg.convert_call_to_index(ds).compute()
counts = xr.zeros_like(ds.call_genotype_index)

for _ in range(1000):
    sim = sg.genedrop(ds, merge=False)
    sim = sg.convert_call_to_index(sim)
    counts += ds.call_genotype_index == sim.call_genotype_index.compute()
counts / 1000
```
```
array([[1.   , 1.   , 1.   , 0.455, 0.486, 0.468, 0.242, 0.495],
       [1.   , 1.   , 1.   , 0.497, 0.5  , 0.483, 0.   , 0.537],
       [1.   , 1.   , 1.   , 0.   , 0.248, 0.26 , 0.505, 0.208],
       [1.   , 1.   , 1.   , 0.495, 0.254, 0.495, 0.478, 0.263],
       [1.   , 1.   , 1.   , 0.516, 0.504, 0.5  , 0.489, 0.349],
       [1.   , 1.   , 1.   , 0.506, 0.531, 0.247, 0.514, 0.519],
       [1.   , 1.   , 1.   , 0.   , 1.   , 1.   , 0.   , 0.531],
       [1.   , 1.   , 1.   , 0.481, 0.   , 0.   , 0.515, 0.5  ],
       [1.   , 1.   , 1.   , 0.487, 0.492, 0.509, 0.525, 0.506],
       [1.   , 1.   , 1.   , 0.   , 0.503, 0.513, 0.   , 0.   ]])
```


which avoids holding multiple simulations in memory at the same time. Although, the use of a python loop somewhat limits performance.

jeromekelleher:
I'll go through this more carefully when I get a chance @timothymillar, but I just wanted to get your thoughts on one point now.

Would ``sim_genedrop`` be better? I worry that ``genedrop`` might be interpreted as a mechanstic operation (extracting the exomes from the dataset or something).

It seems like a reasonable convention to adopt to add ``sim_`` prefix for simulation functions, in a library like this where it's not a core focus.

timothymillar:
Agreed, sim_ is a good convention. Naming things is hard!

timothymillar:
@jeromekelleher I've opted for `simulate_genedrop` for consistence with `simulate_genotype_call_dataset`.

sgkit/stats/genedrop.py
```diff
@@ -0,0 +1,344 @@
...
+    out: ArrayLike,
+) -> None:  # pragma: no cover
+    n_sample, n_parent = parent.shape
+    _, ploidy = genotypes.shape
+    if n_parent != 2:
+        raise ValueError("The parents dimension must be length 2")
+    if ploidy != 2:
+        raise ValueError("Genotypes are not diploid")
+    order = topological_argsort(parent)
+    np.random.seed(seed)
```
jeromekelleher, line 50: using the global rng like this is generally a Bad Idea. 
```
rng = numpy.random.default_rng(seed)
...
idx = rng.randint(2)
```
is a much better pattern. 

Does numba support doing it like this? If not, might be simplest to pass in a per-generated array of random values (even if you don't use them all)

sgkit/stats/genedrop.py
```diff
@@ -0,0 +1,344 @@
...
+) -> ArrayLike:
+    if ploidy < len(genotype):
+        # remove fill values
+        genotype = genotype[genotype > -2]
+    if ploidy != len(genotype):
+        raise ValueError("Genotype ploidy does not match number of alleles")
+    if tau > ploidy:
+        # TODO: this can be an option for encoding somatic genome duplication (with suitable lambda)
+        raise NotImplementedError("Gamete tau cannot exceed parental ploidy")
+    gamete = np.random.choice(genotype, tau, replace=False)
```
jeromekelleher, line 82: ditto here

jeromekelleher (commented):
Generally looks great, but I'm worried about how the global rng state is shared between chunks, and how this could result in weird correlated results. I think we need to be super-careful about randomness here.

The safest way to do it would be to pre-draw all the randomness you would need from an rng instance that you create. Otherwise, if you are going to create a new rng for each gufunc instance, you need to create a set of random seeds, I think. It's really quite tricky...

sgkit/stats/genedrop.py
timothymillar: Sorry for slow the response @jeromekelleher. Numba doesn't seem to support `numpy.random.default_rng`. For reference, the [docs](https://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#random) state "Since version 0.28.0, the generator is thread-safe and fork-safe. Each thread and each process will produce independent streams of random numbers."

sgkit/stats/genedrop.py
```diff
@@ -0,0 +1,344 @@
...
+            # copy founder
+            out[t, 0] = genotypes[t, 0]
+            out[t, 1] = genotypes[t, 1]
+
+
+@numba_jit(nogil=True)
+def _random_gamete_Hamilton_Kerr(
+    genotype: ArrayLike, ploidy: int, tau: int, lambda_: float
+) -> ArrayLike:
+    if ploidy < len(genotype):
```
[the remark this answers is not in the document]
timothymillar: Probably but it gets a little complex. These checks involve evaluation of the relevant arrays. So, an independent validation function could cause an array to be evaluated twice. This may be avoided if the separate validation function is somehow added to the task graph as a dependency of the compute function.

sgkit/stats/genedrop.py
jeromekelleher: That's very tricky then - it's super important that simulators are deterministic, so you can reproduce weird problems later. If there's a reasonable upper limit on the number of random values you need, I'd strongly recommend generating those beforehand with numpy and consuming within the jitted function.

timothymillar: As it stands, the simulations are deterministic if a random seed is specified by the user. If not, then a seed is randomly chosen and used to generate random seeds for the simulation to ensure that the resulting dask array produces the same result when re-evaluated. Would you preference be to require a random seed is given by the user?

sgkit/stats/genedrop.py
```diff
@@ -0,0 +1,345 @@
...
+    "Computation of the inverse additive relationship matrix for autopolyploid
+    and multiple-ploidy populations." Theoretical and Applied Genetics 131: 851-860.
+    """
+    ds = define_variable_if_absent(ds, variables.parent, parent, parent_indices)
+    variables.validate(
+        ds, {parent: variables.parent_spec, call_genotype: variables.call_genotype_spec}
+    )
+    gt = da.asarray(ds[call_genotype].data)
+    parent = da.asarray(ds[parent].data, chunks=ds[parent].shape)
+    if hasattr(seed, "__len__"):
```
timothymillar, line 304: @jeromekelleher, the code relevant to deterministic results starts here. We generate a random seed per variant at call time (unless provided by the user). This is used at evaluation time to determine the inheritance pattern at each variant. The numba random generator is thread safe so this should be sufficient and appears to be based on testing.

jeromekelleher (approved):
Thanks @timothymillar - I've looked up the docs and thought through the randomness and I think you've got it right. Since numba 0.28, the random streams per thread are independent, so we should be all good. It's a subtle point that wasn't obvious to me that numba is reimplementing np.random with these semantics, so might be worth a comment to explain.

timothymillar:
I'll add some comments in the numba functions. Thanks for raising the concern, it made me think it through more carefully. I'll also have another go at improving the performance of the Hamilton-Kerr implementation before I tag for merging.

sgkit/stats/genedrop_numba_fns.py
```diff
@@ -0,0 +1,169 @@
+import numpy as np
+
+from sgkit.accelerate import numba_guvectorize, numba_jit
+from sgkit.typing import ArrayLike
+
+from .pedigree import _compress_hamilton_kerr_parameters, topological_argsort
+
```
jeromekelleher, line 7: Maybe add a technical note about randomness here just so the reader can understand what np.random is doing below?

jeromekelleher (approved):
LGTM - just a quick comment about randomness at the top to help the reader I think.

sgkit/stats/genedrop_numba_fns.py
timothymillar: Yes, forgot that sorry.

PR #1184, opened 2024-02-05
Upgrade sphinx to fix build

tomwhite:


tomwhite:
The new doc packages won't install on Python 3.8. Since we don't need to build documentation on every Python version, I think the simplest way to fix this is to create a separate GitHub Actions workflow to check the documentation (`check-docs.yml`) which runs just on Python 3.9, and change `build.yml` that runs on all supported versions to just run the tests and upload coverage.

jeromekelleher (approved):

PR #1190, opened 2024-02-08
Errors in missing vs pad values in VCF

jeromekelleher:
Through developing the alternative implementation of vcf-to-zarr conversion in #1185 I think there's some bugs in how we're currently handling missing data. Opening this PR for discussion purposes.

There's some related issues around handling mixed ploidy calls, and string missing values, but let's leave those alone for now.

I realise now that the extra fields I added in to the simple test here duplicates later tests - but they're handy for discussion here for now anyway.


sgkit/io/vcf/vcf_reader.py
```diff
@@ -240,7 +240,7 @@ def for_field(
             dims.append(dimension)
             chunksize += (size,)
 
-        array = np.full(chunksize, fill_value, dtype=dtype)
+        array = np.full(chunksize, missing_value, dtype=dtype)
```
tomwhite, line 243: I don't understand this change, but I could be missing the overall motivation. The way I think about this is that all positions start as fill (pad) values, then values are filled in with actual or missing values as they are read in (and according to ploidy). 

Perhaps it would help to spell out a few cases that fail. They could well be in the tests, but there are a lot of changes there.

timothymillar:
@jeromekelleher your interpretation looks correct to me. We should also test for a case where `-2` should appear in an INFO field (e.g., with length `A` or `R`):

```
##INFO=<ID=IDX,Number=R,Type=Integer,Description="Index of each allele">
#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO          FORMAT        SAMPLE1        SAMPLE2
1       100     .       A       C,T     .       .       IDX=0,1,2     GT:HQ         0|1            0|2 
1       200     .       A       C       .       .       IDX=0,1       GT:HQ         0|0            0|1
```

With `max_alt_alleles=2`, the IDX field should correspond to the (variants * alleles) array:

```
[
    [0, 1,  2],
    [0, 1, -2],
]
```

jeromekelleher:
I think this is ready for a look now. There is a little duplication in the tests I've added for the sample VCF and existing ones which have been fixed up, but I think that's OK.

The change is a pretty noisy one I'm afraid, as there's been a few downstream things broken as well (#1195, #1196). I've temporarily skipped those tests to make this more manageable.

jeromekelleher:
Sigh - skipping those tests pushes the required coverage below 100% so the build still fails. 

Any suggestions here? Temporarily push coverage down, and create an issue to track setting it back to 100?

sgkit/io/vcf/vcf_reader.py
jeromekelleher: Ah yes, this wasn't necessary. fixed.

sgkit/tests/io/vcf/test_vcf_reader.py
```diff
@@ -945,29 +1105,32 @@ def test_vcf_to_zarr__parallel_with_fields(shared_datadir, tmp_path):
...
     # check strings have not been truncated after concat_zarrs
     assert_array_equal(
         ds["variant_allele"],
         np.array(
             [
-                ["T", "C", "<NON_REF>", ""],
-                ["T", "<NON_REF>", "", ""],
-                ["T", "G", "<NON_REF>", ""],
+                ["T", "C", "<NON_REF>", sfill],
+                ["T", "<NON_REF>", "", sfill],
```
tomwhite, line 1116: Looks like there should be another `sfill` here?

sgkit/tests/io/vcf/test_vcf_roundtrip.py
```diff
@@ -7,6 +7,7 @@
 from sgkit.tests.io.vcf.utils import assert_vcfs_close, path_for_test
 
 
+@pytest.mark.skip("Temporarily disabling pending bugfix; #1196")
```
tomwhite, line 10: Do you think the roundtrip tests are fixable given the changes in this PR, or is there not enough information being stored in VCF Zarr to support roundtripping?

tomwhite:
> Sigh - skipping those tests pushes the required coverage below 100% so the build still fails.
> 
> Any suggestions here? Temporarily push coverage down, and create an issue to track setting it back to 100?

I'm OK allowing coverage to dip as long as there is a path to get it back to 100%.

sgkit/tests/io/vcf/test_vcf_roundtrip.py
jeromekelleher: They're definitely fixable, I was just trying to keep the scope of the PR down (plus, I'm not that familiar with the vcf writing code).

jeromekelleher: If you think it's a straightforward fix, we could tack another commit on here for simplicity?

jeromekelleher:
Thanks @tomwhite - I'm happy to go with your preferred option. If we can fix the other problems fairly easily in this PR then that would be simplest. I'm just worried about doing too many changes at once.

sgkit/tests/io/vcf/test_vcf_roundtrip.py
tomwhite: I don't know how straightforward it is - the roundtrip tests are quite brittle (but valuable ones to have).

If you can get it working in this PR, that'd be great, but OK to push into another PR if you prefer.

jeromekelleher:
OK, this is ready for another look @tomwhite. I've temporarily worked around some fiddly corner cases - hopefully we'll still have 100% coverage. 

#1197 is a tricky one, but I don't think it's worth getting hung up on.

tomwhite (approved):

jeromekelleher:
OK! Looks like it's ready for a final look now @tomwhite.

@timothymillar, this is a reasonably big change so would be good to get your eyes on it as well if possible.

sgkit/tests/io/vcf/test_vcf_reader.py
```diff
@@ -1055,13 +1222,14 @@ def test_vcf_to_zarr__field_number_R(shared_datadir, tmp_path):
...
 
-    fill = INT_FILL
+    ifill = INT_FILL
+    imissing = INT_MISSING
     ad = np.array(
         [
-            [[40, 14, 0, fill]],
-            [[fill, fill, fill, fill]],
+            [[40, 14, 0, ifill]],
+            [[imissing, imissing, imissing, imissing]],
```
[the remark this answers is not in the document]
timothymillar: This was discussed at the previous meeting. Encoding an absent field is always going to be ambiguous unless we introduce a third sentinel value. Using the `missing` value for all elements is a reasonable compromise.

timothymillar (approved):
LGTM!

Issue #1156, opened 2023-12-14
Set SGKIT_DISABLE_NUMBA_CACHE=1 by default


jeromekelleher:
This is causing persistent problems, with cryptic segfaults being really really hard to debug with Dask (I just lost a whole morning to it!).

We have to turn off the numba cache until this is sorted out somehow.

tomwhite:
Is changing the default in https://github.com/pystatgen/sgkit/blob/main/sgkit/accelerate.py#L6 sufficient?

jeromekelleher:
I guess a warning if it is set also? It's a horrible gotcha.

jeromekelleher:
Partially addressed ion #1191.

To fully address I think we probably need to undocument this variable by removing the mention to it in the docs, and maybe add some comments in the source laying out the issue. Plus update to the changelog.

PR #1191, opened 2024-02-08
Turn off numba cache by default

jeromekelleher:
Partial fix for  #1156

This turns off numba caching by default, and at least lets us use the current git version of sgkit without hitting this issue repeatedly.

To fully address I think we probably need to undocument this variable by removing the mention to it in the docs, and maybe add some comments in the source laying out the issue.

### DIFF
--- a/sgkit/accelerate.py
+++ b/sgkit/accelerate.py
@@ -3,7 +3,7 @@ from typing import Callable
 
 from numba import guvectorize, jit
 
-_DISABLE_CACHE = os.environ.get("SGKIT_DISABLE_NUMBA_CACHE", "0")
+_DISABLE_CACHE = os.environ.get("SGKIT_DISABLE_NUMBA_CACHE", "1")
 
 try:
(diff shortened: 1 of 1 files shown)


tomwhite (approved):
LGTM

jeromekelleher:
Hmm, it seems like there's an unfortunate side effect here of turning off caching. With caching turned off, when we run the functions with NUMBA_DISABLE_JIT=1 to get coverage we get a whole bunch of numba errors and deprecations that don't happen when caching is on.

I'm not sure what to do here, there's quite a few deprecations and errors that seem non-trival to fix. For example, we seem to hit this problem https://github.com/numba/numba/issues/8886 which is non-trival to workaround.

Any thoughts @tomwhite, @timothymillar?

jeromekelleher:
I worked around this here by explicitly turning on caching, so that we can get our builds to pass and at least have our code not segfault by default. Opened #1194 to track the CI issue.

PR #1199, opened 2024-02-23
Avoid disabling jit on guvectorised functions 

### LINKED ISSUE #1194: Numba cache masking compile warnings and deprecations
Turning off the numba cache seems to cause our CI builds to fail when running tests in NOJIT mode (which we use to get coverage of jitted functions). Seems quite tricky to work around as discussed here: https://github.com/pystatgen/sgkit/pull/1191#issuecomment-1936994466

Related: #1156, #1191

timothymillar:
- [x] Fixes #1194

Leaving `NUMBA_DISABLE_JIT=1` for `test_pedigree.py` which contains many jitted functions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,18 +28,14 @@ jobs:
         pip install -r requirements.txt -r requirements-dev.txt
     - name: Run pre-commit
       uses: pre-commit/action@v2.0.0
-    - name: Test with pytest
+    - name: Test with pytest (numba jit disabled)
       env:
-        # Enable cache to workaround build problems
-        # https://github.com/pystatgen/sgkit/issues/1194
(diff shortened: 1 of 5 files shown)


timothymillar:
Hmm, not sure what to do about coverage here

timothymillar:
This still runs tests with the jit disabled for two files:
- `sgkit/tests/test_pedigree.py`
- `sgkit/tests/io/vcf/test_vcf_writer_utils.py`

Both of these files have lots njit functions without guvectorize. This is a bit fragile so we'll see if it sticks around.

I also looked into using pytest to mark certain tests for running with the jit disabled. This doesn't work because the guvectorize functions are compiled at import time which happens even if the tests are deselected.

jeromekelleher (approved):
LGTM.

@tomwhite, what do you think?

tomwhite (approved):

PR #1208, opened 2024-03-05
Partial changes for VCF Character S->U

jeromekelleher:
Updates for vcf-zarr spec change: https://github.com/pystatgen/vcf-zarr-spec/issues/14

Almost working, I think the main issues remaining are in the round-trip tests.

Nobody uses Character columns in the real world I think, so no need to worry too much about compatibility.


tomwhite:
Fixed in https://github.com/sgkit-dev/vcztools/pull/200

PR #1229, opened 2024-06-19
Add 'matching' method option to identity_by_state

### LINKED ISSUE #1227: Option to calculate IBS probabilities via allele matching
The `identity_by_state` method calculates mean probabilities of identity by state (IBS) from call_allele_freqeuncies. The current implementation is fairly efficient when the alleles dimension is small, but not when the alleles dimension is large. An alternative approach is to directly compare allelic states within the call_genotype array. This is likely to be less efficient for small numbers of alleles (O(K*K) where K is the maximum ploidy), but much more efficient with large numbers of alleles (especially in terms of memory). This could be implemented as an optional method `sg.identity_by_state(ds, method='matching')`.

timothymillar:
- [x] Fixes #1227 
- [x] Tests added

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,8 +8,11 @@ Changelog
 0.10.0 (unreleased)
 -----------------------
 
-.. New Features
-.. ~~~~~~~~~~~~
+New Features
+~~~~~~~~~~~~
+
+- Add 'matching' method to :func:`identity_by_state` function.
+  (:user:`timothymillar`, :pr:`1229`, :issue:`1227`)
 
 .. Breaking changes
 .. ~~~~~~~~~~~~~~~~
--- a/sgkit/stats/ibs.py
+++ b/sgkit/stats/ibs.py
@@ -1,6 +1,7 @@
 from typing import Hashable
 
 import dask.array as da
+from typing_extensions import Literal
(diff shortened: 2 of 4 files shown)


jeromekelleher (approved):
LGTM

jeromekelleher:
Requiring a single chunk in the samples dimension is a bit restrictive - perhaps we could add a note on how that could be generalised?

timothymillar:
Thanks @jeromekelleher, I can have a look into chunking in the samples dimension before I merge.

timothymillar:
Do you want to have another look at this @jeromekelleher? It now supports chunking of the samples dimension with quite a bit more code. I would have preferred to avoid having two gufuncs but doing so gives an almost 2x speed up when there is no chunking in the samples dimension (most common case).

Some rough benchmarks show how the 'matching' approach scales much better with the number of alleles. Memory ussage is also much better. The 'frequencies' method is roughly an order of magnitude faster in the biallelic case: 

matching_perf

(Other params where: `n_variant=10_000`, `n_sample=100`, `n_ploidy=4`, `missing_pct=0.01` and chunks of 5000 variants * 50 samples).

jeromekelleher (approved):
LGTM!

Issue #908, opened 2022-09-22
Run on Cubed


tomwhite:
This is an umbrella issue to track the work needed to run sgkit on [Cubed](https://github.com/tomwhite/cubed).

This is possible because Cubed exposes the [Python array API standard](https://data-apis.org/array-api/latest/) as well as [common Dask functions and methods](https://tom-e-white.com/cubed/api.html) like  `map_blocks` and `Array.compute`. Also, there is ongoing work to [Integrate cubed in xarray](https://github.com/pydata/xarray/pull/7019), as a part of exploring [alternative parallel execution frameworks in xarray](https://github.com/pydata/xarray/issues/6807).

tomwhite:
Also, this example shows that Cubed works with Numba (locally at least), which answers @hammer's question here: https://github.com/pystatgen/sgkit/issues/885#issuecomment-1209288596.

tomwhite:
Since I opened this issue almost two years ago, Xarray has added a chunk manager abstraction (https://docs.xarray.dev/en/stable/internals/chunked-arrays.html), which makes it a lot easier to switch from Dask to Cubed as the backend computation engine *without* changing the code to express the computation. The nice thing about this approach is that we can use Dask or Cubed or any other distributed array engine that Xarray might support in the future (such as Arkouda).

I've started to explore what this might look like in https://github.com/tomwhite/sgkit/tree/xarray-apply-ufunc, but the two main ideas are:
1. move from `dask.array.map_blocks` to `xarray.apply_ufunc` for applying functions in parallel,
2. have a way to run the test suite using either Dask or Cubed so we don't have to do all the changes at once.

The code in the branch does this for `count_call_alleles`. As you can see in this commit (https://github.com/sgkit-dev/sgkit/commit/383398214da8a9184fbe8ff9c874a726f68ba72f), another minor benefit of using `xarray.apply_ufunc` is we can use named dimensions like `ploidy` and `alleles` rather than dimension indexes like 2.

This commit (https://github.com/sgkit-dev/sgkit/commit/da8657e994bfad2214eeeebb3c50f4e46d86577a) shows the new pytest command-line option to run on cubed: `--use-cubed`.

I would be interested in any thoughts on this direction @jeromekelleher, @hammer, @timothymillar, @benjeffery, @ravwojdyla, @eric-czech.

[...]

Issue #871, opened 2022-07-07
Avoid dependencies between chunks in variants dimensions caused by unnecessary dask usage


timothymillar:
I've noticed that `count_call_alleles` and some methods using cohorts create unnecessary dependencies between chunks in the variants dimension. For example the current task graph for `observed_heterozygosity` on a dataset with 10 chunks in the variants dimentions looks like this:
<details>
<summary>Task graph</summary>
<br>

![obshet_old](https://user-images.githubusercontent.com/14065102/177886561-2fed2c6b-7c02-4599-9db1-ab1deeacae1c.png)

</details>

In `count_call_alleles` this is a result of using `da.empty` to indicate the number of alleles for a gufunc. In `observed_heterozygosity` (and also `diversity`) this is caused by forcing the `sample_cohort` array to be a dask array which doesn't achieve much because we immediately call compute on that array to get the number of cohorts. Replacing both of these cases with numpy arrays results in the following equivalent task graph:
<details>
<summary>Task graph</summary>
<br>

![obshet_new](https://user-images.githubusercontent.com/14065102/177887189-41e8c297-8e76-47c8-8142-276530617821.png)

</details>

My understanding is that the second task graph should be more efficient to schedule at larger scales (can any dask experts confirm?). Is there any reason not to make such a change? I guess it makes the use of the `sample_cohort` array a little bit opaque.

PR #1249, opened 2024-09-02
Run `test_count_call_alleles` on Cubed

tomwhite:
See #908

### DIFF
--- a/.github/workflows/cubed.yml
+++ b/.github/workflows/cubed.yml
@@ -1,0 +1,33 @@
+name: Cubed
+
+on:
+  push:
+  pull_request:
+  # manual trigger
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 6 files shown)


tomwhite:
This is passing all tests, but when producing a visualization of the Dask graph I noticed that it reintroduces the problem in #871, so shouldn't be merged. (cc @timothymillar)

tomwhite:
In the meeting @jeromekelleher suggested that we could abstract the `map_blocks` bit into a function in sgkit that dispatches either to `da.map_blocks` (for Dask arrays) or `xr.apply_ufunc` (for everything else, including Cubed). This would keep the Dask path as it is today, while allowing other array libraries to work.

tomwhite:
Superseded by #1254

PR #1248, opened 2024-09-02
Pin Dask to fix slow main build

tomwhite:
- [x] Fixes #1247

PR #1261, opened 2024-09-19
Install dask[dataframe] explicitly to fix upstream error

tomwhite:


tomwhite:
The tests now run, but are failing for the same reason as #1248 - so that will need further investigation.

jeromekelleher (approved):

tomwhite:
This is failing due to #1267

Issue #1218, opened 2024-05-15
Python 3.12 conda install fails.


benjeffery:
Seems sgkit can't be installed on 3.12 due to cbgen:

Pinned packages:
  - python 3.12.*

Could not solve for environment specs
The following package could not be installed
└─ sgkit 0.8.0  is installable and it requires
   └─ cbgen with the potential options
      ├─ cbgen [1.0.0|1.0.1] would require
      │  └─ python >=3.7,<3.8.0a0 , which can be installed;
      ├─ cbgen [1.0.0|1.0.1] would require
      │  └─ python >=3.8,<3.9.0a0 , which can be installed;
      ├─ cbgen [1.0.0|1.0.1] would require
      │  └─ python >=3.9,<3.10.0a0 , which can be installed;
      └─ cbgen 1.0.1 would require
         └─ python >=3.10,<3.11.0a0 , which can be installed.

PR #1263, opened 2024-09-24
Test on Python 3.12

tomwhite:


tomwhite:
The first cbgen version that supports Python 3.12 v10.0.5, but this requires NumPy 2, which is why this is failing.

See also #1218

jeromekelleher:
Yeesh, this is looking hairy! Thanks for fighting the good fight here Tom!

tomwhite:
OK, I've got this passing now! There were a few things that needed fixing: the NumPy 2/cbgen/bio2zarr version constraints, and `pkg_resources` no longer being supported in Python 3.12.

jeromekelleher (approved):

PR #1274, opened 2024-10-28
Use sgkit.distarray for PCA

tomwhite:
More progress on #908

This needs https://github.com/cubed-dev/cubed/pull/598 which hasn't been merged yet


tomwhite:
This is working, now that [SVD](https://github.com/cubed-dev/cubed/pull/598) for Cubed has been merged.

This change involves adding a copy of `TruncatedSVD` from Dask-ML, suitably modified to support the Array API, as well as Dask. Since this was the only code we use from Dask-ML it allows us to remove the Dask-ML dependency.

In the future we may be able to remove the copy of `TruncatedSVD` entirely, and use [the original](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html) from scikit-learn - but only once it has been converted to be Array API compatible (note that [regular PCA has already been converted](https://github.com/scikit-learn/scikit-learn/pull/26315)).

jeromekelleher (approved):
Looks great - num_dependencies -= 1!

PR #1309, opened 2025-04-14
Fix vcf2zarr tests

### LINKED ISSUE #1308: ImportError: cannot import name 'vcf2zarr' from 'bio2zarr'
See https://github.com/sgkit-dev/sgkit/actions/runs/14435286128/job/40475317044

It looks like https://github.com/sgkit-dev/bio2zarr/pull/339 caused this.

tomwhite:
- [x] Fixes #1308


tomwhite:
I got `ValueError: Zarr schema format version mismatch: 0.4 != 0.5`, so I regenerated the schema, see 3cae0f6ce99c193abc41a1415bf0c07e79ae3caa.

But now the docs build is failing with

```
"'alleles' is not a valid dimension or coordinate for Dataset with dimensions FrozenMappingWarningOnValuesAccess({'variants': 10879, 'samples': 250, 'FORMAT_AD_dim': 2})"
```

I noticed that for `call_AD` the dimension is now `FORMAT_AD_dim`, not `alleles`.

Any idea why that's happening @benjeffery?

jeromekelleher:
> I noticed that for call_AD the dimension is now FORMAT_AD_dim, not alleles.

That's a good catch @tomwhite, I think we can regard that as bug. Probably something simple happened in the generalisation code, I'll investigate

jeromekelleher:
This is weird @tomwhite, I don't see what would have changed here. The number in VCF is "." in the 1000G files I'm looking at, and I assume it's the same here, so I don't see how we'd ever have gotten "alleles" in there. There was never any special case for AD, and I don't think we should overrule the header anyway.

Did you manually edit the schema to fix this problem before by any chance?

tomwhite:
> This is weird @tomwhite, I don't see what would have changed here. The number in VCF is "." in the 1000G files I'm looking at, and I assume it's the same here, so I don't see how we'd ever have gotten "alleles" in there. There was never any special case for AD, and I don't think we should overrule the header anyway.
> 
> Did you manually edit the schema to fix this problem before by any chance?

Yes, that was it. Sorry!

jeromekelleher:
Great! LGTM

PR #1330, opened 2025-07-01
Add `bcftools`-style filtering

### LINKED ISSUE #1329: Support `bcftools`-style filtering
It would be very useful to have the ability to subset a dataset using `bcftools` expressions for regions (`-r/-t`), filter expressions (`-i/-e`), and samples (`-s`).

This is now possible using the work we have done in `vcztools`, such as https://github.com/sgkit-dev/vcztools/pull/234. In particular, we now have an index that allows efficient region queries (which will address #25).

This will also address #1084.

See an initial prototype in https://github.com/sgkit-dev/bio2zarr/issues/365#issuecomment-2859114862

tomwhite:
- [x] Fixes #1329
- [x] Tests added

### DIFF
--- a/sgkit/filtering.py
+++ b/sgkit/filtering.py
@@ -1,0 +1,167 @@
+from typing import Optional, Union
+
+import numpy as np
+import xarray as xr
+from vcztools import filter as filter_mod
+from vcztools.regions import (
+    parse_regions,
+    parse_targets,
+    regions_to_chunk_indexes,
+    regions_to_selection,
+)
+from vcztools.samples import parse_samples
+
+
+def bcftools_filter(
+    ds,
+    *,
+    regions: Optional[str] = None,
+    targets: Optional[str] = None,
+    include: Optional[str] = None,
+    exclude: Optional[str] = None,
+    samples: Union[list[str], str, None] = None,
+):
+    """Filter a dataset using `bcftools`-style expressions.
+
+    The dataset can be subset in the variants dimension to a set
+    of genomic regions and/or filter expressions that match sites
+    using the `bcftools` `expression language <https://samtools.github.io/bcftools/bcftools.html#expressions>`_
+    Additionally, the samples dimension can optionally be subset
+    to a list of sample IDs.
+
+    Parameters
+    ----------
+    ds
+        The dataset to be filtered.
+    regions
+        The regions to include specified as a comma-separated list of
+        region strings. Corresponds to `bcftools` `-r`/`--regions`.
+    targets
+        The region targets to include specified as a comma-separated list
+        of region strings. Corresponds to `bcftools` `-t`/`--targets`.
+    include
+        Filter expression to include variant sites.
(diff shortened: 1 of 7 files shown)


sgkit/filtering.py
```diff
@@ -0,0 +1,167 @@
...
+    data_arrays = tuple(ds_filter_fields[n] for n in ds_filter_fields.data_vars)
+    da = xr.apply_ufunc(
+        compute_call_mask, *data_arrays, output_dtypes=[bool], dask="parallelized"
+    )
+    ds["call_mask"] = da
+
+    # filter to variants where at least one sample has been selected
+    # note we have to call compute here since xarray indexing with a Dask or Cubed
+    # boolean array is not supported
+    return ds.isel(variants=ds.call_mask.any(dim="samples").compute())
```
jeromekelleher, line 161: What are the implications here for the memory requirements when working on a large dataset?

jeromekelleher (commented):
Cool!

sgkit/filtering.py
tomwhite: Good question. It will materialise a 1D boolean array of length #variants into memory, which unless a restrictive region filter has been applied first could be very large (potentially the whole genome).

This is an area where there is scope to improve - perhaps by making Xarray and the underlying distributed processing engine able to handle this case efficiently, or by using masked arrays in some way?

sgkit/filtering.py
```diff
@@ -0,0 +1,167 @@
...
+            f'Dataset must have uniform chunk sizes in variants dimension, but was {ds.chunks["variants"]}.'
+        )
+    variant_chunksize = chunks[0]
+
+    variant_slice = slice(
+        int(chunk_indexes[0] * variant_chunksize),
+        max(ds.sizes["variants"], int(chunk_indexes[-1] * variant_chunksize + 1)),
+    )
+
+    ds_sliced = ds.isel(variants=variant_slice)
```
tomwhite, line 110: This works but could be very inefficient. Imagine we just want one small region at the start of the genome and one at the end - this would read all of the intervening chunks even though they are not needed.

What we really want is a way to index by a set of slices (or even chunks) in one go. NumPy and Xarray don't provide such a primitive, but it's something we could perhaps build.

cc-ing @keewis @TomNicholas and @alxmrs as we were discussing this exact issue in yesterday's Pangeo Distributed Computing meeting (in the context of a geospatial application Justus is working on).

keewis: I've opened pydata/xarray#10479 to discuss this for `xarray`

tomwhite:
Looks like the build is failing on Python 3.12 due to https://github.com/pyranges/sorted_nearest/issues/10 from pyranges. Not sure why vcztools CI is working on Python 3.12 though.

sgkit/filtering.py
jeromekelleher: I think a 1D boolean array length #variants is fine - it could be improved but it's not terrible. I was worried that it was a O(num_samples) thing.

PR #1335, opened 2025-09-26
Fix high memory usage bug with chunked HK relationship #1333

### LINKED ISSUE #1333: High memory usage when calculating chunked Hamilton-Kerr relationship matrix
This is a bug in `pedigree_kinship` when used with `method='Hamilton-Kerr'`, `return_relationship=True` and `chunks=...`. Converting the (chucked) kinships to relationships requires multiplying by a combination of the (potentially non-equal) ploidies. This is currently being achieved with a non-chunked matrix resulting in high memory usage when evaluating a single chunk of the returned relationship matrix. The offending code block is here:
https://github.com/sgkit-dev/sgkit/blob/main/sgkit/stats/pedigree.py#L1202

This should be simple to fix with a chunked ploidy matrix.

timothymillar:
- [x] Fixes #1333
- [x] Tests added

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -28,8 +28,12 @@ Deprecations
 .. Improvements
 .. ~~~~~~~~~~~~
 
-.. Bug fixes
-.. ~~~~~~~~~
+Bug fixes
+~~~~~~~~~
+
+- Fixed bug causing high memory usage in :func:`pedigree_kinship` when
+  evaluating a chunked relationship matrix with the Hamilton-Kerr method.
+  (:user:`timothymillar`, :pr:`1335`, :issue:`1333`)
 
 .. Documentation
 .. ~~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


jeromekelleher (commented):
LGTM!

Is there a time-cost for rechunking like this on small inputs? (I'm not entirely clear what happens by calling ``rechuink``)

jeromekelleher (approved):

timothymillar:
I don't think there is any cost to this rechunk unless the 'chunks' argument is actually used. By default, this will just 'rechunk' a single chunk array to another single chunk array. Dask doesn't appear to even include the operation in the task graph (with `.visualize(optimize_graph=False)`). 

I tried rechunking a dummy array to its original size 10_000 times in a loop and don't notice any slowdown on `.compute()`. So I'm assuming that this case is recognized at call time and doesn't affect the actual compute graph.

timothymillar:
Looks like the CI failures are unrelated, Docs failing on 3.11 isn't ideal but it seems to be an issue with the ipynb tutorials not executing.

Issue #1338, opened 2026-02-03
Fix CI breakages + questions on support policy


Billyzhang1229:
Hi sgkit community, I want to more actively maintain this toolkit and would like to start by fixing the CI. I’ve fixed some issues, but there are a few points I’m not sure about and would like feedback.

What I’ve done

 1. Pinned pandas < 3.0 to fix CI.
     Dask is not fully supported with pandas 3.0 yet, so this is likely a temporary pin until the ecosystem stabilizes.
 2. Apple M4 + NumPy:
     On Apple M4 chips, NumPy < 2.3.5 can emit runtime warnings in matmul.
     I split NumPy constraints because:
      - Python 3.10 compatibility only works with NumPy < 2.3.0
      - Apple M4 issue is resolved by numpy >= 2.3.5
        Reference:

     https://github.com/numpy/numpy/issues/29820
     https://github.com/numpy/numpy/pull/30102
     https://numpy.org/doc/stable/release/2.3.0-notes.html
 3. xarray xr.concat FutureWarning (data_vars=None default change)
     I made the test explicit with data_vars="all".
     Reference:

     https://github.com/pydata/xarray/issues/8778
     https://github.com/pydata/xarray/pull/10062
 4. Zarr error regex update
     Updated the regex to match the new “contains data” wording in Zarr v3.
     Reference:

     https://zarr.readthedocs.io/en/v3.0.9/_modules/zarr/storage/_common.html

 Questions:

  - Based on the NumPy/Python support tables, are we planning to drop Python 3.10 and NumPy < 2 soon?
    Reference:

    https://devguide.python.org/versions/#supported-versions
    https://numpy.org/neps/nep-0029-deprecation_policy.html#support-table
    https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule
  - How should we handle warnings in CI for Zarr > 3 (ignore in tests vs. require code changes)?

tomwhite:
Thanks @Billyzhang1229, this all looks very sensible.

> Based on the NumPy/Python support tables, are we planning to drop Python 3.10 and NumPy < 2 soon?

We can certainly drop NumPy 1.x, as we've done that in bio2zarr and vcztools recently without any issues.

I would be OK with dropping Python 3.10 too, as we're not planning an sgkit release soon. What do you think @jeromekelleher?

> How should we handle warnings in CI for Zarr > 3 (ignore in tests vs. require code changes)?

[...]

PR #1339, opened 2026-02-03
Fix CI and other issues

Billyzhang1229:
- Pin pandas <3.0 in requirements and pyproject (Dask/pandas 3.0 not fully supported yet)
  - Split NumPy constraints by sys_platform (Apple ARM uses NumPy >=2.3.5)
  - Drop Python 3.10 and NumPy 1.x support (requirements + CI workflows + docs)
  - Make xr.concat data_vars explicit to silence FutureWarning
  - Expand Zarr error regex to accept “contains data”
  - Ignore .hypothesis in pytest collection
  - Filter known runtime warnings (Numba correlation_reduce_cpu, Zarr v3 dtype warning)
  - Fix Windows CI: conda only for msprime, pip for requirements (PEP 508 markers)
  - Docs: update xarray intersphinx link and use PEP 604 unions to avoid typing.Union nitpicks
  - Temporary disable notebook execution due to missing link
  - Update mergify Python targets
  
  #1338


jeromekelleher (commented):
This all LGTM - thanks @Billyzhang1229!

jeromekelleher:
We could also consider dropping Mergify - the built in GitHub support has been working well over in tskit-dev.

jeromekelleher (approved):
LGTM!

Billyzhang1229:
The only failing job is the Cubed workflow.

The Cubed workflow explicitly installs dev builds from GitHub, overriding the pinned dependencies in requirements.txt. In .github/workflows/cubed.yml it runs:

python -m pip install -U git+https://github.com/cubed-dev/cubed.git -U git+https://github.com/cubed-dev/cubed-xarray.git -U git+https://github.com/pydata/xarray.git

The Cubed failure is due to an upstream FutureWarning / behavior change related to tuple-based chunk specification in xarray 2026.1.1.dev18.

tomwhite:
> The only failing job is the Cubed workflow.

It would be OK to skip this until we can fix it.

Billyzhang1229:
Cool, I’ll go ahead and squash the commits.

jeromekelleher:
These commits all seem like they do something well specified, so no need to squash from my perspective

Billyzhang1229:
There are a couple of small dev commits (15c21ad, 27d6a19) where I iterated on the same change. I’ll tidy those up / revert the redundant bits to keep the history clean, but otherwise I’ll leave the commits as-is rather than squashing.

Billyzhang1229:
I’ve tidied things up a bit now. removed the dev commits.

jeromekelleher:
Thanks @Billyzhang1229! Let's merge, as a (big!) step in the right direction.

PR #1344, opened 2026-09-03
Install Windows CI dependencies with pip instead of conda, test 3.12 too

Billyzhang1229:
The Windows workflow started failing in `test_genee` with warnings treated as errors:

```
RuntimeWarning:
Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
the same time. Both libraries are known to be incompatible ...
```

`conda install msprime` on conda-forge now resolves to MKL (`libiomp`) plus `llvm-openmp` (`libomp`), while the pip-installed scientific stack brings its own OpenMP, so threadpoolctl sees both runtimes. I confirmed the failure is independent of the `setup-miniconda` version by running the same job with v2 and v4.

conda was only used on Windows because msprime had no Windows wheels at the time. It now publishes them for Python 3.11 through 3.14, so this switches the job to `setup-python` + pip like the other workflows and drops the miniconda step entirely.

Second commit adds Python 3.12 to the matrix so Windows covers the same versions as the Build workflow.

Verified green on my fork 
- (3.11): https://github.com/Billyzhang1229/sgkit/actions/runs/33775450800/job/100715880709
- (3.12):https://github.com/Billyzhang1229/sgkit/actions/runs/33775450800/job/100715880446

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -17,24 +17,16 @@ jobs:
         python-version: ["3.11"]
 
     steps:
-    - uses: actions/checkout@v4
-    - name: Set up Miniconda with Python version ${{ matrix.python-version }}
-      uses: conda-incubator/setup-miniconda@v2
+    - uses: actions/checkout@v7
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v7
       with:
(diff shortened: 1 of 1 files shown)


jeromekelleher:
This seems like a straightforward win, merging

PR #1345, opened 2026-09-03
Fix Cubed workflow: Python 3.12 and dict chunks

Billyzhang1229:
The Cubed workflow has been red for a while. Two independent problems:

1. `cubed` and `cubed-xarray` now require Python >= 3.12, so the install step failed:
   ```
   ERROR: Package 'cubed' requires a different Python: 3.11.16 not in '>=3.12'
   ```
   First commit bumps the job to 3.12.

2. With that fixed, 10 tests failed because xarray main now warns on tuple chunks and warnings are errors:
   ```
   FutureWarning: Supplying chunks as dimension-order tuples is deprecated.
   It will raise an error in the future. Instead use a dict with dimension names as keys.
   ```
   Second commit rewrites the test parametrizations to use dimension names (`{"variants": 100, "samples": 10, "ploidy": -1}` instead of `(100, 10, -1)`), so the call sites stay as they were. Also updates the rechunking hint in `pca.py` to the dict form. Tests pass locally against both the pinned xarray 2025.3.0 and xarray main.

Verified green on my fork: https://github.com/Billyzhang1229/sgkit/actions/runs/33771138386

### DIFF
--- a/.github/workflows/cubed.yml
+++ b/.github/workflows/cubed.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.11"]
+        python-version: ["3.12"]
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


jeromekelleher (approved):
This seems like a clear win too. Would you like to take a look @tomwhite?

tomwhite (approved):

PR #1347, opened 2026-09-03
Remove Mergify configuration

### LINKED ISSUE #1340: Drop Mergify, move to GitHub approach
I think we're using it on bio2zarr etc.

Billyzhang1229:
Closes #1340. Supersedes #1316.

Mergify hasn't merged anything here for a long time: the recent PRs were all merged by hand, and the config is now rejected by Mergify itself (`Extra inputs are not permitted @ pull_request_rules → actions → queue → method → rebase`), which is why every open PR shows a failing "Mergify Merge Queue" check. bio2zarr and vcztools merge directly on GitHub, so this does the same.

Two things need an admin in the repository settings once this is merged:
- uninstall the Mergify app, so the failing check disappears;
- enable "Automatically delete head branches", which replaces the `delete_head_branch` rule.


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
```
jeromekelleher: Can we leave dependabot out of this for now please? This config will lead to a lot of prs, and it's not clear who will review and merge them

Billyzhang1229: Sure, dropped.

PR #1349, opened 2026-09-04
Fix benchmark workflow

Billyzhang1229:
The benchmarks workflow has been red. Two independent problems:

1. `benchmarks/asv.conf.json` leaves `build_command` commented out, so asv uses
   its default, which runs `pip wheel` without `--no-deps` and drops a wheel for
   every dependency into the build cache. asv 0.6.6 turned more than one wheel
   there into a hard error:
   ```
   ·· Last error: Found multiple wheels in .../asv-build-cache/... Cannot decide correct one
   ```
   First commit sets `build_command` explicitly, which also stops the job
   depending on asv's defaults. A later commit pins asv, since its version is
   not recorded in the results.

2. `count_call_alleles` and `count_cohort_alleles` return lazy dask arrays and
   the benchmarks never computed them, so they were timing the construction of
   the task graph rather than the counting — which is why they have reported an
   almost unchanging 0.12s for years. Both now compute the result, with a numba
   warmup in `setup` so jit compilation is not measured.

The second change moves them from ~0.12s to ~1.1s and resets their history, as
the old values were measuring something else. 

Verified by a full successful run of the workflow on a fork.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -28,7 +28,7 @@ jobs:
         run: |
           sudo apt update -y
           python -m pip install --upgrade pip
-          pip install asv
+          pip install asv==0.6.6
 
       - name: Set and log asv machine configuration
         run: |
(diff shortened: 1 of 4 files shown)


Billyzhang1229:
image
Just hate seeing a red cross here XD

jeromekelleher (approved):

PR #1350, opened 2026-09-05
Remove benchmarks and modernize upstream dependency installation

Billyzhang1229:
The benchmark workflow has a high maintenance cost and is not currently providing useful information. This PR removes the workflow and ASV suite, including the README badge, development dependency, contributor documentation, and related ignore settings.

It also updates upstream dependency installation to work with modern pip:

- Use PEP 508 Git URLs and resolve upstream and test dependencies in one install, without the legacy resolver workaround.
- Remove version bounds from other dependencies while preserving extras and platform markers. Use Zarr's `support/v2` branch because rechunker requires Zarr < 3.
- Fix the editable install arguments and run upstream tests on Python 3.12 and 3.13.

Validation:

- Workflow lint and `git diff --check` passed. Checked that no active ASV or benchmark workflow references remain.
- The rebased tree is identical to the tree before commit cleanup.
- The earlier [upstream workflow run](https://github.com/Billyzhang1229/sgkit/actions/runs/33985234765) installed successfully on Python 3.12 and 3.13. Each reported 1368 passed, 41 failed, and 12 errors; upstream compatibility failures remain for separate follow-up. The full test suite was not rerun after removing ASV.


jeromekelleher (commented):
Thanks for getting this working @Billyzhang1229 . To be honest I think we should just delete the benchmarks workflow, as it's not worth the maintenance burden. It's going to break again.

@tomwhite, what do you think?

Billyzhang1229:
> Thanks for getting this working @Billyzhang1229 . To be honest I think we should just delete the benchmarks workflow, as it's not worth the maintenance burden. It's going to break again.

having a public benchmark to compare against plink might still be worth doing somewhere in the docs or README. But yeah, since we can’t really fix things like the CPU specs for the Actions runners, it’s not actually a very reliable comparison anyway. So happy to remove the benchmarks workflow.

tomwhite:
> To be honest I think we should just delete the benchmarks workflow, as it's not worth the maintenance burden. It's going to break again.

I agree. It's not providing useful information at the moment.

Billyzhang1229:
Thanks @jeromekelleher and @tomwhite. I've removed the benchmarks workflow and ASV suite, along with the README badge, dependency, and related docs.

jeromekelleher (approved):

